PR #16, opened 2018-01-30
add accessibility tests

djwester:
add accessibility checks to make sure we are correctly following standards

### DIFF
--- a/d2l-rubric-criteria-group.html
+++ b/d2l-rubric-criteria-group.html
@@ -27,7 +27,7 @@
 
 		</style>
 		<d2l-table-wrapper>
-			<table class="d2l-table" summary="This table lists criteria and criteria group names in the first column. The first row lists level names and includes scores if the rubric uses a numeric scoring method. A similar row starts any additional criteria group.">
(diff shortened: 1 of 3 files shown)


d2l-rubric-criteria-group.html
```diff
@@ -27,7 +27,7 @@
 
 		</style>
 		<d2l-table-wrapper>
-			<table class="d2l-table">
+			<table class="d2l-table" summary="This table lists criteria and criteria group names in the first column. The first row lists level names and includes scores if the rubric uses a numeric scoring method. A similar row starts any additional criteria group.">
```
alogan519: I can localize this

djwester: do you want to do that after I merge this, or should we put this fie into a different PR?

alogan519: Actually this might not be true in our case "A similar row starts any additional criteria group."

alogan519: ya after is good.

djwester: I have updated the summary to reflect the fact that the WC has each group in it's own table

PR #19, opened 2018-02-05
add minimum column widths

alogan519:
Also right align the out of column

### DIFF
--- a/d2l-rubric-criteria-group.html
+++ b/d2l-rubric-criteria-group.html
@@ -30,12 +30,7 @@
 			th > div {
 				text-align: center;
 			}
-			.levels-header {
-				font-weight: bold;
-			}
 			.custom-points {
-				font-weight: bold;
-				color: var(--d2l-color-galena);
 				margin-bottom: 0.5rem;
 			}
 
@@ -64,9 +59,9 @@
 								<template is="dom-repeat" as="criterionCell" items="[[_getCriterionCells(item)]]">
 									<td selected>
 										<div class="description">
-											<div hidden=[[!_hasCustomPoints(criterionCell)]] class="custom-points">
+											<label class="custom-points d2l-label-text" hidden=[[!_hasCustomPoints(criterionCell)]] class="custom-points">
 												[[criterionCell.properties.points]] points
-											</div>
+											</label>
 											<span>[[_getDescription(criterionCell)]]</span>
 										</div>
 									</td>
--- a/d2l-rubric-criteria-group.html
+++ b/d2l-rubric-criteria-group.html
@@ -10,8 +10,17 @@
 			:host {
 				display: block;
 			}
+			td {
+				min-width: 130px;
+			}
 			td:first-of-type {
 				background-color: var(--d2l-color-regolith);
+			}
+			td:last-of-type {
+				min-width: 0px;
+			}
+			td:last-of-type > div {
+				text-align: right;
 			}
 			th > div {
 				text-align: center;
@@ -55,7 +64,7 @@
 								</td>
 							</template>
 							<td>
-								/[[maxLevelScore]]
+								<div>/[[maxLevelScore]]</div>
 							</td>
 						</tr>
 					</template>
--- a/d2l-rubric-criteria-group.html
+++ b/d2l-rubric-criteria-group.html
@@ -8,9 +8,7 @@
 
 <dom-module id="d2l-rubric-criteria-group">
 	<template>
-		<custom-style include="d2l-typography">
-			<style is="custom-style" include="d2l-typography"></style>
-		</custom-style>
+		<style include="d2l-typography"></style>
 		<style include="d2l-table-style">
 			:host {
 				display: block;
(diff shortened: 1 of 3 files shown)


d2l-rubric-criteria-group.html
```diff
@@ -55,7 +64,7 @@
 								</td>
 							</template>
 							<td>
-								/[[maxLevelScore]]
+								<div>/[[maxLevelScore]]</div>
```
dbatiste: This `/` might not make sense to a screen reader user.

d2l-rubric-criteria-group.html
```diff
@@ -10,9 +10,18 @@
...
 				display: block;
 			}
+			td {
+				min-width: 130px;
+			}
 			td:first-of-type {
 				background-color: var(--d2l-color-regolith);
 			}
+			td:last-of-type {
+				min-width: 0px;
```
dbatiste: For zero px, our standard practice is to just use 0.  i.e. unit-less.

d2l-rubric-criteria-group.html
```diff
@@ -10,9 +10,18 @@
...
+			}
+			td:last-of-type > div {
+				text-align: right;
+			}
 			th > div {
 				text-align: center;
 			}
```
dbatiste: I am surprised we need to override the typography below here.  Should our `d2l-table` styles be updated instead?

d2l-rubric-criteria-group.html
alogan519: ok thanks

d2l-rubric-criteria-group.html
alogan519: We plan to add in some aria text that will be something like "x out of y".

Our API will soon be returning the y.

d2l-rubric-criteria-group.html
alogan519: d2l-table style is left aligned, our design is looking for right alignment since it's a numerical column.

dbatiste: Ah... is was the `bold` part I was wondering about.  :)

d2l-rubric-criteria-group.html
dbatiste: Ah cool.

d2l-rubric-criteria-group.html
alogan519: Oh ok, ya we are looking to bold the header row which is not standard for table. I suppose I good target it with tr:first-of-type as well.

dbatiste: We run into lots of trouble with one-off typography styles when go to reskin things later, and trying to maintain some consistency.  Usually this sort of thing would be either built into `d2l-typography`, or `d2l-table`.

d2l-rubric-criteria-group.html
```diff
@@ -3,66 +3,76 @@
...
-							</th>
-						</template>
-						<th></th>
-					</tr>
-				</thead>
-				<tbody>
-					<template is="dom-repeat" items="[[criteriaEntities]]">
+		<div class="d2l-typography">
+			<d2l-table-wrapper>
+				<table class="d2l-table" summary="This table lists criteria and criteria group names in the first column. The first row lists level names and includes scores if the rubric uses a numeric scoring method">
```
kieranderson: This is not longer being localized

d2l-rubric-criteria-group.html
```diff
@@ -3,66 +3,76 @@
...
-										<div hidden=[[!_hasCustomPoints(criterionCell)]] class="custom-points">
-											[[localize('points', 'number', criterionCell.properties.points)]]
-										</div>
-										<span>[[_getDescription(criterionCell)]]</span>
+							<th></th>
+							<template is="dom-repeat" items="[[levels]]">
+								<th>
+									<div>
+										<label class="d2l-label-text">[[item.properties.name]]</label>
+										<div hidden="[[!item.properties.points]]">[[item.properties.points]] points</div>
```
kieranderson: This is also not being localized

d2l-rubric-criteria-group.html
```diff
@@ -3,66 +3,76 @@
...
-								<div class="levels-header">[[item.properties.name]]</div>
-								<div hidden="[[!item.properties.points]]">[[localize('points', 'number', item.properties.points)]]</div>
-							</th>
-						</template>
-						<th></th>
-					</tr>
-				</thead>
-				<tbody>
-					<template is="dom-repeat" items="[[criteriaEntities]]">
+		<div class="d2l-typography">
```
dbatiste: I think we should be able to count on the `d2l-typography` class being on the `body` element.  Was something not working with it there?

d2l-rubric.html
```diff
@@ -13,14 +14,15 @@
...
-		<style>
+		<style include="d2l-typography">
 			:host {
 				display: block;
 			}
 		</style>
-		<h1>[[title]]</h1>
-		<h2>[[description.properties.text]]</h2>
-		<d2l-rubric-criteria-groups href="[[_getHref(criteriaGroups)]]" token="[[token]]"></d2l-rubric-criteria-groups>
+		<div class="d2l-typography">
```
dbatiste: Same thing here - I think we only really need/want `d2l-typography` class on the body.

d2l-rubric-criteria-group.html
alogan519: Just because we are a component basically. I can try to add this to the body in the demo html that hosts it though and see if that works.

PR #22, opened 2018-02-07
Update versions of hm-constants and fetch-siren-entity-behavior

kieranderson:


d2l-rubric-criteria-group.html
```diff
@@ -103,7 +103,7 @@
 			},
 
 			behaviors: [
-				window.D2L.FetchSirenEntityBehavior,
+				window.D2L.PolymerBehaviors.FetchSirenEntityBehavior,
```
djwester: Why the addition of PolymerBehaviors here?

kieranderson: Was changed in the new version of Fetch-siren-entity

PR #24, opened 2018-02-09
Add group name

alogan519:
Use out of scores from API

### DIFF
--- a/d2l-rubric-criteria-group.html
+++ b/d2l-rubric-criteria-group.html
@@ -96,10 +96,6 @@
 				levels: {
 					type: Array
 				},
-				maxLevelScore: {
-					type: Number,
-					computed: '_getMaxLevelScore(levels)'
-				},
 				criteriaCollectionHref: {
 					type: Object,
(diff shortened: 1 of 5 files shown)


d2l-rubric-criteria-group.html
```diff
@@ -65,7 +69,7 @@
 								</td>
 							</template>
 							<td>
-								<div>/[[maxLevelScore]]</div>
+								<div>/[[item.properties.outOf]]</div>
```
djwester: Does this render ok?  When I was poking at this earlier I had to format it as a string so that the / and the points would render correctly

alogan519: Seemed to render ok on Chrome, I can check it out

djwester: Yeah seems fine for me know - I tried a while ago, so perhaps some of the additional formatting we've pulled in since then is taking care of it now

alogan519: renders OK in firefox as well

PR #27, opened 2018-02-13
Hybridize d2l-rubric to work with polymer 1 and 2

alogan519:


bower-1.x.json
```diff
@@ -0,0 +1,34 @@
+{
```
dbatiste: This file should not be checked in.  We should have `bower-1.x.json` in `.gitignore`.

package.json
```diff
@@ -7,7 +7,7 @@
     "test": "test"
   },
   "scripts": {
-    "postinstall": "bower install",
+    "postinstall": "polymer install --variants",
     "serve": "polymer analyze > analysis.json && polymer serve",
     "test": "npm run test:lint && npm run test:wct",
     "test:lint": "npm run test:lint:wc && npm run test:lint:html",
```
dbatiste: Do you know the reason for the `cross-env LAUNCHPAD_BROWSERS=chrome polymer test -p --skip-plugin sauce`?  Seems this is getting copied around, but I don't _think_ we need/want this.

dbatiste: Also, further below, we can remove the `bower` dev dependency, since `polymer install` takes care of that for us.

CodeBaboon: on Windows machines the tests will pause for minutes (like, 5+) before running while searching for browsers. `LAUNCHPAD_BROWSERS=chrome` hijacks that by just directing it to Chrome. The `cross-env` is necessary for setting the env-var in Windows.

dbatiste: Thanks Dave.  This makes me sad.   :)

alogan519:
Oh sorry guys checked this in for Stacey to help debug some things, will make sure to address concerns before going into master.

package.json
CodeBaboon: The saddest part is that it's been a [known issue](https://github.com/Polymer/web-component-tester/issues/169) in the WCT local browser detection for close to 3 years and has yet to be resolved.

dbatiste: Yeah.  Us Mac users just keep sailing along not realizing.  Thanks for clarifying.  :)

PR #30, opened 2018-02-23
Trigger an iron resize event so that d2l-table-wrapper resizes 

alogan519:
Trigger an iron resize event so that d2l-table-wrapper resizes and will draw the next / previous page buttons when the levels change.

### DIFF
--- a/d2l-rubric-criteria-group.html
+++ b/d2l-rubric-criteria-group.html
@@ -140,15 +140,14 @@
 			_onLevelsEntityChanged: function(levelsEntity) {
 				this._levels = levelsEntity.getSubEntitiesByClass(this.HypermediaClasses.rubrics.level);
 				// trigger a resize event so that the table resizes with the new levels
-				var isPolymer1 = !Polymer.Element;
-				if (isPolymer1) {
-					window.requestAnimationFrame(function() {
-						this.notifyResize();
-					}.bind(this));
-				} else {
+				if (Polymer.Element) {
 					Polymer.RenderStatus.beforeNextRender(this, function() {
 						this.notifyResize();
 					});
+				} else {
+					requestAnimationFrame(function() {
+						this.notifyResize();
+					}.bind(this));
 				}
 			},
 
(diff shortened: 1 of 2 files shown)


d2l-rubric-criteria-group.html
```diff
@@ -137,6 +139,17 @@
 
 			_onLevelsEntityChanged: function(levelsEntity) {
 				this._levels = levelsEntity.getSubEntitiesByClass(this.HypermediaClasses.rubrics.level);
+				// trigger a resize event so that the table resizes with the new levels
+				var isPolymer1 = !Polymer.Element;
+				if (isPolymer1) {
```
dbatiste: In other places where we've need to do the version check, we've just gone `if (Polymer.Element) { ... }`.

d2l-rubric-criteria-group.html
```diff
@@ -137,6 +139,17 @@
...
+					window.requestAnimationFrame(function() {
```
dbatiste: Nitpick: `window` is not actually needed here.

d2l-rubric-criteria-group.html
```diff
@@ -137,6 +139,17 @@
...
+						this.notifyResize();
+					}.bind(this));
+				} else {
+					Polymer.RenderStatus.beforeNextRender(this, function() {
+						this.notifyResize();
+					});
```
dbatiste: Note: for some reason we've had to still do `.bind(this)` when wiring up to these status events.  It might have been that eslint doesn't know that `this` is being bound under the covers.  Not sure... ?

d2l-rubric-criteria-group.html
alogan519: Ok, so you think I should flip the logic? Or just avoid the extra variable.

d2l-rubric-criteria-group.html
alogan519: Ya it passed through es lint, guessing because this is passed as the first param.

PR #32, opened 2018-02-28
Changes to get ie 11 to work, pull in polyfills for fetch and promise.

alogan519:
Use d2l-t* elements instead of table elements.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -33,13 +33,13 @@
     "d2l-localize-behavior": "^1.1.0",
     "d2l-typography": "^6.1.2",
     "polymer": "1 - 2",
-    "iron-resizable-behavior": "^2.1.0",
-    "fetch": "^2.0.3",
-    "promise-polyfill": "PolymerLabs/promise-polyfill#^2.0.1"
+    "iron-resizable-behavior": "^2.1.0"
   },
   "devDependencies": {
+    "fetch": "^2.0.3",
     "iron-component-page": "^2.0.0",
     "iron-demo-helpers": "^2.0.0",
+    "promise-polyfill": "PolymerLabs/promise-polyfill#^2.0.1",
     "web-component-tester": "^6.5.0"
   },
   "variants": {
--- a/bower.json
+++ b/bower.json
@@ -33,7 +33,9 @@
     "d2l-localize-behavior": "^1.1.0",
     "d2l-typography": "^6.1.2",
     "polymer": "1 - 2",
-    "iron-resizable-behavior": "^2.1.0"
+    "iron-resizable-behavior": "^2.1.0",
+    "fetch": "^2.0.3",
+    "promise-polyfill": "PolymerLabs/promise-polyfill#^2.0.1"
   },
   "devDependencies": {
     "iron-component-page": "^2.0.0",
--- a/bower.json
+++ b/bower.json
@@ -25,16 +25,16 @@
     "demo"
   ],
   "dependencies": {
-    "d2l-fetch-siren-entity-behavior": "Brightspace/d2l-fetch-siren-entity-behavior#^4.0.2",
-    "d2l-table": "BrightspaceUI/table#^1.2.0",
-    "d2l-fetch": "Brightspace/d2l-fetch#^1.8.0",
-    "d2l-hm-constants-behavior": "git://github.com/Brightspace/d2l-hm-constants-behavior.git#^5.1.0",
     "d2l-colors": "^3.1.2",
+    "d2l-fetch": "Brightspace/d2l-fetch#^1.8.0",
+    "d2l-fetch-siren-entity-behavior": "Brightspace/d2l-fetch-siren-entity-behavior#^4.0.2",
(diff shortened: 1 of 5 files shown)


d2l-rubric-criteria-group.html
```diff
@@ -35,48 +40,54 @@
 			}
 
 		</style>
-		<d2l-table-wrapper>
-			<table class="d2l-table" summary="[[localize('rubric_summary_a11y')]]">
-				<thead>
-					<tr>
-						<th>
+		<style type="text/css">
+			.btn {
```
djwester: What is this button styling for?

d2l-rubric-criteria-group.html
```diff
@@ -35,48 +40,54 @@
...
+				border-radius: 0;
+			}
+			table.table tbody tr {
+				transition: initial;
+			}
+		</style>
+		<d2l-table summary="[[localize('rubric_summary_a11y')]]">
```
awikkerink: Summary won't be read by screen readers. Maybe use `aria-describedby`?

d2l-rubric-criteria-group.html
```diff
@@ -35,48 +40,54 @@
 			}
 
 		</style>
-		<d2l-table-wrapper>
-			<table class="d2l-table" summary="[[localize('rubric_summary_a11y')]]">
-				<thead>
-					<tr>
-						<th>
+		<style type="text/css">
```
jstefaniuk-d2l: Do you need to have 2 `<style>` tags? Or can it all be moved to one?

d2l-rubric-criteria-group.html
alogan519: Thanks for catching this, forgot I was playing around with this.

d2l-rubric-criteria-group.html
djwester: Current rubric table uses summary, which is why we did this here. Aria is probably a more standard thing though, so probably worth changing

d2l-rubric.html
```diff
@@ -5,6 +5,9 @@
 <link rel="import" href="../d2l-hm-constants-behavior/d2l-hm-constants-behavior.html">
 <link rel="import" href="../d2l-typography/d2l-typography-shared-styles.html">
 <link rel="import" href="./localize-behavior.html">
+<link rel="import" href="../promise-polyfill/promise-polyfill.html">
+
+<script src="../fetch/fetch.js"></script>
```
dbatiste: We should be importing `d2l-fetch`, yes?

jstefaniuk-d2l: I may be wrong but I don't think d2l-fetch includes the polyfill

ryantmer: Oh, okay, spent another 14 seconds looking. This repo is already using `d2l-fetch` correctly, TIL. Generally I'd say it's up to the thing _using_ `d2l-rubric` (e.g. BSI) to include the polyfills, but maybe there's a case where that's not sufficient here?

alogan519: So one advantage of doing the polyfill here was that the polymer demo works on ie 11 now.

jstefaniuk-d2l: You may just want to put the import on the demo page in that case

ryantmer: Yeah, I'd suggest what @jstefaniuk-d2l said, as well as moving it to the `devDependencies` in that case :)

alogan519: ok, sounds good.

bower.json
```diff
@@ -33,11 +33,14 @@
     "d2l-localize-behavior": "^1.1.0",
     "d2l-typography": "^6.1.2",
     "polymer": "1 - 2",
-    "iron-resizable-behavior": "^2.1.0"
+    "iron-resizable-behavior": "^2.1.0",
+    "d2l-offscreen": "^3.0.3"
```
dbatiste: Nitpick: our general practice is to keep the deps in alpha order.  Just helps when searching through things sometimes.

awikkerink: Bower doesn't enforce alpha order when running `bower install`. It kinda sucks

dbatiste: Yeah, it's a minor manual adjustment when including dependencies.

awikkerink:
Maybe `role=grid` could be used on d2l-table to fix the aria errors?

awikkerink:
I think grid may be a more appropriate role anyway.

https://github.com/GoogleChrome/accessibility-developer-tools/blob/master/src/js/Constants.js doesn't define table as a valid role.

alogan519:
Yup switching to role grid fixes it, Anthony should I fix that here or make a change to d2l-table?

bower.json
alogan519: yup, I can reorder manually.

test/components/d2l-rubric-criteria-group.html
```diff
@@ -24,7 +24,7 @@
 			</template>
 		</test-fixture>
 		<script>
-			a11ySuite('basic-criteria');
-		  </script>
+			a11ySuite('basic-criteria', ['badAriaRole', 'ariaRoleNotScoped']);
```
awikkerink: Maybe put a `FIXME` here to note why these are disabled?

PR #33, opened 2018-03-05
Remove outof display for text only rubrics

alogan519:
Updates to match design closer
Add demo for text only rubrics

### DIFF
--- a/d2l-rubric.html
+++ b/d2l-rubric.html
@@ -34,11 +34,13 @@ Polymer Web-Component to display rubrics
 		</style>
 		<slot></slot>
 		<d2l-rubric-criteria-groups href="[[_getHref(_criteriaGroups)]]" token="[[token]]"></d2l-rubric-criteria-groups>
-		<div class="out-of-container">
-			<div class="out-of-text">
-				[[localize('total')]] &nbsp &nbsp / [[entity.properties.outOf]]
+		<template is="dom-if" if="[[_hasOutOf(entity)]]">
+			<div class="out-of-container">
+				<div class="out-of-text">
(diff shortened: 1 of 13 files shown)


d2l-rubric.html
```diff
@@ -34,11 +34,13 @@
 		</style>
 		<slot></slot>
 		<d2l-rubric-criteria-groups href="[[_getHref(_criteriaGroups)]]" token="[[token]]"></d2l-rubric-criteria-groups>
-		<div class="out-of-container">
-			<div class="out-of-text">
-				[[localize('total')]] &nbsp &nbsp / [[entity.properties.outOf]]
+		<template is="dom-if" if="[[_hasOutOf(entity)]]">
+			<div class="out-of-container">
```
jstefaniuk-d2l: Since there's not a lot to render here, something like this may be a good candidate for `hidden$=` rather than `dom-if`

awikkerink: `dom-if` is good for wrapping expensive elements, because it prevents the elements from being created until they are needed. But `hidden` is good enough and better performance (there is some extra stamping that goes on with `dom-if`)

alogan519: ok, maybe you guys can help out with what I was seeing. With using hidden I wasn't getting it to actually be hidden until entity was populated. So while it was doing the fetch entity the container was visible and then it would go hidden once the entity came down so it wasn't great visually.

awikkerink: Does `entity` have a default value? And does `_hasOutOf(entity)` return false if entity is null?

alogan519: Thanks, adding a default value to entity of null worked

PR #35, opened 2018-03-21
Create a mobile view which will be displayed when the screen is less …

alogan519:
…than 615px

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -28,13 +28,16 @@
     "d2l-colors": "^3.1.2",
     "d2l-fetch": "Brightspace/d2l-fetch#^1.8.0",
     "d2l-fetch-siren-entity-behavior": "Brightspace/d2l-fetch-siren-entity-behavior#^4.0.2",
-    "d2l-hm-constants-behavior": "git://github.com/Brightspace/d2l-hm-constants-behavior.git#^5.1.0",
+    "d2l-hm-constants-behavior": "git://github.com/Brightspace/d2l-hm-constants-behavior.git#^5.4.0",
+    "d2l-icons": "^4.10.0",
     "d2l-localize-behavior": "^1.1.0",
     "d2l-offscreen": "^3.0.3",
     "d2l-table": "BrightspaceUI/table#^1.2.0",
     "d2l-typography": "^6.1.2",
+    "iron-media-query": "^2.1.0",
     "iron-resizable-behavior": "^2.1.0",
-    "polymer": "1 - 2"
+    "polymer": "1 - 2",
+    "web-animations-js": "^2.3.1"
   },
   "devDependencies": {
     "fetch": "^2.0.3",
--- a/bower.json
+++ b/bower.json
@@ -36,8 +36,7 @@
     "d2l-typography": "^6.1.2",
     "iron-media-query": "^2.1.0",
     "iron-resizable-behavior": "^2.1.0",
-    "polymer": "1 - 2",
-    "web-animations-js": "^2.3.1"
+    "polymer": "1 - 2"
   },
   "devDependencies": {
     "fetch": "^2.0.3",
--- a/d2l-rubric-criterion-mobile.html
+++ b/d2l-rubric-criterion-mobile.html
@@ -1,12 +1,11 @@
 <link rel="import" href="../polymer/polymer.html">
-<link rel="import" href="../d2l-fetch-siren-entity-behavior/d2l-fetch-siren-entity-behavior.html">
 <link rel="import" href="../d2l-fetch/d2l-fetch.html">
 <link rel="import" href="../d2l-icons/d2l-icon-button.html">
 <link rel="import" href="../d2l-typography/d2l-typography-shared-styles.html">
-<link rel="import" href="../web-animations-js/web-animations.min.html">
 <link rel="import" href="./localize-behavior.html">
 <link rel="import" href="./d2l-rubric-levels-mobile.html">
 <link rel="import" href="../d2l-hm-constants-behavior/d2l-hm-constants-behavior.html">
+<link rel="import" href="./siren-entity-behavior.html">
 
 <dom-module id="d2l-rubric-criterion-mobile">
 	<template>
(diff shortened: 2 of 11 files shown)


d2l-rubric-criterion-mobile.html
```diff
@@ -0,0 +1,216 @@
...
+					return;
+				}
+				this._selectedCriterionText = criterionCells[selected].entities[0].properties.text;
+			},
+
+			_handleTapLeft: function() {
+				if (this._selected > 0) {
+					this._selected--;
+
+					this.$.animate.animate([
```
CodeBaboon: Is it necessary to use a library for this animation? Seems like it should just be a CSS class with transitions perhaps.

d2l-rubric-levels-mobile.html
```diff
@@ -0,0 +1,201 @@
...
+			<div class="levels">
+				<template is="dom-repeat" items="[[_levelEntities]]">
+					<div class$="[[_getLevelClassName(index, selected)]]" on-tap="_selectLevel" data-index="[[index]]">
+						<div class$="[[_getLevelTextClassName(index, selected)]]">
+							[[item.properties.name]]
+						</div>
+					</div>
+				</template>
+			</div>
+			<template is="dom-if" if="[[_hasOutOf(outOf)]]">
```
jstefaniuk-d2l: There isn't a lot to render inside this template, generally for something like this we'd use a `span` with a `hidden$=` attribute rather than a `dom-if` to avoid the perf hit caused by the `dom-if`

d2l-rubric-levels-mobile.html
```diff
@@ -0,0 +1,201 @@
...
+				<div class="level out-of">
+					<div class="level-name">
+						&mdash; / [[outOf]]
```
jstefaniuk-d2l: I think this should be a langterm

d2l-rubric-criterion-mobile.html
```diff
@@ -0,0 +1,216 @@
...
+						this._outOf = entity.properties.outOf;
+						this._criterionCells = entity.getSubEntitiesByClass(this.HypermediaClasses.rubrics.criterionCell);
+					}.bind(this));
+			},
+
+			_selectedCriterion: function(criterionCells, selected) {
+				if (!criterionCells) {
+					return;
+				}
+				this._selectedCriterionText = criterionCells[selected].entities[0].properties.text;
```
jstefaniuk-d2l: Will there only ever be 1 entity inside the criterionCell? Would it be worth doing a `getSubEntitiesByClass` here?

d2l-rubric-criterion-mobile.html
```diff
@@ -0,0 +1,216 @@
...
+				return selected === this._total - 1;
+			},
+
+			_getSelectedLevelText: function(levelTitle, selected, criterionCells) {
+				if (!levelTitle) {
+					return null;
+				}
+				// check for overrides
+				var points = this._selectedLevelPoints;
+				if (criterionCells && criterionCells[selected].hasClass(this.HypermediaClasses.rubrics.overridden)) {
```
jstefaniuk-d2l: will there ever be a case where `criterionCells[selected]` is undefined? (like perhaps if `criterionCells` had a length of 0?)

d2l-rubric-levels-mobile.html
alogan519: so I ended up doing this because the div inside has display:flex so it was overriding the display:none, not sure if there is another way? Adding an extra div messed with the layout.

d2l-rubric-criterion-mobile.html
alogan519: I can look at that, I think the library is just a polyfill for browsers that don't support https://developer.mozilla.org/en-US/docs/Web/API/Web_Animations_API

awikkerink:
I'm a bit hesitant to have 2 different "views" for different screen sizes. Could you maybe put some of the common properties/methods in behaviors?

d2l-rubric-criterion-mobile.html
alogan519: Ya that probably makes sense instead of the magic number here, there are 2 entities I think, one for feedback as well.

alogan519:
We talked about the 2 different views being quite different since one view is a table and one is a list so I'm open for ideas on how we might be able to make them more common.

Sure I can look at adding in some behaviors for a lot of the fetching of links / entities, some repetition for sure there.

d2l-rubric-criterion-mobile.html
alogan519: Hmm I think it would be good to guard against that case.

d2l-rubric-levels-mobile.html
jstefaniuk-d2l: Often times people throw:
```
[hidden] {
    display: none !important;
}
```
Into the css for the file, `!important` is often to be avoided but this is an appropriate use imo

alogan519: Ok thanks Jeff.

d2l-rubric-criterion-mobile.html
jstefaniuk-d2l: Transitions and css keyframe animation both have pretty good support I think
Transitions:
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Transitions/Using_CSS_transitions

Css keyframes and how to use them:
https://developer.mozilla.org/en-US/docs/Web/CSS/@keyframes
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Animations/Tips

d2l-rubric-levels-mobile.html
jstefaniuk-d2l: @awikkerink was that comment meant for this line: https://github.com/Brightspace/d2l-rubric/pull/35/files/99e036e7909a562435c08d6af67549a04e77bc1c#diff-fb4531b1a4ff20a99bc5044abd6bd8f0R19 ?

awikkerink: Ah yes. I got mixed up. Sorry

alogan519: @awikkerink basically the media query is driving a dom-if statement so it's populating a property, not sure how to do this in css or why it's better?

+		<iron-media-query query="(min-width: 615px)" query-matches="{{_largeScreen}}"></iron-media-query>
+		<template is="dom-if" if="[[_largeScreen]]">

d2l-rubric-criterion-mobile.html
dbatiste: Agree, it's possible to do a lot with CSS animations and keyframes.  I'd avoid doing animations in JavaScript unless absolutely necessary, and even then I'd question the animation design.

d2l-rubric-levels-mobile.html
awikkerink: You can do media-queries in css and hide/show the elements via CSS. But the `dom-if` makes sense here to prevent the components from getting loaded.

bower.json
```diff
@@ -28,13 +28,16 @@
...
+    "d2l-icons": "^4.10.0",
     "d2l-localize-behavior": "^1.1.0",
     "d2l-offscreen": "^3.0.3",
     "d2l-table": "BrightspaceUI/table#^1.2.0",
     "d2l-typography": "^6.1.2",
+    "iron-media-query": "^2.1.0",
     "iron-resizable-behavior": "^2.1.0",
-    "polymer": "1 - 2"
+    "polymer": "1 - 2",
+    "web-animations-js": "^2.3.1"
```
dbatiste: So hopefully we don't need this.

d2l-rubric-criterion-mobile.html
```diff
@@ -0,0 +1,216 @@
...
+
+<dom-module id="d2l-rubric-criterion-mobile">
+	<template>
+		<style>
+			:host {
+				display: block;
+			}
+			.criterion-name {
+				@apply --d2l-body-standard-text;
+				margin-top: 18px;
```
dbatiste: Many times this sort of whitespace should be `rem`.  Designers don't mention it because they don't realize (or they forget) that we have variable root font-size depending on what the user picks in Account Settings.  Anytime we want proportion of whitespace to text to remain the same, we should be using `rem`.  Anyway, something to consider here (and below).

dbatiste:
Probably can safely add [strip-whitespace](https://www.polymer-project.org/1.0/docs/devguide/local-dom#strip-whitespace) to templates.  Minor perf improvement, but also can avoid some frustration with pesky empty text nodes.

pbrown-d2l:
We will have some things to modify afterwards for Mobile view, but this will let us start showing people

PR #42, opened 2018-04-04
Render the html content instead of the plain text

alogan519:


jstefaniuk-d2l:
I'd do some testing to make sure that a user can't put things in the html that break your formatting

stuff like this:
```
<div style="height:100px; width:10000px; background:red;"></div>
```

In parent-portal we added an overflow-x to the container to fix that case

jstefaniuk-d2l:
When this component eventually gets put into parent-portal, we'll need to swap out the `s-html` with a user-defined `d2l-richtext` component, and the HM will need to be changed to run the html through the html annotator (like this: http://search.dev.d2l/source/xref/Lms/core/le/dropbox/D2L.LE.Dropbox.Extensibility/HM/Folders/Serializers/Siren/SirenFoldersSerializer.cs#136) so that we can gracefully handle things parents can't access like quicklinks

PR #46, opened 2018-04-06
Sdharme/mobile accessibility and ui cleanup

d2l-sdharme:
- Adding keyboard accessibility to mobile view for levels
- Reduce criteria group font size and bold
- Fix space between criteria groups
- Modify langterm to show `/ 4` instead of `- / 4` and reduce `out-of` div width

### DIFF
--- a/d2l-rubric-criteria-group.html
+++ b/d2l-rubric-criteria-group.html
@@ -36,6 +36,8 @@
 			}
 			.group-name {
 				@apply --d2l-body-standard-text;
+				font-size: 16px;
+				font-weight: bold;
 				text-align: left;
 				background-color: var(--d2l-table-header-background-color);
 			}
@@ -101,7 +103,7 @@
 						</template>
 						<template is="dom-if" if="[[_hasOutOf(entity)]]">
 							<d2l-td class="out-of">
-								<div>/[[item.properties.outOf]]</div>
+								<div> [[localize('outOf', 'outOf', item.properties.outOf)]] </div>
 							</d2l-td>
 						</template>
 					</d2l-tr>
--- a/d2l-rubric-criteria-group.html
+++ b/d2l-rubric-criteria-group.html
@@ -35,8 +35,7 @@
 				font-weight: 700;
 			}
 			.group-name {
-				@apply --d2l-body-standard-text;
-				font-size: 16px;
+				@apply --d2l-body-compact-text;
 				font-weight: bold;
 				text-align: left;
 				background-color: var(--d2l-table-header-background-color);
--- a/d2l-rubric-levels-mobile.html
+++ b/d2l-rubric-levels-mobile.html
@@ -28,14 +28,21 @@
 				overflow: hidden;
 				text-align: center;
 				color: var(--d2l-color-galena);
+				outline: none;				
 			}
 			.level:hover {
 				cursor: pointer;
+				background-color: var(--d2l-color-gypsum);
 			}
 			.level.selected {
 				background-color: var(--d2l-color-regolith);
 				border: solid 1px var(--d2l-color-galena);
 				box-shadow: inset 0 1px 3px 0 rgba(0, 0, 0, 0.1);
+			}
+			.level:focus, .level.selected:focus {
+				background-color: var(--d2l-color-gypsum);
+				box-shadow: 0 0 0 4px rgba(0, 111, 191, 0.4);
+				z-index: 0;
 			}
 			.level.selected + .level {
 				border-left: none;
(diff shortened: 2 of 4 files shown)


d2l-rubric-criteria-group.html
```diff
@@ -36,6 +36,8 @@
 			}
 			.group-name {
 				@apply --d2l-body-standard-text;
+				font-size: 16px;
```
dbatiste: Are we able to use the [--d2l-body-compact-text](https://github.com/BrightspaceUI/typography/blob/master/d2l-typography-shared-styles.html) instead, or does the line-height mess things up.  At minimum we should be using `rem`... so `0.8rem`, but ideally we use the appropriate typography mixin.  :)

d2l-rubric-levels-mobile.html
```diff
@@ -28,15 +28,22 @@
...
 			}
 			.level.selected {
 				background-color: var(--d2l-color-regolith);
 				border: solid 1px var(--d2l-color-galena);
 				box-shadow: inset 0 1px 3px 0 rgba(0, 0, 0, 0.1);
 			}
+			.level:focus, .level.selected:focus {
+				background-color: var(--d2l-color-gypsum);
+				box-shadow: 0 0 0 4px rgba(0, 111, 191, 0.4);
+				z-index: 0;
```
dbatiste: What purpose does the `z-index` serve here?

d2l-rubric-levels-mobile.html
```diff
@@ -76,15 +83,16 @@
...
 				cursor: default;
+				background-color: transparent;
 			}
 			[hidden] {
 				display: none !important;
 			}
 		</style>
 		<div class="levels-container">
-			<div class="levels">
+			<div class="levels" role="tablist">
```
dbatiste: Looks like we're missing some stuff for tab implementation.  See [W3C guide](https://www.w3.org/TR/wai-aria-practices/#tabpanel#x3-21-tabs) for implementation of roles, corresponding properties, and keyboard interaction.  I might have a sample I can dig up from last year's ARIA workshop if it helps.  

I'm sort of assuming "tabs" makes semantic sense - @pbrown-d2l , @alogan519 , and I briefly chatted about it a day or two ago and general feeling was that it does make sense, but I don't know enough about the context to have a strong opinion.

d2l-rubric-levels-mobile.html
```diff
@@ -156,6 +164,26 @@
 				'_getSelectedProperties(_levelEntities, selected)'
 			],
 
+			_onKeyDown: function (event) {
+				switch (event.keyCode) {
+					case 13:
```
dbatiste: While technically not needed, usually I put the keycode values in a little private object and refer to them by name.  Makes it easier to understand later, or for others.  :)

d2l-rubric-levels-mobile.html
alogan519: Looks like the role tablist should go on the container above this one, and we should add tabpanel role in the criterion description.

Keyboard interaction seemed ok from what I saw.

d2l-rubric-criteria-group.html
d2l-sdharme: Using `--d2l-body-compact-text` works

d2l-rubric-levels-mobile.html
d2l-sdharme: Without this, if a level is focused on, and a level directly beside that level is hovered upon, the hover background of the hovered level would go in front of box-shadow of the focused level.

d2l-rubric-levels-mobile.html
dbatiste: I _think_ we may be missing some aria-controls wire-up, and selected state properties.

d2l-rubric-levels-mobile.html
```diff
@@ -76,15 +83,16 @@
...
 			[hidden] {
 				display: none !important;
 			}
 		</style>
 		<div class="levels-container">
-			<div class="levels">
+			<div class="levels" role="tablist">
 				<template is="dom-repeat" items="[[_levelEntities]]">
-					<div class$="[[_getLevelClassName(index, selected)]]" on-tap="_selectLevel" data-index="[[index]]">
+					<div class$="[[_getLevelClassName(index, selected)]]" on-tap="_selectLevel" data-index="[[index]]" role="tab" on-keydown="_onKeyDown" tabindex="0" aria-selected$="{{_isSelected(index, selected)}}">
```
alogan519: _isSelected can just be a one way binding, [[]] instead of {{}}

Can you put each param on its own line, this line is getting long.

d2l-rubric-levels-mobile.html
```diff
@@ -156,6 +164,36 @@
 				'_getSelectedProperties(_levelEntities, selected)'
 			],
 
+			_keyCodes: {
+				ENTER: 13,
+				LEFT: 37,
+				RIGHT: 39
+			},
+
+			_isSelected: function(index, selected) {
```
alogan519: Can we keep together the _keyCodes and _onKeyDown defintions, maybe move this above or below them.

PR #48, opened 2018-04-10
Add in initial translations, will have to fix the pluralization of po…

alogan519:
…ints when that is ready

### DIFF
--- a/lang/ar.html
+++ b/lang/ar.html
@@ -1,0 +1,24 @@
+<link rel="import" href="../../polymer/polymer.html">
+
+<script>
+	window.D2L = window.D2L || {};
+	window.D2L.PolymerBehaviors = window.D2L.PolymerBehaviors || {};
+	window.D2L.PolymerBehaviors.Rubric = window.D2L.PolymerBehaviors.Rubric || {};
+	window.D2L.PolymerBehaviors.Rubric.LocalizeBehavior = window.D2L.PolymerBehaviors.Rubric.LocalizeBehavior || {};
+
+	/*
+	* Ar lang terms
+	* @polymerBehavior D2L.PolymerBehaviors.Rubric.LocalizeBehavior.LangArBehavior
+	 */
+	D2L.PolymerBehaviors.Rubric.LocalizeBehavior.LangArBehavior = {
+		ar: {
+			'levelNameAndPoints': '{levelName} - {number} {number, plural, one {point} other {points}}',
+			'levelNameAndPercentage': '{levelName} - {number} %',
+			'outOf': '/\u00A0{outOf}',
+			'points': '{number} {number, plural, one {point} other {points}}',
+			'percentage': '{number} %',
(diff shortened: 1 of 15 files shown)


lang/ar.html
```diff
@@ -0,0 +1,24 @@
...
+	 */
+	D2L.PolymerBehaviors.Rubric.LocalizeBehavior.LangArBehavior = {
+		ar: {
+			'levelNameAndPoints': '{levelName} - {number} {number, plural, one {point} other {points}}',
+			'levelNameAndPercentage': '{levelName} - {number} %',
+			'outOf': '/\u00A0{outOf}',
+			'points': '{number} {number, plural, one {point} other {points}}',
+			'percentage': '{number} %',
+			'rubricSummaryA11y': 'يسرد هذا الجدول أسماء المعايير وأسماء مجموعات المعايير في العمود الأول. يسرد الصف الأول أسماء المستويات ويحتوي مجموع الدرجات إذا كانت آلية التقييم تستخدم طريقة رقمية لتسجيل النقاط.',
+			'totalOutOf': 'المجموع     / {outOf}'
```
jstefaniuk-d2l: Will html even render multiple spaces in a row?

lang/de.html
```diff
@@ -18,7 +18,7 @@
 			'points': '{number} {number, plural, one {point} other {points}}',
 			'percentage': '{number} %',
 			'rubricSummaryA11y': 'In der ersten Spalte dieser Tabelle wird der Name der Kriterien und Kriteriengruppen aufgeführt. Die erste Zeile enthält die Namen der Niveaus sowie Punktzahlen, sofern für das Bewertungsschema eine numerische Bewertungsmethode verwendet wird.',
-			'totalOutOf': 'Insgesamt     / {outOf}'
+			'totalOutOf': 'Insgesamt \u00A0 \u00A0 / {outOf}'
```
CodeBaboon: Why use non-breaking spaces and not multiple elements + margins or padding or something? This feels weird

alogan519: Ok ya I was looking to give some flexibility in translation but I guess it's more part of the design.

dbatiste: Interesting.  That flexibility might be needed for certain translations, but I am not sure.

alogan519: So would you just have this be Total then and do the / {outOf} in code?

CodeBaboon: We have a bunch of logic around displaying these values, I think it's mostly in here: https://github.com/Brightspace/d2l-activity-details-ui/pull/46/ but @jstefaniuk-d2l can confirm.

This makes me think we should break this out into some kind of grade display web component that can be shared.

alogan519: Ya it's similar, the design was meant to be basically a space that a number would eventually go there. Some alignment might be helpful.
Talked with our designer and we will switch from the 2 spaces to a dash.

PR #49, opened 2018-04-11
TA124669 - added d2l-alert

pbrown-d2l:


d2l-rubric.html
```diff
@@ -41,9 +47,9 @@
 			}
 		</style>
 		<slot></slot>
-		<d2l-rubric-loading hidden$="[[_showContent]]"></d2l-rubric-loading>
-		<div hidden$="[[_showContent]]" class="out-of-loader"></div>
-		<div hidden$="[[!_showContent]]">
+		<d2l-rubric-loading hidden$="[[_hideLoading(_showContent,_hasAlerts)]]"></d2l-rubric-loading>
+		<div hidden$="[[_hideLoading(_showContent,_hasAlerts)]]" class="out-of-loader"></div>
```
djwester: Is the naming of  _hideLoading and _hideOutOf a bit confusing?  In some ways they return 'opposites' of each other.  Should we name these differently?

siren-entity-behavior.html
```diff
@@ -40,11 +50,40 @@
...
+			).then(
+				function(entity) {
 					this.entity = entity;
-				}.bind(this));
+				}.bind(this)
+			).catch(
+				function() {
+					this._addAlert(
+						'error',
+						'failedToLoadRubric',
```
djwester: Does the Alert name not need to be translated?

d2l-rubric.html
pbrown-d2l: I think they are good, they are asking whether something should be hidden. If true, the element is hidden.

siren-entity-behavior.html
pbrown-d2l: not the alert name. But the message does

d2l-rubric.html
pbrown-d2l: It was getting confusing at it was written before because it was using showcontent and negated showcontent to make decisions. This will help devs new to this code see what's going on

d2l-rubric.html
```diff
@@ -16,6 +17,11 @@
 
 <dom-module id="d2l-rubric">
 	<template strip-whitespace>
+		<template is="dom-repeat" items="[[_alerts]]">
```
alogan519: This should go under the <style> tag I think.

lang/en.html
```diff
@@ -18,6 +18,7 @@
 			'points': '{number} {number, plural, one {point} other {points}}',
 			'percentage': '{number} %',
 			'rubricSummaryA11y': 'This table lists criteria and criteria group name in the first column. The first row lists level names and includes scores if the rubric uses a numeric scoring method.',
+			'rubricLoadingErrorMessage': 'Sorry, we were unable to display the rubric.',
```
alogan519: Should actually go above rubricSummaryA11y :)

I can move it as part of my PR as I have a conflict here anyway. I'll have to test this in another language to see if it handles it missing from the other language files.

PR #51, opened 2018-04-12
add aria roles for tab panels and add aria-controls and controlled by

alogan519:


d2l-rubric-criterion-mobile.html
```diff
@@ -241,8 +242,8 @@
 				}
 			},
 
-			_handledDescriptionAnimation: function(animation) {
-				var element = this.$.animate;
+			_handledDescriptionAnimation: function(animation, selected) {
+				var element = this.shadowRoot.querySelector('#' + 'level-description-panel' + selected);
```
alogan519: ah, this doesn't work in polymer 1

d2l-rubric-criterion-mobile.html
```diff
@@ -241,8 +242,8 @@
 				}
 			},
 
-			_handledDescriptionAnimation: function(animation) {
-				var element = this.$.animate;
+			_handledDescriptionAnimation: function(animation, selected) {
+				var element = this.$.description.querySelector('#' + 'level-description-panel' + selected);
 				element.classList.remove('slide-from-right');
 				element.classList.remove('slide-from-left');
 				void element.offsetWidth; // triggers a reflow
```
dbatiste: Forced reflow is usually undesirable for per reasons.  I am guessing this is to ensure the page is updated before the `animation` class is added?  If possible it would be better to use wrap that in `requestAnimationFrame`...
```
requestAnimationFrame(function() {
     element.classList.add(animation);
})
```

Even better would be to use fast-dom (we have a wrapper for it) for all DOM mutation and measurements to avoid possible forced reflows.  Let me know if you want more info or examples on how to use it.

d2l-rubric-criterion-mobile.html
```diff
@@ -116,6 +116,9 @@
 			.level-name {
 				font-weight: 700;
 			}
+			[aria-hidden], [hidden] {
```
jstefaniuk-d2l: Setting `[aria-hidden]` to `display: none` seems odd to me since just using normal `hidden` will hide thing from screen readers as well

d2l-rubric-criterion-mobile.html
```diff
@@ -124,29 +127,37 @@
...
-			<div class="criterion-middle" id="animate">
-				<div class="level-name">
-					[[_getSelectedLevelText(_selectedLevelName, _selected, _criterionCells)]]
+			<template is="dom-repeat" items="[[_criterionCells]]" as="criterionCell" indexAs="index">
+				<div
+					id="level-description-panel[[index]]"
+					class="criterion-middle"
+					aria-labelledby$="level-tab[[index]]"
+					role="tabpanel"
+					aria-hidden$="[[!_isLevelSelected(index, _selected)]]"
```
jstefaniuk-d2l: As I mentioned above, if you're already using `hidden` here, aria-hidden is redundant

d2l-rubric-criterion-mobile.html
```diff
@@ -214,19 +218,16 @@
 				this._criterionCells = entity.getSubEntitiesByClass(this.HypermediaClasses.rubrics.criterionCell);
 			},
 
-			_selectedCriterion: function(criterionCells, selected) {
-				if (!criterionCells) {
-					return;
-				}
-				this._selectedCriterionText = criterionCells[selected].getSubEntityByClass(this.HypermediaClasses.text.description).properties.html;
+			_hasDescription: function(criterionCell) {
+				return !!criterionCell.getSubEntityByClass(this.HypermediaClasses.text.description).properties.html;
```
jstefaniuk-d2l: Will this subentity ever be missing?

alogan519: No I don't think so

d2l-rubric-criterion-mobile.html
alogan519: right, thanks Jeff

d2l-rubric-criterion-mobile.html
alogan519: Ok thanks Dave, great catch. This was because we basically kept sharing the same element and swapping out the text but now we have different elements for each level description so this doesn't seem needed.

PR #52, opened 2018-04-12
Fix RTL reading for total div

d2l-sdharme:


d2l-sdharme:
Rebased branch with new master branch

PR #53, opened 2018-04-24
fixing langterms

pbrown-d2l:


lang/fr.html
```diff
@@ -18,7 +18,7 @@
 			'outOf': '/\u00A0{outOf}',
 			'points': '{number} {number, plural, one {point} other {points}}',
```
djwester: Point/Points isn't in here in French?

pbrown-d2l: That is the translation

djwester: oh ok, goes to show how well I learned my French :-p

PR #55, opened 2018-04-27
Djwester/update tests

djwester:
The test for the text only stuff were 'passing' but that is only because the expects weren't running.  I could give it any data (or no data) and would still work.

I fixed a couple of things 
1) for some reason flush just won't work - I tried (and tried) and it just doesn't seem to work for me.  The only way I could get it to work was with an explicit timeout call (not nice I know so if anyone else wants to take a shot at it, please do!)
2) The path to elements inside the component is different if the shadow dom is used, so I had to put a condition for that
3) we can't use the data from the demo/data area since we are using relative paths and those paths will resolved to absolute paths that are relative to the test directory and hence the files won't be found. I have added in a copy call to the package.json to copy the files into the test directory before running the tests.  Seem comments on that file as we need to decide if we want to do it that way or not.

I also update the data we are using to reflect that way the api is currently set up.


package.json
```diff
@@ -15,7 +15,8 @@
     "test:lint:wc": "polymer lint",
     "test:local": "npm run test:lint && npm run test:wct:local",
     "test:wct": "polymer test -p",
-    "test:wct:local": "cross-env LAUNCHPAD_BROWSERS=chrome polymer test -p --skip-plugin sauce"
+    "copy:windows":"xcopy demo\\data test\\components\\data /E /I /Y",
```
djwester: This command will copy the files from demo/data into the test directory so that they are available for the tests to use.  The other thing we could do is just duplicate the data and have it in both places.

If we do go this way there is one obvious thing  - we might want to create commands for linux devs to use 

The other thing we will need to do is to update travis.yml to include a copy command
- cp -rf demo/data tests/components/data

Without that command (as you can see on the CI run for the push), the test will fail since there won't be any data.

So, discussion time:  Do we want to duplicate the data, or do the required cleanup to make this copy stuff work correctly?

test/components/d2l-rubric-text-only.js
```diff
@@ -17,12 +23,15 @@ suite('<d2l-rubric-text-only>', function() {
 
 	suite('Text Only Rubric', function() {
 		test('Out of container is hidden if using a text only rubric', function(done) {
-			var outOfContainer;
-			flush(function() {
-				outOfContainer = Polymer.dom(myElement.root).querySelector('.out-of-container');
+			myAsyncFunction(function(myElement){
```
djwester: If anyone has ideas on how to make this work without an explicit wait, please let me know.  As I understand it, flush is supposed to do that, but I can't get it to work

PR #56, opened 2018-04-27
Djwester/regenerate data

djwester:
a script that can be used to pull data from rubrics that have been defined on the QA sites.  It pull in the data via the hypermedia api and then changes the link paths to be relative paths and writes the data into the correct file structure based on the file structure indicated by the the api calls.

This allows us to easily regenerate the data if the api format changes slightly or if we want to add different kinds of rubrics to the demos or tests.

I've wrapped the python in js and then setup a command in the package.json so that we can just run this with 
>npm run regenerate_data <qa site url> <daily password>


package.json
```diff
@@ -35,5 +36,8 @@
     "frau-ci": "^1.33.2",
     "polymer-cli": "^1.5.7",
     "web-component-tester-istanbul": "^0.10.0"
+  },
+  "dependencies": {
+    "python-shell": "^0.5.0"
```
alogan519: should be in dev dependencies I think

PR #58, opened 2018-05-03
Initial version of d2l-rubric-editor

mdgbayly:
Currently just supports rendering a text only analytic rubric with autogrow text areas for entering criterion names, descriptions and feedback.

All the changes are additive with the exception of including the editor components in the test, demo, and doc pages.

Includes d2l-textarea autogrow text areas for editing criteria names, descriptions and feedback.
The d2l-textarea is currently being pulled from a public mdgbayly repository. We are working with Dave B and Dave L to get this promoted to a BrightspaceUI component.

Travis builds are failing because cannot run sauce tests from a fork. In future we will use branches.


editor/d2l-rubric-editor.html
```diff
@@ -0,0 +1,140 @@
...
+			<d2l-alert type="[[item.alertType]]">
+				[[item.alertMessage]]
+			</d2l-alert>
+		</template>
+		<slot></slot>
+		<d2l-rubric-loading hidden$="[[_hideLoading(_showContent,_hasAlerts)]]"></d2l-rubric-loading>
+		<div hidden$="[[_hideLoading(_showContent,_hasAlerts)]]" class="out-of-loader"></div>
+		<div hidden$="[[_hideOutOf(_showContent,_hasAlerts)]]">
+			<d2l-rubric-criteria-groups-editor href="[[_getHref(_criteriaGroups)]]" token="[[token]]"></d2l-rubric-criteria-groups-editor>
+			<div class="out-of-container" hidden="[[!_hasOutOf(entity)]]">
```
wongvincent: Should the `hidden` here be an attribute?

editor/d2l-rubric-criteria-group-editor.html
```diff
@@ -0,0 +1,356 @@
...
+			},
+
+			_localizeLevelOutOf: function(type, points) {
+				if (points === undefined || points === null) {
+					return;
+				}
+				return this.localize(type, 'number', points.toString());
+			},
+
+			_getRowCount: function(criteria) {
```
alogan519: Looks like a lot of these helper methods were copied over and aren't used, maybe check for and remove unused methods.

editor/d2l-rubric-criteria-group-editor.html
```diff
@@ -0,0 +1,356 @@
...
+
+		<d2l-rubric-loading hidden$="[[_showContent]]"></d2l-rubric-loading>
+
+		<div class="criteria-group">
+			<div class="row header">
+				<div class="cell col-first"></div>
+				<template is="dom-repeat" items="[[_levels]]" as="level">
+					<div class="cell"><d2l-text-input value="[[level.properties.name]]"></d2l-text-input></div>
+				</template>
+				<template is="dom-if" if="[[_hasOutOf(entity)]]">
```
alogan519: We have been trying to use <div hidden=> instead of dom-if as it's less expensive computation wise

editor/d2l-rubric-criteria-group-editor.html
```diff
@@ -0,0 +1,356 @@
...
+							</template>
+						</div>
+					</div>
+					<template is="dom-if" if="[[_hasOutOf(entity)]]">
+						<div class="cell col-last points">/ <d2l-text-input value=""></d2l-text-input></div>
+					</template>
+				</div>
+			</template>
+			<div class="row footer">
+				<div class="cell">+ ADD CRITERION</div>
```
alogan519: Checkout the localize-behaviors, should add this in there and call localize

Can just add it in english everywhere for now until you can translate it.

editor/d2l-rubric-editor.html
```diff
@@ -0,0 +1,140 @@
+<link rel="import" href="../../polymer/polymer.html">
```
alogan519: We were talking a bit about the idea of just having a mode for d2l-rubric instead of making a new top level component. Initial thoughts are like rubric and rubric groups are shared by all of our "modes" and then the rubric group component is custom - so like right now we have desktop-preview and mobile-preview and the d2l-rubric-groups component has the logic what to display. So in this case d2l-rubric would take a mode like create, edit, preview and pass that down to d2l-rubric-groups and then that component would pick which group component is rendered with a dom-if.

Thoughts?

Would be easier to add things to d2l-rubric and d2l-rubric-groups but then you would have to test all the modes to make sure you didn't break anything. Or having separate top level components could require more duplication but might be easier to follow and test.

editor/d2l-rubric-criteria-group-editor.html
mdgbayly: The focus for this PR was mostly on getting an initial rendering working with editable autogrow text areas. I anticipate we'll look to refactor some of these helper methods, as many of them are utilized by both the viewer and the editor. We also have expectations for a need to add some kind of entity store once we allow actually editing/save too, so was deferring any of that refactoring work, but I can remove anything we are not using right now.

components.html
```diff
@@ -0,0 +1,2 @@
+<link rel="import" href="./d2l-rubric.html">
```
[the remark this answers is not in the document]
mdgbayly: It's used in the index.html page so that the `iron-component-page` can pull editor components into the doc pages. Because the editor components are not imported by the d2l-rubric.html

editor/d2l-rubric-editor.html
mdgbayly: That's an interesting idea. My preference would be to keep the separate for now to keep things simpler and to allow us to rev our different designs independently. We have a lot of additional settings outside of the rubric groups that we need to allow editing for. It is still an open question as to whether they will go into d2l-rubric-editor, whether we'll try and use regular MVC controls (unlikely) or whether these will go into some other higher level `page` component.   So for now until we iron those types of questions out and get further along with our implementation, I think it will be simpler to keep things separate. As things become more solidified we can look for opportunities to abstract.

image

PR #59, opened 2018-05-04
Added feedback row for rubrics

walinga:


d2l-rubric-criteria-group.html
```diff
@@ -109,7 +130,17 @@
 								<div> [[localize('outOf', 'outOf', item.properties.outOf)]] </div>
 							</d2l-td>
 						</template>
-					</d2l-tr>
+                   </d2l-tr>
+					<template is="dom-if" if="[[_hasFeedback(item)]]">
```
alogan519: can you do this as a <div hidden="[[!_hasFeedback(item)]]>, it's less expensive then dom-if.

d2l-rubric-criteria-group.html
```diff
@@ -59,6 +59,27 @@
 				display: block;
 				overflow-x: hidden;
 			}
+
+            .feedback-arrow {
+               margin-top: -45px;
+               margin-left: 20px;
+               width: 0;
+               height: 0;
+               border-top: 12px solid var(--d2l-table-header-background-color);
```
jstefaniuk-d2l: This block can probably be shortened as
```
border: 12px solid var(--d2l-table-header-background-color);
border-bottom-color: lightgray;
```

Also should `lightgray` be replaced with a d2l color var?

d2l-rubric-criteria-group.html
walinga: Unfortunately, I couldn't find a way to do that without breaking the whole table

PR #60, opened 2018-05-04
Dwesterveld/read assessments data

djwester:


djwester:
Refactored the api updating script a bit so I could use it to get an cleanup some of the assessments stuff.

The get_assessments script is pretty dirty for now, but should help if we want to try out some additional combinations in the demo

PR #66, opened 2018-05-10
Refactor rubric edit components to utilize an entity-store.

mdgbayly:
Refactor the edit components to utilize an entity store rather than making each component responsible for fetching it's data.

Includes refactoring the criteria-group-editor to split up the levels and criterion into more fine grained components that are all resource based (take an href and token). This makes it simpler to manage the fetching and invalidation of different resources during the edit operations that we will be implementing.

The entity store supports expanding embedded entities and storing them independently in the store keyed by their ID (resource URL), to enable the resource components to render independently fetched resources and resources fetched in the context of more coarse grained API responses.

This code is currently using a copy of the entity-store from the inspirations-parallel-lms repo, enhanced to utilize `d2l-siren-fetch-entity-behavior`, and to support for expanding embedded entities.

As the components in the `polymer-siren-mixins` repo become more stable we will plan to refactor to utilize the shared entity store.


test/components/entity-store.js
```diff
@@ -0,0 +1,64 @@
...
+
+	suite('smoke test', function() {
+
+		test('can fetch leaf entity', function(done) {
+			window.D2L.Rubric.EntityStore.addListener(
+				'data/rubrics/organizations/text-only/199/groups/176/criteria/623/0.json',
+				'foozleberries',
+				function(entity) {
+					var description = entity && entity.getSubEntityByClass('description').properties.html;
+					expect(description).to.equal('Proper use of grammar');
```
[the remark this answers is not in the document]
mdgbayly: ah - ok - thanks - was planning on asking how the new regen stuff was intended to be used!

store/entity-editor-behavior.html
```diff
@@ -0,0 +1,124 @@
+<script>
```
alogan519: we should probably switch over to use this as well

mdgbayly: it is still a little bit experimental so maybe let us run with it for a week or so - maybe once we have a full end to end edit workflow working we can see if it makes sense to generalize it to the viewer too - don't want to break your stuff which is expected to be a bit more stable than the edit use cases right now.

PR #67, opened 2018-05-10
Some accessibility improvements

walinga:


d2l-rubric-criteria-group.html
```diff
@@ -268,11 +265,12 @@
...
 			},
 
-			_getRowCount: function(criteria) {
+			_getRowCount: function(criteria, assessmentResult) {
 				if (!criteria) {
 					return 0;
 				}
-				return criteria.length + 1; // criteria + levels row
+				var feedbackCount = this._getFeedbackCount(criteria, assessmentResult);
+				return criteria.length + feedbackCount + 1; // criteria + feedback + levels row
```
djwester: I wonder if it would be ok to not count the feedback rows when reporting the table size?  

The aria-owns tag makes them get read almost as part of the criteria row they belong to (i.e. the tab order is: Criteria Name, Feedback, criteria cells).  I feel like it doesn't make the experience that much better to say at table has 7 rows instead of 4 rows if we have 3 'rows' of feedback in there.

If we get rid of this we can get rid of the _getFeedbackCount funciton as well and we'll see some ever so minor performance improvements

djwester:
There are still a few oddities, but I don't know if we can do much better than this.

d2l-rubric-criteria-group.html
walinga: I agree. In my latest commit, I moved feedback to an imaginary 7th column for each row, but I'm not sure if that's any better than what we had

PR #70, opened 2018-05-11
Fix bug causing empty feedback box to appear

walinga:


assessment-result-behavior.html
```diff
@@ -33,9 +33,10 @@
 				if (selectedCriterionCellEntity) {
 					var link = selectedCriterionCellEntity.getLinkByRel(this.HypermediaRels.Rubrics.criterionCell);
 					var rowLink = selectedCriterionCellEntity.getLinkByRel(this.HypermediaRels.Rubrics.criterion);
+					var feedback = null;
```
djwester: I think we need to make the actual assessmentEntity default to null.

I don't think this fixes the issue

walinga: Yeah, you're probably right

walinga:
I think the build failure was unrelated to our changes. I think we can probably merge this

PR #75, opened 2018-05-11
Double border fix

walinga:


assessment-result-behavior.html
```diff
@@ -53,6 +58,35 @@
...
+				if (selectedCriterionCellEntity) {
+					var thisLevel = selectedCriterionCellEntity.getLinkByRel(this.HypermediaRels.Rubrics.level).href;
+					cellRelationsMap[prevLink] = (prevLevel === thisLevel);
+
+					prevLink = selectedCriterionCellEntity.getLinkByRel(this.HypermediaRels.Rubrics.criterionCell).href;
+					prevLevel = thisLevel;
+
+					var feedback = selectedCriterionCellEntity.getSubEntityByClass(this.HypermediaClasses.rubrics.feedback);
+					if (feedback && feedback.properties && feedback.properties.text) {
+						prevLevel = null;
```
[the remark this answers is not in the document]
walinga: You're right, I should actually be using properties.html instead since feedback does not get displayed otherwise

d2l-rubric-criteria-group.html
```diff
@@ -69,6 +69,12 @@
 				border-left: 1px solid var(--d2l-color-celestine);
 				border-bottom: 1px solid var(--d2l-color-celestine);
 			}
+			.bottom {
```
jstefaniuk-d2l: rather than using important here, try making it more specific (like `.criterion-cell.bottom {`) so that css will prioritize it 

I like to use `!important` for debugging, but it tends to cause annoying errors in prod code so only use it when absolutely necessary

walinga: Cool, thanks for the tip!

PR #88, opened 2018-05-18
Vwong/us96646 - Edit Level Name - Call Level Action with name changes

wongvincent:


editor/d2l-rubric-level-editor.html
```diff
@@ -15,7 +16,11 @@
 
 		</style>
 
-		<d2l-text-input value="[[entity.properties.name]]" aria-label="todo"></d2l-text-input>
+		<d2l-text-input
+			id="level-name"
+			value="[[entity.properties.name]]"
+			on-change="_saveName"
+			aria-label="level name"></d2l-text-input>
```
mdulat: assuming this will be localized?

wongvincent: Yep. Would be nice to change it at the same time the "criterion name" aria-label is changed

PR #105, opened 2018-05-30
fix various ie11 errors

mdulat:


store/siren-action-behavior.html
```diff
@@ -51,11 +51,9 @@
 		_appendHiddenFields: function(action, fields) {
 			if (action.fields && action.fields.forEach) {
 				action.fields.forEach(function(field) {
-					if (field.type !== 'hidden' || field.value === undefined) {
-						return;
+					if (field.type === 'hidden' && field.value !== undefined) {
+						fields.append(field.name, field.value);
 					}
-					// if the field is specified multiple times, assume it is intentional
-					fields.append(field.name, field.value);
```
mdulat: Note this particular change wasn't an ie11 issue, I just thought it made more sense logically in this case.

PR #110, opened 2018-06-04
Mbayly/add experimental criterion animation

mdgbayly:
This is just putting in place some building blocks to enable animation when we implement the add criterion story.

The difficulty for us for in implementing animation is that we are relying on the hypermedia responses and the automatic re-rendering to generate the UI changes in response to actions, so it would be circumventing that if we were to manage the animations manually. 

So this change puts in place some controls/logic for determining when a new entity (in this case a criterion) is rendered and hence is a candidate for a CSS transition.

I've also added the foundation for an opacity/height based animation for the criterion.

Animating this stuff is tricky as described in this blog post, because we want the height of a criterion to be dynamic and you cannot animate to an auto value.  I've opted for a variation of two of the techiques described in the blog post, setting a default max height and then animating to that height, then removing the max-height after the animation completes.  This might work reasonably well for an empty criterion, although it might become more problematic when we make the feedback display optional.

https://css-tricks.com/using-css-transitions-auto-dimensions/

I've also added support for a parent element to indicate to child components that the parent is animating so the children should not.

This is the not so clearly named `animating` property. If this is passed to a child component with the value true, it means the parent is animating so even if the child thinks it is a new entity, it should not animate.  This is partly to prevent the criterion from animating when they are first rendered, but could also be useful in other scenarios.

I think this stuff is going to get even more tricky/problematic when we try and animate levels as that affects multiple components, and the simple approach I'm using here of animating when the observed entity changes from undefined to defined, won't work so well because a polymer component might be used to render a different but newly added entity e..g if inserting into a list.

### DIFF
--- a/editor/d2l-rubric-criteria-editor.html
+++ b/editor/d2l-rubric-criteria-editor.html
@@ -71,19 +71,27 @@
 				D2L.PolymerBehaviors.Rubric.LocalizeBehavior
 			],
 
-			observers: [
-				'_onEntityChanged(entity)',
-			],
-
-			_onEntityChanged: function(entity) {
+			_onEntityChanged: function(entity, oldEntity) {
 				if (!entity) {
 					return;
 				}
 				this._criteriaEntities = entity.getSubEntitiesByClass(this.HypermediaClasses.rubrics.criterion);
-				this.animating = false;
-				Polymer.RenderStatus.afterNextRender(this, function() {
-					this.animating = false;
-				}.bind(this));
+				// EXPERIMENTAL animation/transition handling. If oldEntity is undefined, then
+				// this represents the initial render so indicate to child components that we
+				// are animating so that child components don't animate on initial render.
+				// After render, reset the animation flag so that subsequently added criterion will animate.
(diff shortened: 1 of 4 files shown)


editor/d2l-rubric-criteria-editor.html
```diff
@@ -56,6 +58,10 @@
 				hasOutOf: {
 					type: Boolean,
 					value: false
+				},
+				animating: {
```
dbatiste: For this sort of thing, if we name with leading underscore, docs generator will treat as private and omit from document pages.

store/entity-editor-behavior.html
```diff
@@ -119,6 +124,21 @@
 			this.splice('_alerts', 0, this._alerts.length);
 			this._hasAlerts = false;
 		},
+
+		_transitionElement: function(element, maxHeight) {
+
+			element.style.maxHeight = maxHeight + 'px';
```
dbatiste: This looks like a good candidate for fastdom (we have an import that can be used).

PR #117, opened 2018-06-11
update script and use the latest data

djwester:


demo/data/rubrics/organizations/custom-points/198.json
```diff
@@ -47,7 +47,7 @@
             "rel": [
                 "https://api.brightspace.com/rels/organization"
             ],
-            "href": "data/rubrics/organizations/custom-points.json"
+            "href": "data/rubrics/custom-points.json"
```
djwester: this href isn't used, so this change doesn't really matter - not sure why the script it making this be different now, but shouldn't affect anything

PR #119, opened 2018-06-11
Use the onEntityChange override instead, this was causing the behavio…

alogan519:
…r entityChanged method to not be called


ChrisKraljevicD2L:
Tested in BfP, this gets point values for criteria levels and the "out of" column working :+1:

Selected levels for criteria and feedback still aren't working though

PR #123, opened 2018-06-13
explicitly call FetchSirenEntityBehavior.registered to populate white…

mdgbayly:
…lists - needs revisiting


mdgbayly:
@alogan519 @ChrisKraljevicD2L Just a heads up. We ran into issues with d2l-rubrics/d2l-rubrics-editor with the upgrade to d2l-fetch-siren-entity-behaviour 5.x.  We had been using it from our `entity-store` implementation as a library rather than a true Polymer behavior, which was probably a bit hacky...

With the upgrade to the version that uses whitelists, this broke things because the `registered` method, where the whitelist monostate is setup, was never being called. To get things working and to fix the changes that deployed to BSI (but never deployed to LMS), I've added a call to `registered` whenever we call `fetchEntityWithToken`, in order to get things working, but will probably want to revisit that.

PR #127, opened 2018-06-14
Add trash icon

wongvincent:
Add trash icon.
Icon is visible only when criterion has delete action and there is >1 criterion in the rubric.
Hover on trash can reveals a tooltip


editor/d2l-rubric-criterion-editor.html
```diff
@@ -17,10 +18,11 @@
 	<template strip-whitespace>
 		<style include="d2l-rubric-editor-cell-styles">
 			:host {
+				position: relative;
 				display: flex;
 				flex-direction: row;
 				justify-content: space-between;
-				overflow: hidden;
+				background-color: white;
```
mdulat: Is background-color in here twice?

wongvincent: indeed. good catch

PR #133, opened 2018-06-29
Remove empty criteria columns for holistic rubrics

walinga:


alogan519:
Looks like there is a holistic class on the top level rubric siren entit

PR #140, opened 2018-07-04
Us97438 add level buttons

mdulat:
Added level buttons on either end. 
Added disabled trashcan buttons for each level so that we can see what it looks like with tall level headings.
Removed large tooltips for d2l-button-icon types since they already have browser tooltips built-in (as requested by design team).

### DIFF
--- a/editor/d2l-rubric-criterion-editor.html
+++ b/editor/d2l-rubric-criterion-editor.html
@@ -103,6 +103,11 @@
 				width: 1.5rem;
 			}
 
+			.col-last-empty {
+				min-width: calc(1rem + 44px); /* button width = 44px including border */
+				flex: 0 0 0;
+			}
+
 			.criterion-remove {
 				position: absolute;
 				margin: 1rem 0;
@@ -166,6 +171,7 @@
 			</div>
 		</div>
 		<div hidden="[[!hasOutOf]]" class="cell col-last points">/ <d2l-text-input value="" aria-label="todo" prevent-submit></d2l-text-input></div>
+		<div hidden="[[hasOutOf]]" class="cell col-last-empty"></div>
 		<div class="criterion-remove">
 			<d2l-button-icon
 				id="remove"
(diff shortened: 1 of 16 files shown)


editor/d2l-rubric-criterion-editor.html
```diff
@@ -166,16 +171,14 @@
...
 		<div hidden="[[!hasOutOf]]" class="cell col-last points">/ <d2l-text-input value="" aria-label="todo" prevent-submit></d2l-text-input></div>
+		<div hidden="[[hasOutOf]]" class="cell col-last-empty"></div>
 		<div class="criterion-remove">
 			<d2l-button-icon
 				id="remove"
 				hidden$="[[!_showRemove(entity, moreThanOneCriterion)]]"
 				icon="d2l-tier1:delete"
 				text="[[localize('removeCriterion')]]"
 			></d2l-button-icon>
-			<template is="dom-if" if="[[_showRemove(entity, moreThanOneCriterion)]]">
```
wongvincent: They requested no tooltip for delete criterion?

mdulat: Correct, because it already has a built-in tooltip that comes with the title attribute (which is assigned the text value). We were getting double tool-tips.

editor/d2l-rubric-levels-editor.html
```diff
@@ -19,13 +21,44 @@
...
+				align-items: center;
+				justify-content: flex-end;
+			}
+			.col-last {
+				display: flex;
+				flex-direction: row;
+				align-items: center;
+				justify-content: flex-start;
+			}
+			.empty {
```
wongvincent: I find calling the class `empty` a little confusing.
I wonder if it makes sense to add an attribute to the element if hasOutOf is true. Then we can use a cssSelector and apply different styles if it has that attribute instead of adding and removing classes programatically.

mdulat: yes, attribute is simpler / cleaner.

PR #149, opened 2018-07-10
Pass in parameters to call functions on entity change

walinga:


d2l-rubric-criteria-group.html
```diff
@@ -247,8 +247,8 @@
 				return entity && entity.properties && entity.properties.outOf;
 			},
 
-			_showRowHeaders: function() {
-				return this.rubricType !== 'holistic';
+			_showRowHeaders: function(rubricType) {
+				return rubricType !== 'holistic';
 			},
```
djwester: Do you also need to update the call to _showRowHeaders in the _getColumnCount?

walinga: Good point

d2l-rubric-criterion-cell.html
```diff
@@ -105,7 +105,7 @@
 				}
 			},
 
-			_showSelectedIcon: function(assessmentResult) {
+			_showSelectedIcon: function(entity, assessmentResult) {
```
alogan519: is this in case entity changes we rerender?

walinga: Yeah, I'm hoping this will do the trick

PR #152, opened 2018-07-11
Fix issue with extra borders in Polymer 1

walinga:


d2l-rubric-criteria-group.html
```diff
@@ -123,8 +123,10 @@
 						</template>
 						<template is="dom-repeat" items="[[_getCriterionCells(criterion)]]" as="criterionCell">
 							<d2l-td class$="[[_getCriteriaClassName(criterionCell, assessmentResult)]]">
-								<div class="cell-left-border" hidden="[[!_isSelected(criterionCell, assessmentResult)]]"></div>
-								<div class="cell-bottom-border" hidden="[[!_hasBottom(criterionCell, assessmentResult, noBottomCells, criterionNum, _criteriaEntities)]]"></div>
+								<template is="dom-if" if="[[_isSelected(criterionCell, assessmentResult)]]">
```
walinga: Looks like just adding this `dom-if` does the trick. Removing the null default value for `assessmentResult` just causes more problems

PR #154, opened 2018-07-11
Increase height of empty criterion description cell editor

emil-furniss:
Tweak styling of the criterion description cell editor to have a min height twice that of the predefined feedback cell.
https://trello.com/c/yoAtPcBX/18-rubric-criterion-description-cell-height-should-should-be-double-compared-to-predefined-feedback

### DIFF
--- a/editor/d2l-rubric-criterion-description-editor.html
+++ b/editor/d2l-rubric-criterion-description-editor.html
@@ -13,6 +13,7 @@
 			:host {
 				display: flex;
 				flex-grow: 1;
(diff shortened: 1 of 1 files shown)


mdulat:
Just want to check - this was tested with multiple criterion, right? And doesn't mess up the rest of the "grid"

emil-furniss:
Correct. It doesn't mess up the grid, applies evenly to all description cells, and doesn't interfere with adding or removing criteria or it's animation.

PR #155, opened 2018-07-11
Update to dependency versions using the newest icons

svanherk:
No uses of `<d2l-icon-button>`

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -25,16 +25,16 @@
     "demo"
   ],
   "dependencies": {
-    "d2l-alert": "^3.0.1",
+    "d2l-alert": "^3.1.0",
     "d2l-colors": "^3.1.2",
     "d2l-fetch": "Brightspace/d2l-fetch#^1.8.0",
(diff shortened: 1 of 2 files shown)


bower.json
```diff
@@ -25,16 +25,16 @@
...
   ],
   "dependencies": {
-    "d2l-alert": "^3.0.1",
+    "d2l-alert": "^3.1.0",
     "d2l-colors": "^3.1.2",
     "d2l-fetch": "Brightspace/d2l-fetch#^1.8.0",
     "d2l-fetch-siren-entity-behavior": "Brightspace/d2l-fetch-siren-entity-behavior#^5.0.2",
     "d2l-hm-constants-behavior": "git://github.com/Brightspace/d2l-hm-constants-behavior.git#^5.8.0",
-    "d2l-icons": "^4.10.0",
+    "d2l-icons": "^5.0.0",
```
mdulat: If we need this version it should probably be updated in the resolutions too?

dlockhart: Yes, you're correct @mdulat. Ideally you can just remove it from the resolutions.

PR #156, opened 2018-07-11
update resolution for icons

mdulat:


svanherk:
Nice catch, totally missed the resolution, thanks Margaret! Is it even needed anymore?  I tried just removing it and everything seems to work.

emil-furniss:
I don't think we need a resolution at the moment, there doesn't appear to be any conflict.

PR #157, opened 2018-07-12
Some build cleanup - Also trying to reduce sauce usage by only running two browsers per build

djwester:
…to sauce


djwester:
What I've done here is a bit unconventional - I'm randomly picking two os/browser combinations from a list of supported platforms.  Let me know what you think.  The purpose of it is to reduce the number of minutes we spend on Sauce labs while still giving us good platform coverage in the long run.

I do have 1 concern around this and that is around re-runs.  If a build fails and then we just re-start it on Travis it might run with a different platform next time and potentially pass if the problem was indeed platform specific. 

I think this is not going to happen much and I will be monitoring summary statistics in Sauce labs itself so I should be able to see if a particular platform fails more than others so hopefully this won't cause us too many issues.  I would also hope that developers won't just mindlessly restart builds without first taking a look at if a failure is legitimate.

PR #159, opened 2018-07-12
Update version to 0.70.0

walinga:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "d2l-rubric",
-  "version": "0.69.0",
+  "version": "0.70.0",
```
djwester: I drafted a new release, so we should have this version now.

PR #160, opened 2018-07-13
Fix IE11 spacing

emil-furniss:
Fixes https://trello.com/c/Yy5aMio0/5-rubric-ie11-funky-spacing
IE doesn't accept `flex` to be set to `0 0 0` without any units ("invalid rule"), so changing it to `0 0 0%` fixes the width of the empty last column. The extra padding fixes the width of the first column and is in line with how the header row is styled.
This has been tested on Chrome, Chrome mobile view, Firefox, Edge, and IE.

### DIFF
--- a/editor/d2l-rubric-criterion-editor.html
+++ b/editor/d2l-rubric-criterion-editor.html
@@ -65,6 +65,10 @@
 				flex-direction: column;
 			}
 
+			.cell {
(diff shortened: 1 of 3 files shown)


emil-furniss:
Turns out there is a documented bug and workaround for this case
https://github.com/philipwalton/flexbugs#flexbug-7

PR #170, opened 2018-07-23
Add Reverse Levels Order Button, hook up button and error handling

stevenao:


editor/d2l-rubric-editor-menu.html
```diff
@@ -0,0 +1,54 @@
...
+				justify-content: flex-end;
+				overflow: hidden;
+				padding-top: 0.5rem;
+				padding-bottom: 1.1rem;
+				padding-right: 46px;
+			}
+			[hidden] {
+				display: none;
+			}
+			:host-context([dir='rtl']) {
```
mdgbayly: I'm not convinced both these rtl declarations are required? Did you try it in a demo page?  I was testing 'rtl' with some of the drag-and-drop stuff i was working on, and it seemed like to me that in all browsers apart from Chrome, 'RTL' is broken in the Polymer 2 mainline variant, regardless of which of these declarations you use.  But the newer `:dir(rtl)` format seems to work in the 1.x variant anyway. As :host-context is deprecated, I'm not sure we need to keep adding both and can probably just use the newer `:dir()` format?

editor/d2l-rubric-editor-menu.html
```diff
@@ -0,0 +1,54 @@
...
+				reverseLevelsEnabled: {
+					type: Boolean,
+					value: false
+				},
+			},
+			behaviors: [
+				D2L.PolymerBehaviors.Rubric.LocalizeBehavior
+			],
+			_handleReverseTap: function() {
+				this.fire('reverse-levels');
```
mdgbayly: Should we use a more expressive name here rather than just `reverse-levels`? Most of our other events include `d2l-rubric-...`.

editor/d2l-rubric-editor-menu.html
stevenao: yup, I copied from the other component. I just keep the newer one now

PR #172, opened 2018-07-23
TA126868 Show criteria group name if action exists

emil-furniss:
Shows group name text input are if action exists and there is more than one group.


editor/d2l-rubric-criteria-group-editor.html
```diff
@@ -37,10 +37,22 @@
 
 			.header {
 				border-bottom: var(--d2l-table-border);
```
mdgbayly: @emil-furniss 

I had to refactor some of this layout to support drag and drop and I removed some of the intermediate `<div>` elements like the header as they were no longer adding any value.

https://github.com/Brightspace/d2l-rubric/pull/174

So I think we are going to have some conflicts when these both get merged.

But regardless, I'm not keen on the approach of using `position: absolute` here in order to position the group name within the first cell of the levels editor.  It's going to be difficult to keep things consistent as the design evolves with approaches like that.

I think a better approach would be to utilize a `<slot>` in the d2l-rubric-levels-editor, which the  `d2l-rubric-criteria-group-editor` can use to distribute the name field that it controls.

PR #179, opened 2018-07-26
US98251 -- Add Overall Score display to rubric preview

mpharoah-d2l:
Does not show the selected level in graded view yet.


d2l-rubric.html
```diff
@@ -236,6 +240,18 @@
 				return !!outOf || outOf === 0;
 			},
 
+			_hasOverallScore: function(entity) {
+				return (
+					entity &&
+					entity.hasLink('https://rubrics.api.brightspace.com/rels/overall-levels') &&
```
alogan519: this should be added to the HMConstants behavior, can you also add in something for me? (would save a version bump)

d2l-rubric.html
```diff
@@ -143,6 +144,9 @@
 				</div>
 			</div>
 		</div>
+		<template is="dom-if" if="[[_hasOverallScore(entity)]]">
```
alogan519: you could probably do a <div hidden="[[!_hasOverallScore(entity)]]"> here, I have been told it is faster to do it that way.

mpharoah-d2l: Wouldn't that still try to build the children though?

d2l-rubric.html
mpharoah-d2l: Where is the repo that has that?

mpharoah-d2l: Nevermind, found it

d2l-rubric.html
djwester: There are sometimes issues in 1.x with hidden as well - I think it will be ok here since the function has an explicit parameter, but we if switch to using hidden we need to make sure it works in 1.x

mpharoah-d2l: I think using `hidden` would make it "crash" since it would try to get the href of a null link when there's no overall score rel.

djwester:
Should we add any tests for the new page?

d2l-rubric.html
alogan519: Ya to use hidden you would need to have it handle null cases.

Ya I think the general rule we have been following is to put simple stuff in hidden and complex things in dom-if. Not sure of the tradeoffs of building the children vs the dom-if stamping. Guess just leav dom-if here?

test/components/d2l-rubric.html
```diff
@@ -25,6 +25,13 @@
 			</template>
 		</test-fixture>
 		<script src="d2l-rubric-text-only.js"></script>
+		
+		<test-fixture id="rub-overall-score">
+			<template>
+				<d2l-rubric href='static-data/rubrics/organizations/overall-score/200.json' token='foozleberries'></d2l-rubric>
```
mdgbayly: @mpharoah-d2l @emil-furniss These tests seem to be failing for me locally. Confused why they didn't fail in CI?
@emil-furniss Not entirely sure why these tests were switched to use `static-data`? We've been using `static-data` as a supposedly temporary (but not done a good job at going back to these) workaround where we need siren actions in the test data and the APIs don't have them yet. But that shouldn't generally be needed for the viewer components?

mpharoah-d2l: Passes locally for me. You might have broken dependencies or something. Are you on master?

emil-furniss: For me the tests pass locally when run in Chrome via `polymer serve`, but fail on `chrome 67 [1.x]` when run from the command line. And yes, `static-data` is generally only used for the editor demo and tests where API actions are needed.

emil-furniss: @mdgbayly Actually, deleting and reinstalling `bower-components` and `bower-components-1.x` and rerunning `npm install` made the cli tests pass. Though that doesn't change the fact that `static-data/rubrics/organizations/overall-score/200.json` doesn't appear to exist...

PR #182, opened 2018-07-27
fix error handling

stevenao:


editor/d2l-rubric-editor.html
```diff
@@ -119,7 +123,7 @@
 		<d2l-rubric-loading hidden$="[[_hideLoading(_showContent,_hasAlerts)]]"></d2l-rubric-loading>
 		<div hidden$="[[_hideLoading(_showContent,_hasAlerts)]]" class="out-of-loader">
 		</div>
-		<div hidden$="[[_hideOutOf(_showContent,_hasAlerts)]]">
+		<div id="rubric-editor-container" hidden >
```
wongvincent: Could we set the hidden attribute using a property instead?  `hidden$=[[!_showEditor]]`
`default: false`
Then we don't need the `_displayEditor` function and we can just do `this._showEditor = true` or `false`

stevenao: i usually try to leave data-binding for displaying value only thinking of the overhead it adds. Also, I want to keep the template look more static rather than having all sorts of conditions embedded in it.

PR #185, opened 2018-07-31
Polymer dom-repeat not updating when entity order changes

wongvincent:


mdulat:
Does the reorder API call not return link headers to cause re-fetching of the criteria in the right order? Or does the dom-repeat just think that the changed entity isn't actually changed?

wongvincent:
The API returns the criteria in the correct order. The `onEntityChanged` function is even called, but the dom-repeat ends up updating the DOM back to what it was before! :(

mdulat:
I would at least add a comment there to indicate why it's necessary to update the property there, but it does feel icky. I think @mdgbayly should have a look as well.

PR #190, opened 2018-08-01
Fixed rub-overall-score test fixture

mpharoah-d2l:


emil-furniss:
I'm getting 404s from the overall-levels link href `static-data/rubrics/organizations/text-only/199/overallLevels.json` in `static-data/rubrics/organizations/text-only/199.json`. Is that behavior expected? 
overall-levels-404

mpharoah-d2l:
The rubric shouldn't include an overall-levels link if it doesn't have overall levels enabled. I just forgot to remove it from that one

emil-furniss:
Last little thing: Could you change the href for the `rub-text-only` fixture to `data/rubrics/organizations/text-only-no-edit/199.json`?
That way we minimize the usage of `static-data` which we eventually want to get rid of.

PR #191, opened 2018-08-01
Telemetry skeleton implementation

walinga:


d2l-rubric-criteria-group.html
```diff
@@ -196,6 +202,11 @@
 				if (!entity) {
 					return;
 				}
+				// The first time that the entity loads, we send out an Open event
```
alogan519: looks like it's just a "TelemetryEvent"

telemetry-behavior.html
```diff
@@ -0,0 +1,84 @@
...
+	window.D2L.PolymerBehaviors = window.D2L.PolymerBehaviors || {};
+	window.D2L.PolymerBehaviors.Rubric = window.D2L.PolymerBehaviors.Rubric || {};
+
+	/**
+	 * Behavior for sending telemetry messages to the telemetry service
+	 * @polymerBehavior
+	 */
+	D2L.PolymerBehaviors.Rubric.TelemetryBehaviorImpl = {
+
+		_logEvent: function(entity, isMobile, eventType) {
```
djwester: Do we have any specs on what/when we want to log things?

d2l-rubric-criteria-group.html
djwester: Should we have a way to turn this on/off? - Maybe LD flag?

telemetry-behavior.html
walinga: The [feature](https://rally1.rallydev.com/#/57732444928d/detail/portfolioitem/feature/210511399256) asks for a number of things that we should log. My plan is to have `isMobile`, `rubricMode`, and `originTool` on the default telemetry event, along with standard event data (tenant ID, browser, etc.). Then on the performance event, we can add in metrics for rubric rendering.

Not sure if this is the right spot that we want to be sending the event, but that's flexible so we can easily change it if we need to.

d2l-rubric-criteria-group.html
walinga: The event `Action` is "Open". We might add different event types in the future (eg. a "Close" event), so I wanted to specify what kind of event we are sending

walinga: LD flag coming soon in a separate PR

PR #193, opened 2018-08-01
Add ability to grade rubrics, can select cells, add feedback and edit feedback

alogan519:


assessment-result-behavior.html
```diff
@@ -52,9 +67,9 @@
 
 					var feedback = selectedCriterionCellEntity.getSubEntityByClass(this.HypermediaClasses.rubrics.feedback);
 					var feedbackHTML = feedback && feedback.properties && feedback.properties.html;
-
+					var feedbackText = feedback && feedback.properties && feedback.properties.text;
```
walinga: Might be a bit cleaner to pass in `feedback.properties` to `_updateAssessmentMap` and then select the HTML and text within that function

assessment-result-behavior.html
```diff
@@ -63,12 +78,24 @@
 					// If there is no selected cell (the rubric is partially assessed) then skip this row
 					prevLink = null;
 				}
+				if (!this.readOnly) {
+					// build a map of criterion-cell hrefs to assessment-criterion-cell hrefs
+					var criterionCells = criterionEntity.getSubEntitiesByClass('assessment-criterion-cell');
```
walinga: Can we put this `if` statement in a helper function? This for-loop is pretty big already, so it's good to break it up when we can

alogan519: sounds good

store/siren-action-behavior.html
```diff
@@ -138,8 +138,39 @@
...
+
+		performSirenActionJSON: function(action, fields) {
+			var headers = new Headers();
+			this.token && headers.append('Authorization', 'Bearer ' + this.token);
+			var url = new URL(action.href, window.location.origin);
+			headers.set('Content-Type', action.type);
+			var json = {};
+			fields.forEach(function (field) {
+				json[field.name] = field.value;
+			});
```
awikkerink: The type of body will be incorrect if the action type is not json. Maybe integrate this logic into `performSirenAction` when the type is json, or throw an error.

awikkerink: I could see the use of `performSirenActionJSON`, but I would expect to pass in the body as JSON.

Maybe call it `performSirenActionRaw`? Since the hidden fields are not grabbed?

Maybe also create a `getSirenFieldsJSON` method which returns a JSON object instead of `FormData` or `URLSearchParams`.

assessment-result-behavior.html
```diff
@@ -43,57 +56,75 @@
...
 			return assessmentMap;
 		},
 
-		_updateAssessmentMap: function(assessmentMap, selectedCriterionCellEntity, link, level, feedbackHTML) {
+		_updateAssessmentMap: function(assessmentMap, selectedCriterionCellEntity, link, level) {
 			var rowLink = selectedCriterionCellEntity.getLinkByRel(this.HypermediaRels.Rubrics.criterion);
 
-			if (feedbackHTML) this.anyFeedback = true;
-
+			if (feedbackHTML) {
```
walinga: This file's looking pretty good now! I've opened a PR onto your branch that fixes an issue with `anyFeedback` and cleans up a few more things

d2l-rubric-feedback.html
```diff
@@ -32,33 +38,127 @@
...
-			[[localize('feedback')]]
-		</div>
-		<div class="feedback-text">
-			<slot></slot>
+			.feedback-wrapper {
+				padding: 0.5rem;
+				border: 1px solid transparent;
+			}
+
+			.feedback-wrapper-editable:hover {
```
walinga: Does this class ever get used?

alogan519: No, I think eventual plan for is to change background, could remove.

walinga: I guess it doesn't hurt to keep it if you plan to use it later

assessment-result-behavior.html
```diff
@@ -110,19 +142,96 @@
...
+				this._selectCriterionCell(this._assessmentCriterionCellMap[criterionCellHref]);
+			}
+		},
+
+		_selectCriterionCell: function(criterionCell) {
+			var action = criterionCell.getActionByName('select-criterion-cell');
+
+			if (action) {
+				var fields = action.fields;
+				var setAsSelectedField = fields.find(function(field) { return field.name === 'setAsSelected'; });
```
alogan519: This field is no longer there, we should remove it here and from saveAssessmentFeedback

PR #194, opened 2018-08-01
Switch to using d2l-fetch-auth for rubric editing token management

mdgbayly:
We need to support a user's auth token expiring and automatically renewing during long editing sessions. 

Initially we were setting the token as an attribute of the editor by grabbing a token from the LMS when the editor is initially rendered. However, this does not provide any mechanism to auto-renew the token if it expires.

So as a work around we are switching the d2l-rubric-editor to rely on d2l-fetch-auth to manage the auth token, because this supports automatic token refresh in the context of a fetch request if the token has expired.

In order to retain the current functionality of the entity store and to enable the rubric view/assessment components to continue to utilize token attributes, we will pass an empty token in the d2l-rubric-editor. This ensure the entity-behavior and entity-store continue to function as before but also allows us to defer to d2l-fetch-auth to add a token header if the token in the component is empty.

At the same time, I have modified the entity store to use d2l-fetch directly rather than going through d2l-fetch-siren-entity-behavior. This was not strictly required for this change (we could have conditionally called the d2l-fetch-siren-entity-behavior API that does not take a token), but as previously discussed, we had already had to modify the fetch implementation used for refetching after siren actions to enable us to have more control over caching, and in general I did not feel d2l-fetch-siren-entity-behavior was adding much value for us (and the whitelisting in that component just made development more complicated!). So with this change the fetch of entities is now consistent.


store/entity-behavior.html
```diff
@@ -78,11 +78,7 @@
 				this.removeListener();
 			}
 			this.removeListener = window.D2L.Rubric.EntityStore.addListener(href, token, this._entityChangedHandler);
-			var entity = window.D2L.Rubric.EntityStore.fetch(href, token);
-			if (entity.status !== 'fetching') {
-				// Allows class/mixin to override _entityChanged
-				this._entityChanged(entity.entity);
```
awikkerink: This is here because if the entity is already fetched the `_entityChangedHandler` isn't called.

mdgbayly: Yeh - in order to support returning a promise from EntityStore.fetch, i modified it to notify when the entity already exists:
https://github.com/Brightspace/d2l-rubric/blob/master/store/entity-store.html#L171

Not entirely sure I like that approach and it can mean that if you have multiple components listening on the same entity, they would all be notified again.

The alternative was to then the returned promise right in this method rather than checking the entity.status, but that would mean you would be notified twice for the more usual case that the entity does not exist yet.

If you have other ideas of how to support returning promises from fetch I'd be interested to hear them.

The main reason for adding promises was so that we could call fetch in response to a siren action particularly when we are processing link headers, so we can wait on the linked resources until we re-render the primary entity being operated on.

https://github.com/Brightspace/d2l-rubric/blob/master/store/siren-action-behavior.html#L149

And this was done with the hope being that it would lead to a smoother re-rendering after an operation.

awikkerink: I think calling notify twice should be okay since Polymer won't detect the change if `entity` doesn't change

PR #195, opened 2018-08-01
Use OSX 10.13

wongvincent:


wongvincent:
Took 3 tries.. but it finally passed?!

PR #197, opened 2018-08-02
Mpharoah/misc overall score fixes

mpharoah-d2l:
* Fixed overall level score description potentially overflowing its container
* Fixed overall level points using the wrong lang term
* Fixed bug where scroll buttons would sometimes not appear until the window is resized


d2l-rubric-overall-score.html
```diff
@@ -119,14 +125,17 @@ <h4>[[level.properties.name]]</h4>
 					return;
 				}
 				this._levels = entity.getSubEntitiesByClass(this.HypermediaClasses.rubrics.overallLevel) || [];
+				Polymer.RenderStatus.afterNextRender(this, function() {
```
djwester: Would the fastdom.measure() method be appropriate here?

mpharoah-d2l: It does both measurement and mutation, so I don't know if fastdom is usable here. There shouldn't be too much cost to using `afterNextRender` since this only gets called one time when the siren object is fetched

PR #202, opened 2018-08-03
remove unnecessary resolutions

mdulat:


wongvincent:
probably need to enter `winpty bash`, then do bower install and see where the dependency conflict is

mdulat:
Looks like Dave fixed the ref in textarea, so I re-ran the build, should pass now (fingers crossed)

PR #212, opened 2018-08-14
Add overall level description and feedback UI

wongvincent:
- Reuse existing criterion-description and criterion-feedback for overall level description and feedback.
- Rename criterion-description and criterion-feedback to be more generic. Update tests as well.
- Add static-data for overall-levels (temporary until regen data is merged) and add tests to test overall level description and feedback

Because we are using the existing component, this PR does the following on both overall description and overall feedback...
- checks if PUT action exists to allow editing
- calls the PUT API on change.
- aria announces which x of y overall description/feedback we are on


mdgbayly:
Looks like this is superseded by other PRs so closing.

PR #213, opened 2018-08-14
us98812 name editor support overall feedback (ui)

stevenao:


editor/d2l-rubric-overall-levels-editor.html
```diff
@@ -6,57 +6,69 @@
...
-				margin-top: 5px; /* temp - remove later */
 				display: block;
 			}
 
 			* {
 				box-sizing: border-box;
 			}
 
+			.gutter-left {
+				margin-left: calc(var(--d2l-rubric-editor-gutter-width) - 0.3rem);
```
mdgbayly: Not entirely clear to me why we need to set left/right margins on the gutters rather than just giving them widths and letting them flow within the flexbox? But OK if it works.

stevenao: For the overall score section, gutter left starts at 40px and gutter right starts at width - 40px - basically to left and right align the two level add buttons to the Criteria Group editor body. The designers want it the way they have in the mock-up.

mdgbayly: OK. Just seems a bit complex having the margins on those two internal divs. Wouldn't it be simpler to just add those margins to the styling of the overall levels component where it is added in d2l-rubric-editor?

editor/d2l-rubric-overall-levels-editor.html
```diff
@@ -103,10 +106,6 @@
 			}
 
 			#level-header .col-center div:last-of-type {
-				/* display: flex;
-				flex-direction: row;
-				align-items: center;
-				justify-content: flex-start; */
 				border-top-right-radius: var(--d2l-table-border-radius);
```
mdgbayly: Does this work with RTL?
Also wasn't clear to me where the internal level left/right borders are coming from, but assuming it all works.

stevenao: yeah... works on the demo page. I will do a more global testing soon.

PR #215, opened 2018-08-15
Regen api data with overall levels

wongvincent:


mpharoah-d2l:
So is there just a QA site somewhere with rubrics that the test data is generated from? Is that how it works for regenerating data?

djwester:
@mpharoah-d2l essentially yes - You give the script the URL to the qa site you want to use. The rubrics in the script are defined on all QA sites.

mpharoah-d2l:
How are the rubrics added to the QA sites? Does it create them through API calls, or is it just static database data? It would be good to add a rubric with UsesOverallScore = false so we can test both cases.

PR #219, opened 2018-08-16
Fix IE11 group name focus issue

emil-furniss:
Addresses https://trello.com/c/NcXxPkiH/29-ie11-when-adding-a-new-criterion-group-focus-goes-to-first-criterion-group-name-not-the-newly-added-one-as-intended

In cases where the DOM is slow to update, the new group is not yet present when trying to select it for setting the focus to its name. ~So in those cases, an explicit delay is added to give the DOM time to update so that the correct group will be selected.
(I'm not huge on the setTimeout approach, even if the delay is only 50 ms, so any suggestions for alternative approaches welcome!)~

### DIFF
--- a/editor/d2l-rubric-criteria-groups-editor.html
+++ b/editor/d2l-rubric-criteria-groups-editor.html
@@ -112,7 +112,7 @@
 						setTimeout(function() {
 							this.fire('iron-announce', { text: this.localize('groupAdded') }, { bubbles: true });
 						}.bind(this), 2000);
-					}.bind(this), function() {
+					}.bind(this)).catch(function() {
 						this._waitingForGroups = false;
 					}.bind(this));
 				}
--- a/editor/d2l-rubric-criteria-groups-editor.html
+++ b/editor/d2l-rubric-criteria-groups-editor.html
@@ -107,11 +107,13 @@
 				var action = this.entity.getActionByName('create');
 				if (action) {
 					this._waitingForGroups = true;
-					this.performSirenAction(action).then(function() {
+					this.performSirenAction(action).then( function() {
 						this.fire('d2l-rubric-criteria-group-added');
 						setTimeout(function() {
 							this.fire('iron-announce', { text: this.localize('groupAdded') }, { bubbles: true });
 						}.bind(this), 2000);
+					}.bind(this), function() {
+						this._waitingForGroups = false;
 					}.bind(this));
 				}
 			},
--- a/editor/d2l-rubric-criteria-groups-editor.html
+++ b/editor/d2l-rubric-criteria-groups-editor.html
@@ -77,6 +77,11 @@
 				_canCreate: {
 					type: Boolean,
 					computed: '_canCreateCriteriaGroup(entity)'
+				},
+
+				_waitingForGroups: {
+					type: Boolean,
+					value: false
 				}
 			},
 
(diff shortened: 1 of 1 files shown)


editor/d2l-rubric-criteria-groups-editor.html
```diff
@@ -106,10 +106,18 @@
 						setTimeout(function() {
 							this.fire('iron-announce', { text: this.localize('groupAdded') }, { bubbles: true });
 						}.bind(this), 2000);
-					}.bind(this)).then(function() {
 						var allGroups = Polymer.dom(this.root).querySelectorAll('d2l-rubric-criteria-group-editor');
-						var lastGroup = allGroups[allGroups.length - 1];
-						lastGroup.$$('d2l-text-input').$$('input').select();
+						// If the DOM takes too long to update (as in IE11), delay setting the focus
```
djwester: can we make this only happen if the browser is IE11 or is this also sometimes too slow on other browsers too?

djwester: Also, just a thought, but could we maybe fire and event or something so we don't have to wait an hardcoded amount of time? What if there are many groups? does this get slower? How do we know that 50 ms is enough time?

mdulat: Is there a way to observe dom changes? I thought there was some sort of dom-changed observer, but I can't immediately find a reference. If we can observe when dom changes, or even when elements within dom-repeat change length, that could be a trigger for setting focus.

emil-furniss: It only appears to affect IE11, so we could limit it to that. 
And yes, there is a `dom-change` event so I'll look into that.

emil-furniss:
Now using an event listener and the `dom-change` event to trigger refocusing.

editor/d2l-rubric-criteria-groups-editor.html
```diff
@@ -118,6 +115,15 @@
 			},
 			_showGroupName: function(entity) {
 				return entity.entities.length > 1;
+			},
+			_groupsDomComplete: function() {
+				this.fire('d2l-rubric-group-complete');
```
mdulat: Since this is all handled within the same component, you can use a boolean property to track whether to focus on dom change, instead of using an event. I.e. set it to true where you add the event listener, and set it to false where you remove the listener, then just do the focus within this function if property is true. In this case I think you'll also need to set it to false if the sirenAction fails (i.e. catch block).

emil-furniss: Good point, I clearly had my event blinkers on

editor/d2l-rubric-criteria-groups-editor.html
```diff
@@ -101,15 +106,12 @@
...
 					this.performSirenAction(action).then(function() {
 						this.fire('d2l-rubric-criteria-group-added');
 						setTimeout(function() {
 							this.fire('iron-announce', { text: this.localize('groupAdded') }, { bubbles: true });
 						}.bind(this), 2000);
-					}.bind(this)).then(function() {
-						var allGroups = Polymer.dom(this.root).querySelectorAll('d2l-rubric-criteria-group-editor');
-						var lastGroup = allGroups[allGroups.length - 1];
-						lastGroup.$$('d2l-text-input').$$('input').select();
 					}.bind(this));
```
mdulat: I think you should set _waitingForGroups to false if sirenAction fails, otherwise it'll focus last group next time dom changes.

mdulat: Does this work the same as using a catch block? Just make sure you test it with a failed siren response to make sure it works.

emil-furniss: You're probably right that using `catch` is safer, as it will catch errors thrown by the success function as well as if the promise is rejected.

PR #225, opened 2018-08-21
Add in a queue for siren actions that we are sending

alogan519:


assessment-result-behavior.html
```diff
@@ -8,6 +8,12 @@
 	window.D2L.PolymerBehaviors = window.D2L.PolymerBehaviors || {};
 	window.D2L.PolymerBehaviors.Rubric = window.D2L.PolymerBehaviors.Rubric || {};
 
+	// Create a promise queue for siren actions
+	window.D2L = window.D2L || {};
```
walinga: Looks like this line is the same as line 7

PR #245, opened 2018-08-30
US96675 add points values to levels

mdulat:


editor/d2l-rubric-level-editor.html
```diff
@@ -19,15 +19,26 @@
...
+				margin-left: 0.5rem;
 			}
 
 			:host-context([dir='rtl']) .operations {
 				text-align: left;
 			}
 
 			d2l-button-icon {
-				padding-top: 0.5rem;
+
```
emil-furniss: Is this empty rule intentional? And does it behave with rtl?

mdulat: oops, nope. Let me fix that up...

PR #248, opened 2018-09-04
Display criterion out-of value as text

emil-furniss:
https://rally1.rallydev.com/#/29180338367d/detail/task/251038054796
outof
This will only be functional after the LMS change to modify the criterion `update` action goes in, since it was decided not to read the value from the property: https://git.dev.d2l/projects/CORE/repos/lp/pull-requests/10716/overview

### DIFF
--- a/editor/d2l-rubric-criterion-editor.html
+++ b/editor/d2l-rubric-criterion-editor.html
@@ -88,6 +88,12 @@
 				padding: 0;
 			}
 
+			.out-of {
+				display: flex;
+				justify-content: center;
+				padding-top: 1.4rem;
+			}
+
 			.criterion-detail .criterion-feedback-header {
 				@apply --d2l-body-compact-text;
 				flex: 0 0 auto;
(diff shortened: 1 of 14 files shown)


editor/d2l-rubric-criterion-editor.html
```diff
@@ -313,8 +328,20 @@
 						this.fire('d2l-rubric-criterion-deleted');
 					}.bind(this));
 				}
-			}
+			},
+
+			_getOutOfValue: function(entity) {
+				if (this.hasOutOf) {
+					var action = entity.getActionByName("update");
+					if (action && action.fields[1] && action.fields[1].name === "outOf") {
```
mdulat: Is there a way to get a field by name instead of assuming it will be at position 1?

emil-furniss:
I've had to resort to ugly-but-reliable `for`-loops when getting the field and class values (to avoid using hard coded indices), to ensure IE11 compatibility

editor/d2l-rubric-criterion-editor.html
```diff
@@ -313,8 +336,27 @@
...
+			},
+
+			_getOutOfValue: function(entity) {
+				if (this.hasOutOf) return entity.properties.outOf;
+			},
 
+			_isOutOfEditable: function(entity) {
+				if (this.hasOutOf) {
+					var action = entity.getActionByName("update");
+					for (var i = 0; i < action.fields.length; i++) {
```
mdulat: Why not use action.getFieldByName()?

emil-furniss: didn't realize that was a thing

editor/d2l-rubric-criterion-editor.html
```diff
@@ -313,8 +336,26 @@
...
+			},
 
+			_isOutOfEditable: function(entity) {
+				if (this.hasOutOf) {
+					var action = entity.getActionByName("update");
+					if (!action) return false;
+					var field = action.getFieldByName("outOf");
+					if (field) {
+						for (var i = 0; i < field.class.length; i++) {
+							if (field.class[i] === "readonly") return false;
```
mdulat: you can use hasClass() method instead of for loop here

PR #249, opened 2018-09-05
Fixed score overrides to work on non-Chrome browsers

mpharoah-d2l:
* Simplified binding of blur event handler
* Defer focusing the text input box until we know it is actually rendered to avoid an immediate blur event in most browsers
* Trigger the focus automatically when the `editing-score` property changes instead of explicitly calling it

This PR fixes the score overrides being broken on non-Chrome browsers, as well as the total score not updating when changed.

### DIFF
--- a/d2l-rubric-criteria-group.html
+++ b/d2l-rubric-criteria-group.html
@@ -494,9 +494,6 @@
 				}
 				var criterionNum = event.model.get('criterionNum');
 				this.editingScore = criterionNum;
-				fastdom.mutate(function() {
-					Polymer.dom(this.root).querySelector('#score-inner' + criterionNum).focus();
-				}.bind(this));
 			}
 		});
 	</script>
--- a/d2l-rubric-editable-score.html
+++ b/d2l-rubric-editable-score.html
@@ -42,6 +42,7 @@
 				step="any"
 				min="0"
 				max="100000"
+				on-blur="_blurHandler"
 			>
 			</d2l-text-input>
 			<div class="right">[[_localizeOutOf(entity)]]</div>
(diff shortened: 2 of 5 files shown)


d2l-rubric-criteria-group.html
```diff
@@ -494,9 +494,6 @@
 				}
 				var criterionNum = event.model.get('criterionNum');
 				this.editingScore = criterionNum;
-				fastdom.mutate(function() {
```
mpharoah-d2l: This is now handled by the editable-score component when its editing-score property changes as a result of the above assignment, so we no longer need to explicitly focus the element with fastdom.

d2l-rubric-editable-score.html
```diff
@@ -121,25 +117,19 @@
...
-				}
-				elem.addEventListener('blur', this._boundBlurHandler);
-			},
-
-			detached: function() {
-				var elem = Polymer.dom(this.root).querySelector('d2l-text-input');
-				if (!elem) return;
-				elem.removeEventListener('blur', this._boundBlurHandler);
-			},
-
```
mpharoah-d2l: Adding and removing the event handlers this way was unnecessary and didn't seem to always work correctly. Just using the normal method of binding event handlers in Polymer works fine.

d2l-rubric-editable-score.html
```diff
@@ -121,25 +117,19 @@
...
 			focus: function() {
 				var elem = Polymer.dom(this.root).querySelector('d2l-text-input');
 				elem.focus();
 			},
 
+			_editingScoreChanged: function(newValue) {
+				if (this._isEditingScore(this.criterionNum, newValue)) {
+					Polymer.RenderStatus.afterNextRender(this, function() {
+						this.focus();
```
mpharoah-d2l: Wait until the text input is actually rendered before focusing it. This fixes the scores not being editable on Firefox.

PR #251, opened 2018-09-05
TA128117 Update total score UI

emil-furniss:
https://rally1.rallydev.com/#/29180338367d/detail/userstory/217784770628
total score design

### DIFF
--- a/editor/d2l-rubric-criteria-groups-editor.html
+++ b/editor/d2l-rubric-criteria-groups-editor.html
@@ -21,7 +21,7 @@
 				padding-bottom: 24px;
 			}
 
-			.group-footer {
+			.groups-footer {
 				display: flex;
 				flex-direction: row;
 				justify-content: space-between;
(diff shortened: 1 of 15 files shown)


editor/d2l-rubric-criteria-groups-editor.html
```diff
@@ -78,6 +89,15 @@
 				_waitingForGroups: {
 					type: Boolean,
 					value: false
+				},
+
+				_hasTotalScore: {
+					type: Boolean,
+					computed: '_hasTotalScore(entity)'
```
mdulat: This parameter "entity" doesn't have a corresponding argument in the function definition. To ensure _hasTotalScore gets updated when the entity is updated you could instead attach an observer to totalScore (which will be updated if entity is updated in rubric-editor), and have that observer update this._hasTotalScore appropriately (then you don't need the _hasTotalScore() function, just use the property).

editor/d2l-rubric-criteria-groups-editor.html
```diff
@@ -150,9 +151,8 @@
 				var lastGroup = allGroups[allGroups.length - 1];
 				lastGroup.$$('d2l-text-input').$$('input').select();
 			},
-			_hasTotalScore: function() {
-				if (this.totalScore) return true;
-				return false;
+			_totalScoreChanged: function(totalScore) {
+				totalScore ? this._hasTotalScore = true : this._hasTotalScore = false;
```
mdulat: This can be simplified to `this._hasTotalScore = totalScore;`

PR #252, opened 2018-09-07
Hacky fix for getting stuck in the saving... state

alogan519:


d2l-rubric.html
```diff
@@ -316,7 +316,11 @@
 				var saveStateEl = this.$['rubric-save-status'];
 				({
 					'd2l-siren-entity-save-start': function() { saveStateEl.start(); },
-					'd2l-siren-entity-save-end': function() { saveStateEl.end(); },
+					'd2l-siren-entity-save-end': function() {
+						// for Polymer 1 sometimes we aren't getting all end events so reset on end event
+						saveStateEl._saving = 1;
+						saveStateEl.end();
```
alogan519: the end() function will decrement the _saving counter by 1. 

I think this should be pretty safe to do since we are queuing our events anyway.

PR #254, opened 2018-09-07
Mbayly/editor horiz scrolling

mdgbayly:
This PR integrates the d2l-scroll-wrapper from d2l-table to enable a 'fancy' style of horizontal scrolling when there are too many levels to fit in the space allocated to the editor.

I had to tweak how to remove how we set the variable width on the criterion name, as I could not make this play nicely with the scrolling. I verified with Joseph, and the current width is what we agreed. 

I also replaced some of the min-widths we had been using on the first and last columns to use explicit `width` properties instead, as this is actually what we are intending for those columns (we don't expect them to resize as the window is resized and they never did anyway because of the way the flex properties are configured.

The scroll-wrapper is a bit flaky at times in that the buttons (mostly the right hand button) doesn't disappear correctly. This seems to be due to rounding errors with the way it is calculating the scroll offsets and will most likely need a fix to d2l-scroll-wrapper to resolve.

@mdulat @emil-furniss If you are setting any min-widths on the levels/criterion last columns for supporting points rubrics, I suggest you switch to using just a regular `width` property as well.

### DIFF
--- a/editor/d2l-rubric-criterion-editor.html
+++ b/editor/d2l-rubric-criterion-editor.html
@@ -96,8 +96,7 @@
 			}
 
 			.col-last[text-only] {
-				min-width: calc(1rem + 2.2rem); /* button width = 44px including border */
-				flex: 0 0 0%;
+				width: calc(1rem + 2.2rem); /* button width = 44px including border */
 			}
 
 			.col-last d2l-text-input {
--- a/editor/d2l-rubric-criteria-editor.html
+++ b/editor/d2l-rubric-criteria-editor.html
@@ -16,6 +16,7 @@
 		<style include="d2l-rubric-editor-cell-styles">
 			:host {
 				display: block;
+				flex: 1 1 auto;
 			}
 
 			* {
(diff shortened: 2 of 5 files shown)


editor/d2l-rubric-criterion-editor.html
```diff
@@ -102,8 +102,7 @@
 			}
 
 			.col-last[text-only] {
-				min-width: calc(1rem + 2.2rem); /* button width = 44px including border */
-				flex: 0 0 0%;
+				width: calc(1rem + 2.2rem); /* button width = 44px including border */
```
mdulat: I know the min-width was set this way before, but if this is accommodating the width of the button, shouldn't it be calculated as `1rem + 44px`, since the button width won't change if the font size changes? Same thing in the levels editor.

mdgbayly: @mdulat I think it is correct to use `rem` because i believe most of the d2l-button/button/icon styles are set using `rem`.

mdulat: You're right, looks like everything except d2l-icon uses rem. (so everything that embeds d2l-icon uses rem)

PR #259, opened 2018-09-11
fix up down arrow visibility not refreshing, and also

stevenao:
hide drag handle with it is the last item in the criteria group.


editor/d2l-rubric-criteria-editor.html
```diff
@@ -313,17 +313,18 @@
...
-
-			_isFirst: function(index) {
+			_hideDragHandle: function(index, criterionCount) {
+				return !this._canReorder || (this._isFirst(index,criterionCount) && this._isLast(index,criterionCount));
+			},
+			_isFirst: function(index, criterionCount) {
 				return index === 0;
 			},
-			_isLast: function(index) {
+			_isLast: function(index, criterionCount) {
```
mdgbayly: Seems kind of weird to be passing in criterionCount and then ignoring it?
I guess in the case of `_isFirst`, it's to force a re-evaluation when the number of criteria changes? For `isLast` it could probably have used the `criterionCount` property instead of looking at `_criteriaEntities.length`, but no big deal I guess.

stevenao: It is more of a way to force the _isFirst and _isLast on the arrow buttons to be re-evaluated. Joseph is seeing the buttons states not being updated when adding a new `criteria`. 

I trust `_criteriaEntities.length` more than the `dom-repeat.renderedCounter`. Not sure what optimization it is doing behind the scene for this rendering.

PR #263, opened 2018-09-12
add start range display for overall levels

stevenao:


editor/d2l-rubric-overall-levels-editor.html
```diff
@@ -96,7 +96,7 @@
 			}
 
 			.col-center div {
-				min-width: 7rem;
+				min-width: 10rem;
```
mdgbayly: I think this will conflict with some of the changes I made for the scrolling of the overall levels editor. I think I removed this min-width here and just allowed it to inherit the min-width that we use for the criterion cells to keep it consistent.

editor/d2l-rubric-overall-level-editor.html
```diff
@@ -20,15 +20,25 @@
...
-			:host-context([dir='rtl']) .operations {
-				text-align: left;
+			:host-context([dir='rtl']) #range-start {
+				margin-right: 0;
+				margin-left: 0.5rem;
 			}
 
 			d2l-button-icon {
-				padding-top: 0.5rem;
+				float: right;
```
mdgbayly: Is there a reason for using float rather than flex box.  float is so `noughties` ;-).
I think with flexbox you won't have to worry about `rtl`?

editor/d2l-rubric-overall-levels-editor.html
stevenao: yeah, I saw that. You mean to inherit from the overall-level-editor? Can I put the min-width in the overall-level-editor?

mdgbayly: My intention was that the min-width is the same for all 'cells'. So it would just be inherited from the `d2l-rubric-editor-cell-styles.html`. Does it need to be specified differently for the overall levels editor?

editor/d2l-rubric-overall-level-editor.html
```diff
@@ -20,15 +20,25 @@
 				box-sizing: border-box;
 			}
 			.operations {
-				text-align: right;
+				padding-top: 0.5rem;
+			}
+
+			#range-start {
+				width: 56px;
```
mdgbayly: Do we want to fix the size to 56px? Not sure what we're doing with the criteria group level points, but that seems like it would be too small for a multi digit point value, and should probably be specified in rem anyway? Maybe it fits two digits? Just took a look at the criteria group points styling. It looks like they added a `size` attribute of 1 which also seems to give a default size of 56px. That might be a better approach than specifying an actual pixel width. Although 1 char seems a bit small. Maybe the designers only expect points to ever be 1 or two digits, and I guess if 2 digits fit even when you set size to 1 that might be sufficient.

editor/d2l-rubric-overall-level-editor.html
stevenao: I am using `float` just for simplicity. It is more straight forward to me.

mdgbayly: I think it can cause the layout to wrap if the lang term for the `or more` is bigger than expected. Might be an edge case and open to question as to what is the preferred behaviour, but I think if you used flex, the trash can will stay to the right of the box and the label will wrap I think.

editor/d2l-rubric-overall-levels-editor.html
stevenao: there is more text (`or more`) in the overall-level cell. So, the width needs to be at least `10rem`. Not sure if other lang terms will take up more spaces.

editor/d2l-rubric-overall-level-editor.html
stevenao: yeah.. that gives me a reason to use flexbox now.

editor/d2l-rubric-overall-levels-editor.html
mdgbayly: Ok - the current min-width was originally set for text rubrics based on eyeballing having 4 levels fit into a 'standard' maximised LMS desktop display without causing scrolling. Such that if you add a fifth level, it scrolls rather than causing further compression of the level.

But yeh, if the current min is too small for overall levels, it might be too small for criteria levels which have points too. In which case we can update it. Or if it works better to keep them different I'm ok with that too.

PR #267, opened 2018-09-14
US96676 Handle level patch errors

emil-furniss:
When saving name or points fails and the error response contains details about the error as a json payload, display the message from that payload as the bubble error message and suppress the generic "We were unable to save your rubric" error alert.
If there are no details provided, display generic error alert and use "Saving points failed"/"Saving name failed" for the bubble.

### DIFF
--- a/editor/d2l-rubric-criteria-group-editor.html
+++ b/editor/d2l-rubric-criteria-group-editor.html
@@ -172,8 +172,13 @@
 
 			_handleSaveError: function(e) {
 				this._clearAlerts();
-				this.fire('iron-announce', { text: this.localize('rubricSavingErrorAriaAlert') }, { bubbles: true });
-				this._addAlert('error', e.detail.error.message, this.localize('rubricSavingErrorMessage'));
+				// only handle generic errors (without details)
+				try {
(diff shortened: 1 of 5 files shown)


editor/d2l-rubric-criteria-group-editor.html
```diff
@@ -172,8 +172,13 @@
 
 			_handleSaveError: function(e) {
 				this._clearAlerts();
-				this.fire('iron-announce', { text: this.localize('rubricSavingErrorAriaAlert') }, { bubbles: true });
-				this._addAlert('error', e.detail.error.message, this.localize('rubricSavingErrorMessage'));
+				// only handle generic errors (without details)
+				try {
+					var errObj = e.detail.error.string ? e.detail.error.string.properties : e.detail.error.json.properties;
```
mdulat: Why not just use an if/else to check whether error details exist? I found this confusing to follow - why we're setting a variable that's never used...

emil-furniss: This was my attempt at limiting long `if` conditions since there are a bunch of scenarios to deal with, but the more I look at it the uglier this gets.

PR #273, opened 2018-09-18
remove duplicate score when editing the score (seems to be only Polym…

alogan519:
…er 2)


alogan519:
best viewed ignoring whitespace changes

PR #276, opened 2018-09-19
Add in a tabindex for criterion cells and scores

alogan519:
Add tabindex for overall level overrides
Enter key should select the correct action - assess/unassess


mpharoah-d2l:
I think the whitespace change to/from tabs to/from spaces in `d2l-rubric-overall-score.html` or something.

alogan519:
Ya i think it's probably just changing all the line endings to lf

PR #279, opened 2018-09-24
US99693 - Remove archived text components

svanherk:
Remove `d2l-text-input` and `d2l-textarea` and use `d2l-inputs`'s `d2l-input-text` and `d2l-input-textarea` components instead.  That way, we can remove the archived repos from BSI.


mdgbayly:
@svanherk @alogan519 

If no-one has any complaints and assuming Stacey is ready for this to be merged, I'd like to get it in ASAP, so that we can:

a)  regression test early, and 
b) because I'm doing some other 'text input' refactoring so we can start supporting HTML editor in the rubric editor, so I'll get conflicts on that PR unless we get this one in first.

PR #285, opened 2018-09-25
Select all text when overriding score / total score

alogan519:


d2l-rubric-editable-score.html
```diff
@@ -126,6 +126,10 @@
 			focus: function() {
 				var elem = Polymer.dom(this.root).querySelector('d2l-input-text');
```
mpharoah-d2l: I think you can just use `var elem = this.$['text-area']` to get this

alogan519: so I think here I am looking for the d2l-input-text as there are some custom focus handling that happens (although probably would work your way as well).

mpharoah-d2l: Yeah, and it has id 'text-area' right? using `this.$.<id>` should give you the element with local ID <id>, so `this.$['text-area']` is equivalent to `Polymer.dom(this.root).querySelector('#text-area')`, but more efficient

mpharoah-d2l: There's only one `d2l-input-text` and it has id `text-area`, so they should be exactly equivalent

alogan519: oh ok I see what you are saying now.

PR #288, opened 2018-09-26
US96680 - HTML editor: Save and Load description content

wongvincent:
- If HTML is enabled, content saves when a description editor onBlur event occurs, matching the behavior of the textarea.

- Checked that HTML editor and non-html editor still save and load descriptions (including emojis) correctly.
- Checked that when creating description using html editor, then turning html-editor off and vice versa, content still shows up correctly. In the case where html-editor was on when it was created, then turned off, we will see the tags similar to what happens when editing an html post in Activity Feed `<p>abc</p>`
- Checked that description saving fail causes failed tooltip to appear.

Bugs:
- Selecting an emoticon to insert from the popup counts as a focus-out, so it's triggering a save

### DIFF
--- a/editor/d2l-rubric-html-editor.html
+++ b/editor/d2l-rubric-html-editor.html
@@ -73,7 +73,7 @@
 					 app-root=[[_appRoot]]
 					 lang-tag=[[language]]
 					 fullpage-enabled=0
-					 content=[[_encodeURIComponent(content)]]
+					 content=[[_encodeURIComponent(initValue)]]
 					 toolbar=[[_toolbar]]
 					 plugins=[[_plugins]]
 					 object-resizing=[[objectResizing]]>
@@ -84,6 +84,7 @@
 		role="textbox"
 		placeholder$=[[placeholder]]
 		aria-label$=[[ariaLabel]]
+		on-focus="_handleFocus"
 		on-blur="_handleBlur"></div>
 	</d2l-html-editor>
 	</template>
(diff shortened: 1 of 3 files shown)


editor/d2l-rubric-html-editor.html
```diff
@@ -103,6 +104,9 @@
 						return 'htmleditor-' + Date.now();
 					},
 				},
+				content: {
```
mdgbayly: Is this `content` property being used?

mdgbayly: Oh - maybe I'm just not following how content get's initialized from value? It looks like it happens the other way around?

wongvincent: `content` is used only once to set the content tinyMCE will initially load with.
Any changes afterwards, I am saving in a `value` property.

mdgbayly: Oh - ok - initializing content from value was in the stuff i copied from AF.

mdgbayly: Not sure whether we could get away with just having a single `value` property which we use to initilize the `d2l-html-editor`'s content property, or if that causes weird issues with re-initialization when we update the value property in the blur handler?

wongvincent: I've tried combining the `value` and `content` property.
So the initial value we pass to the html-editor is encoded with numeric references. Whereas, when we `_getContent()`, they are character references. e.g. numeric: `&#160;`, character `&nbsp;`.

So I think it's simpler and probably clearer to just leave them as 2 separate properties. I could probably rename the property so it's clearer that the value we pass in is only used when initializing the html-editor.

PR #294, opened 2018-09-28
Vwong/us96680 html editor user preferences

wongvincent:


editor/d2l-rubric-editor.html
```diff
@@ -86,6 +91,10 @@
 			is: 'd2l-rubric-editor',
 
 			properties: {
+				richTextEnabled: {
```
mdulat: Where does this value actually get set?

wongvincent: From RubricEdit.Renderer.cs in the LMS. When it creates `d2l-rubric-editor`, it includes `token` and `href` attributes, so now I'm also passing a `rich-text-enabled` attribute if the user account setting has not disabled rich text.

mdulat: Ah, right. Sweet.

PR #295, opened 2018-10-01
Hide the html description when there isn't one since it adds an extra…

alogan519:
… line of padding basically

### DIFF
--- a/d2l-rubric-overall-score.html
+++ b/d2l-rubric-overall-score.html
@@ -151,7 +151,7 @@
 					<span class="overall-level-text">
 						<span>[[_localizePoints(level)]]</span>
 						<br hidden="[[!_scoreVisible(level)]]">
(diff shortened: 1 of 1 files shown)


d2l-rubric-overall-score.html
```diff
@@ -151,7 +151,7 @@ <h4>
 					<span class="overall-level-text">
 						<span>[[_localizePoints(level)]]</span>
 						<br hidden="[[!_scoreVisible(level)]]">
-						<s-html html$="[[_getDescriptionHtml(level)]]"></s-html>
+						<s-html hidden="[[!_hasDescriptionHtml(level)]]" html$="[[_getDescriptionHtml(level)]]"></s-html>
```
mpharoah-d2l: You could just do `hidden="[[!_getDescriptionHtml(level)]]"` instead of adding a new function

alogan519: Sure that sounds good

PR #304, opened 2018-10-03
Fix html-editor not updating to reflect content after cell shifted

wongvincent:
Pass a key that is unique for description (criterion and overallevel) cell. It is based off the entity self href or level + self href (in the case of criterionCell as they are always 0, 1, 2...)
When a cell is shifted and the html-editor is re-used, the shifted cell will pick up a new key value. This will notify the html-editor that it needs to re-set content to reflect its new value.

### DIFF
--- a/editor/d2l-rubric-description-editor.html
+++ b/editor/d2l-rubric-description-editor.html
@@ -67,6 +67,7 @@
 		</div>
 		<d2l-rubric-text-editor
 			id="description"
+			key="[[_computedKey]]"
 			aria-invalid="[[_isAriaInvalid(_descriptionInvalid)]]"
 			aria-label="[[_getAriaLabel(ariaLabelLangterm, criterionName, entity.properties)]]"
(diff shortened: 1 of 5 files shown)


editor/d2l-rubric-description-editor.html
```diff
@@ -67,6 +67,7 @@
 		</div>
 		<d2l-rubric-text-editor
 			id="description"
+			key="[[_computedKey]]"
```
mdgbayly: Maybe we should call this `key-link-rels` or something to make it clear it is not the actual key?

editor/d2l-rubric-description-editor.html
```diff
@@ -99,6 +100,14 @@
 					type: String,
 					value: ''
 				},
+				key: {
```
mdgbayly: Maybe we should call this `key-link-rels` or something to make it clear it is not the actual key?

PR #306, opened 2018-10-03
US99856 remove duplicate rubric and add info to describe rubric types.

mdulat:


demo/d2l-rubric-editor.html
```diff
@@ -60,28 +60,23 @@ <h3>Rubric Editor, text only, single group, with edit permissions</h3>
 		<d2l-rubric-editor href="data/rubrics/organizations/text-only/242.json" token=""></d2l-rubric-editor>
 	</demo-snippet>
 
-	<h3>Rubric, multiple groups</h3>
-	<demo-snippet>
-		<d2l-rubric-editor href="data/rubrics/organizations/multiple-groups/203.json" token="foozleberries"></d2l-rubric-editor>
```
emil-furniss: So no more text-only, multiple-group rubric?

mdulat: That one wasn't text-only, it was identical to the other multiple-group one

PR #315, opened 2018-10-04
trello: fix double delete click

stevenao:


editor/d2l-rubric-criterion-editor.html
```diff
@@ -403,15 +403,20 @@
 				return '';
 			},
 
-			_handleDeleteCriterion: function() {
+			_handleDeleteCriterion: function(e) {
 				var action = this.entity.getActionByName('delete');
 				if (action) {
+					var deleteButton = e.currentTarget;
+					var deleteButtonDisableState = deleteButton.disabled;
+					deleteButton.disabled = true;
```
emil-furniss: Why can't we just set `disabled = true` here, and then in the `finally` block `disabled = false`?

stevenao: yeah, was trying to play defensively. But now, I think if it was previously disabled. There won't be anything button click triggered anyway. Will update it.

emil-furniss: and we're sure that a disabled button doesn't fire events?

stevenao: yup, it won't trigger the event.

PR #316, opened 2018-10-04
Use 'Range Start' messages for holistic rubric percentages

emil-furniss:
Error messages taken from overall levels, as approved by Joseph.

### DIFF
--- a/editor/d2l-rubric-level-editor.html
+++ b/editor/d2l-rubric-level-editor.html
@@ -150,6 +150,10 @@
 				_showPoints: {
 					type: Boolean,
 					computed: '_computeShowPoints(hasOutOf, entity)'
+				},
+				_usesPercentage: {
+					type: Boolean,
+					computed: '_computeUsesPercentage(entity)'
 				}
 			},
 			observers: [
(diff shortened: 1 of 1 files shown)


editor/d2l-rubric-level-editor.html
```diff
@@ -198,15 +205,15 @@
...
+					if (this._usesPercentage) {
 						return '%';
 					}
 					return this.localize('pointsAbbreviation');
 				}
 				return '';
 			},
 			_showAltPercentFormat: function(percentageFormatAlternate, entity) {
-				return entity && entity.hasClass(this.HypermediaClasses.rubrics.percentage) && percentageFormatAlternate;
+				return entity && this._usesPercentage && percentageFormatAlternate;
```
mdulat: This function would be cleaner now if it took usesPercentage as a parameter instead of entity, since it doesn't really need entity anymore (since it's checked in the computation of usesPercentage).

emil-furniss: Good point

editor/d2l-rubric-level-editor.html
```diff
@@ -212,8 +212,8 @@
 				}
 				return '';
 			},
-			_showAltPercentFormat: function(percentageFormatAlternate, entity) {
-				return entity && this._usesPercentage && percentageFormatAlternate;
+			_showAltPercentFormat: function(percentageFormatAlternate) {
```
mdulat: I think if you pass usesPercentage as a parameter instead of using `this.` then the function will get re-evaluated when that property changes. Otherwise it won't, and I believe we will in future allow people to change rubric types and such on the fly.

PR #317, opened 2018-10-05
Alogan/out of one line

alogan519:


d2l-rubric-levels-mobile.html
```diff
@@ -33,6 +32,7 @@
 				color: var(--d2l-color-galena);
 				outline: none;
 				flex: 1;
+				height: 32px;
```
alogan519: fixes an issue on mobile where input is larger than 32 px

PR #318, opened 2018-10-05
Fire an event when the total score changes

alogan519:


alogan519:
Event fires when data is loaded (goes from null to a number), number is a string:

event.detail.score

PR #322, opened 2018-10-09
US100352 - Lang terms DA and FI

wongvincent:
Creating template for DA and FI. They have not been translated and shouldn't be until we wrap up rubrics, at which point this and the other lang files will be translated.

### DIFF
--- a/lang/da.html
+++ b/lang/da.html
@@ -11,7 +11,7 @@
 	* @polymerBehavior D2L.PolymerBehaviors.Rubric.LocalizeBehavior.LangDaBehavior
 	 */
 	D2L.PolymerBehaviors.Rubric.LocalizeBehavior.LangDaBehavior = {
-		da: {
+		daDk: {
 			'addCriteriaGroup': 'Add Criteria Group',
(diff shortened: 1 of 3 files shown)


lang/da.html
```diff
@@ -0,0 +1,94 @@
...
+	window.D2L.PolymerBehaviors = window.D2L.PolymerBehaviors || {};
+	window.D2L.PolymerBehaviors.Rubric = window.D2L.PolymerBehaviors.Rubric || {};
+	window.D2L.PolymerBehaviors.Rubric.LocalizeBehavior = window.D2L.PolymerBehaviors.Rubric.LocalizeBehavior || {};
+
+	/*
+	* Da lang terms
+	* @polymerBehavior D2L.PolymerBehaviors.Rubric.LocalizeBehavior.LangDaBehavior
+	 */
+	D2L.PolymerBehaviors.Rubric.LocalizeBehavior.LangDaBehavior = {
+		daDk: {
```
xraysaas: what's the reason why this was changed from `da`?

wongvincent: test failing when named `da` :(

PR #324, opened 2018-10-10
fix duplicate invalid tooltips

mdulat:
Fixes https://trello.com/c/xWQadvGj/43-for-the-required-field-we-are-still-seeing-the-duplicate-error-message
Note it did not require changes to d2l-inputs after all, as that component already contained the code that was supposed to handle this. For some reason that code no longer worked because the 'invalid' event was no longer being trapped. 
Instead, this solution stops setting the 'required' field of the text input and textarea components, since the only thing setting that was accomplishing was getting those unwanted tooltips. The red borders are obtained by setting 'aria-invalid' instead.

### DIFF
--- a/editor/d2l-rubric-criterion-editor.html
+++ b/editor/d2l-rubric-criterion-editor.html
@@ -406,14 +406,16 @@
 				var action = this.entity.getActionByName('delete');
 				if (action) {
 					var deleteButton = e.currentTarget;
-					deleteButton.disabled = true;
+					deleteButton.setAttribute('disabled', '');
 					var name = this.entity.properties.name;
 					this._transitionElement(this, 0);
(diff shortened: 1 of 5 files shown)


editor/d2l-rubric-criterion-editor.html
```diff
@@ -165,8 +165,7 @@
 				hover-styles
 				value=[[entity.properties.name]]
 				placeholder="[[_getNamePlaceholder(localize, displayNamePlaceholder)]]"
-				on-change="_saveName"
-				required="[[_nameRequired]]">
```
emil-furniss: Does this mean that we could remove all the other stuff attached to this property? (and in the other files as well?)

mdulat: What other stuff? Like the _nameRequired stuff? No, because we use it to determine whether to display the tooltip and set the aria-invalid.

PR #325, opened 2018-10-11
ensure HTML editor save events are handled consistently by preventing…

mdgbayly:
… actions that add/remove levels/criteria from triggering a form submission which causes TinyMCE to fire an unnecessary save event. This can cause a criterion cell to be saved to the wrong cell index. Also removes the need to compare HTML on save.

@wongvincent I discovered that some of the weird behaviour you had been seeing with things saving/rendering inconsistently appears to be mostly caused by a tinymce change event that is triggered as a result of tinymce listening on form submit events.  And when we clicked on buttons like our add/delete level and delete criterion, this was causing a form submit event. 

To work around this I added `type="button"` attributes to a number of the relevant buttons. 

I didn't added it to the Add Criterion or Add Criterion Group buttons as they didn't seem to be causing weird behavior.

I do think there is still the potential for issues if other things cause unexpected save events when the rubric is addding/removing levels/criterion due to the signature of the criterion cell API (being index based). But provided we can avoid those unexpected events I think we'll be good.

### DIFF
--- a/editor/d2l-rubric-text-editor.html
+++ b/editor/d2l-rubric-text-editor.html
@@ -99,18 +99,17 @@
 			_onInputChange: function(e) {
 				e.stopPropagation();
 				var value = (e.detail && e.detail.content) || e.target.value || '';
-				if (this._decodeHtml(value) !== this._decodeHtml(this.value)) {
-					this.debounce('fireInputChange', function() {
-						this.fire('change', { value: value });
-					}, 1000);
 
-					var activeElementId = document.activeElement && document.activeElement.id;
-					var focusedOnInput = activeElementId && e.target.querySelector('#' + activeElementId);
-					// If it is a blur event (no longer focused on input), fire change event immediately to
-					// avoid race conditions with other events.
-					if (!focusedOnInput) {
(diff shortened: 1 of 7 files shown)


editor/d2l-rubric-text-editor.html
```diff
@@ -99,18 +99,17 @@
...
-					// If it is a blur event (no longer focused on input), fire change event immediately to
-					// avoid race conditions with other events.
-					if (!focusedOnInput) {
-						this.flushDebouncer('fireInputChange');
-					}
+				this.debounce('fireInputChange', function() {
+					this.fire('change', { value: value });
+				}, 1000);
+
+				var activeElementId = document.activeElement && document.activeElement.id;
```
mdgbayly: @wongvincent I'm a little bit uneasy about this blur handling, because I don't think it will work if shadow DOM is enabled. However, given that won't be for a while yet, I'm OK to let it go. I think we probably could put in code to make it work with shadow roots too, but it's a little bit confusing how to make that work, and given it's not going to be easy to test anyway I guess we can wait.

PR #329, opened 2018-10-13
the html editor toolbars are breaking the drag image for chrome becau…

mdgbayly:
…se they are absolute positioned above the criterion, even when they are not visible. To workaround, explicitly hide/show the toolbar container when the editor is hidden/shown

### DIFF
--- a/editor/d2l-rubric-criteria-editor.html
+++ b/editor/d2l-rubric-criteria-editor.html
@@ -48,10 +48,7 @@
 			}
 
 			.dnd-placeholder::after {
-				/* These margins need to match the width of the left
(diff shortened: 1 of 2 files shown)


editor/d2l-rubric-criteria-editor.html
```diff
@@ -48,10 +48,7 @@
 			}
 
 			.dnd-placeholder::after {
-				/* These margins need to match the width of the left
```
mdgbayly: Not part of the drag image fix, but I noticed these were off and that we now had a property for them so fixed at the same time.

PR #330, opened 2018-10-16
include insert stuff in rubric html editor

mdgbayly:


editor/d2l-rubric-html-editor.html
```diff
@@ -49,6 +49,10 @@
 		d2l-html-editor .d2l-richtext-editor-container > p:first-of-type {
 		  margin-top: 0;
 		}
+
```
mdgbayly: This max-width style doesn't actually have any effect yet due to the shadow scoping rules.
I'm gong to have to add a similar style to the main rubrics page, similar to how we remove the top/bottom margins from the first/last paragraphs.

But including here for good measure as it may work better once we move to true shadow DOM and declares the intent.

The reason for this is because embedding large iframes like YouTube videos seem to break the flex box sizing, causing some level columns to be too big.

PR #341, opened 2018-10-24
Add aria alerts for validation errors

emil-furniss:
Addresses https://trello.com/c/hQPZZ7pI/20-a11y-add-an-aria-alert-on-validation-error

Previously there was no a11y handling of validation errors. Now when validation of a field fails, it will be announced using `iron-announce` with the same lang term as the tool-tip.

For components with more than 2 validation error handling locations, I added a dedicated `_handleValidationError` to toggle the error tool-tip and make the announcement.

(Also found and fixed a small bug in the overall levels, where the name of the `update-range-start` action had not been updated, so the required property was not being set correctly).

### DIFF
--- a/editor/d2l-rubric-overall-level-editor.html
+++ b/editor/d2l-rubric-overall-level-editor.html
@@ -184,14 +184,14 @@
 				var action = this.entity.getActionByName('update-name');
 				if (action) {
 					if (this._nameRequired && !e.target.value.trim()) {
-						this._handleValidationError('_nameInvalid', 'nameIsRequired')
+						this._handleValidationError('_nameInvalid', 'nameIsRequired');
 					} else {
 						this._toggleBubble('_nameInvalid', false);
 						var fields = [{ 'name': 'name', 'value': e.target.value }];
 						this.performSirenAction(action, fields).then(function() {
 							this.fire('d2l-rubric-overall-level-saved');
 						}.bind(this)).catch(function() {
-							this._handleValidationError('_nameInvalid', 'nameSaveFailed')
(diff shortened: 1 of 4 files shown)


editor/d2l-rubric-overall-level-editor.html
```diff
@@ -200,23 +200,33 @@
...
+				this.fire('iron-announce', { text: msgText }, { bubbles: true });
+			},
+			_getErrMsg: function(e, altMsg) {
+				if (e && !e.hasOwnProperty('stack')) {
+					var errObj = e.string ? e.string : e.json;
+					if (errObj.hasOwnProperty('properties')) {
+						return errObj.properties.detail;
+					}
+				}
+				console.log("Alt: " + altMsg);
```
mdulat: Remove console.log stmt before merging.

editor/d2l-rubric-overall-level-editor.html
```diff
@@ -184,14 +184,14 @@
 				var action = this.entity.getActionByName('update-name');
 				if (action) {
 					if (this._nameRequired && !e.target.value.trim()) {
-						this._toggleBubble('_nameInvalid', true, this.localize('nameIsRequired'));
+						this._handleValidationError('_nameInvalid', 'nameIsRequired')
```
mdulat: This and other related statements below are missing semi-colon at the end. Do you have eslint plugin in your editor? It should complain about that.

emil-furniss: Oh wow that's messy, I'm so sorry. And yes, I have it installed but it didn't complain - I'll look into that.

PR #342, opened 2018-10-24
Fix total score not showing up if it has value zero

mdulat:
Fixes https://trello.com/c/XNns2jHW/17-rubrics-when-total-score-is-0-total-score-is-hidden

### DIFF
--- a/editor/d2l-rubric-criteria-groups-editor.html
+++ b/editor/d2l-rubric-criteria-groups-editor.html
@@ -100,7 +100,7 @@
 				},
 
 				totalScore: {
-					type: Number,
+					type: String,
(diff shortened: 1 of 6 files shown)


mdulat:
Vincent, the problem with your suggestion is that the lang term 'dashOutOf': '\u2014 / {outOf}' treats an outOf value of 0 as a null, and makes that an empty string.

wongvincent:
> Vincent, the problem with your suggestion is that the lang term 'dashOutOf': '\u2014 / {outOf}' treats an outOf value of 0 as a null, and makes that an empty string. (I tried something similar initially too).

Ah ok

PR #343, opened 2018-10-24
fixes alignment problem on mobile of header

mdulat:
Fixes https://trello.com/c/MdgGZqb3/22-mobile-android-and-iphone-portrait-view-scoring-and-reverse-buttons-look-messy by putting the right gutter after a block surrounding the two header components instead of after the last component of the header.


emil-furniss:
Does this mean that the header will look like this
`left gutter`
`header content`
`right gutter`
`rest of rubric`
as opposed to this
`left gutter` `header element`
`header element` `right gutter`
`rest of rubric`
? If so, nice solution!

PR #358, opened 2018-11-09
Unfocusing from the edit score text field focus the criterion score.

ihair:


d2l-rubric-criteria-group.html
```diff
@@ -169,6 +169,7 @@
 									read-only="[[readOnly]]"
 									editing-score="{{editingScore}}"
 									criterion-num="[[criterionNum]]"
+									parentElement="[[editableScoreContainer]]"
```
apalaniuk: Since this collides with [`Node.parentElement`](https://developer.mozilla.org/en-US/docs/Web/API/Node/parentElement), this will either need to be renamed, or some other method of either getting the element itself (pass in a selector/predicate that the child uses with [`D2L.Dom.findComposedAncestor'](https://github.com/Brightspace/d2l-polymer-behaviors-ui/blame/8f57a045d1dcccc62bd003a5075fec0daab8368d/README.md#L31), or maybe fire a custom event that the ancestor listens to that resets focus, etc.

d2l-rubric-editable-score.html
```diff
@@ -105,6 +105,10 @@
 				readOnly: {
 					type: Boolean,
 					value: true
+				},
+				parentElement: {
```
apalaniuk: Since this collides with [`Node.parentElement`](https://developer.mozilla.org/en-US/docs/Web/API/Node/parentElement), this will either need to be renamed, or some other method of either getting the element itself (pass in a selector/predicate that the child uses with [`D2L.Dom.findComposedAncestor`](https://github.com/Brightspace/d2l-polymer-behaviors-ui/blame/8f57a045d1dcccc62bd003a5075fec0daab8368d/README.md#L31), have the ancestor set the event listener on the descendant, or maybe have the descendant fire a custom event that the ancestor listens to that resets focus, etc.)

mpharoah-d2l:
You have a linting error to fix. You can run lint locally with `npm run test:lint`

d2l-rubric-editable-score.html
```diff
@@ -106,7 +106,7 @@
 					type: Boolean,
 					value: true
 				},
-				parentElement: {
+				parentNode: {
```
apalaniuk: this is closed, but FYI - [I think this would have the same issue](https://developer.mozilla.org/en-US/docs/Web/API/Node/parentNode)

PR #369, opened 2018-11-21
fix console error

mdulat:
image


omsmith:
This shouldn't be the case. If you'd like there used to be an if statement to handle either case. Or it could be `Promise.resolve` around entity. But the version that should be getting installed (is this on your local and your bower_components are out of date?) is definitely a Promise here

mdulat:
> This shouldn't be the case. If you'd like there used to be an if statement to handle either case. Or it could be `Promise.resolve` around entity. But the version that should be getting installed (is this on your local and your bower_components are out of date?) is definitely a Promise here

Yes, after fighting with my local BSI to update, this is now fixed. I will close this PR.

omsmith:
:+1:

PR #375, opened 2018-11-23
Display hide score checkbox

emil-furniss:


emil-furniss:
@mdulat Yes, saving/calling the API will be a separate story. Sorry, I should have put that in the description.
And as far as I can tell it doesn't need an aria-label, as the `<label for="">` construct takes care of the labeling, but I'll make sure to check.

PR #377, opened 2018-11-23
Ihair/focus criterion score when edit score text field closed

ihair:


d2l-rubric-editable-score.html
```diff
@@ -106,6 +106,10 @@
 				readOnly: {
 					type: Boolean,
 					value: true
+				},
+				parentNode: {
```
apalaniuk: [So I think this will have the exact same issue as parentElement because parentNode is a property of an HTMLElement (Node)](https://developer.mozilla.org/en-US/docs/Web/API/Node/parentNode). If I fix the linting error and run tests, I get

```
Test run ended in failure: Error thrown outside of test function: Cannot assign
to read only property 'parentNode' of object '#<PolymerGenerated>'
  HTMLElement._initializeProperties at /components/polymer/lib/mixins/element-mi
xin.html:531
```

Maybe `focusableAncestor` (or something better)?

d2l-rubric-editable-score.html
```diff
@@ -153,6 +157,7 @@
 				if (event.keyCode === 13) { // enter key
 					event.target.blur();
 					event.stopPropagation();
+					setTimeout(function(){ this.parentNode.focus(); }.bind(this), 0);
```
apalaniuk: May want to check that `parentNode`* (rather, its renamed-to-avoid-collisions-version) actually exists, as it seems not-impossible to be using the mobile set of elements on a device with a keyboard  (eg. shrunk screen on a low-res laptop maybe?) where `parentNode`* isn't set

Or, maybe this actually grab the host's `parentElement` and try to focus it, rather than passing it down

PR #379, opened 2018-11-23
Vwong/US102102

wongvincent:
This PR
 - Creates "Edit Rubric" **title** for single page rubrics
- Moves **save-status** to be beside title
- Moves **status dropdown** to be beside save-status and visible only if it is a single page rubric
- Moved the dropdown menu functions into its own behavior so that it isn't duplicated in the header and in the rubric-structure-editor.

Will work on moving error handling into the header in another PR, as this PR is already a bit too massive. Sorry!


wongvincent:
Closed this because I created a new PR that uses slots instead of this complicated mess

PR #380, opened 2018-11-23
Vwong/us102102

wongvincent:
This PR
 - (Single page rubric only) Renders page **title** "Edit Rubric"
- Renders **save-status** beside title
- (Single page rubric only) Renders **status dropdown** beside save-status
- Moved the dropdown menu functions into its own behavior so that it isn't duplicated in the header and in the rubric-structure-editor.

Will work on adding error handling into the header in another PR, as this PR is already a bit too massive. Sorry!


editor/d2l-rubric-editor.html
```diff
@@ -62,7 +68,41 @@
 				padding-top: 0.5rem;
 			}
 
+			#status-dropdown {
```
mdgbayly: I'm confused by these duplicated `#status-dropdown` declarations? why do we set `margin-right` twice for example

wongvincent: good catch, it shouldn't be duplicated!

mdgbayly: ok - same for the rtl ones?

PR #382, opened 2018-11-26
Ihair/added clear feedback icon

ihair:


d2l-rubric-feedback.html
```diff
@@ -156,6 +180,11 @@
 				this.saveAssessmentFeedback(this.criterionHref, feedback);
 			},
 
+			_clearFeedback: function() {
+				this.saveFeedback("");
```
mpharoah-d2l: You need to use single quotes here or it will fail linting

d2l-rubric-feedback.html
```diff
@@ -117,6 +135,12 @@
 						return this._blurHandler.bind(this);
 					}
 				},
+				_boundClearFeedback: {
```
mpharoah-d2l: This looks to be unnecessary

d2l-rubric-criteria-group.html
```diff
@@ -266,6 +266,10 @@
 				'_updateFeedbackDisplay(_criteriaEntities, assessmentResult)'
 			],
 
+			ready: function() {
+				this.addEventListener('close-feedback', this._closeFeedback.bind(this));
```
mpharoah-d2l: Instead of adding an event listener this way, you can add `on-close-feedback="_closeFeedback"` to the `<d2l-rubric-feedback>` element in the HTML above.

ihair: So pass it in as a property and then call it within d2l-rubric-feedback?

mpharoah-d2l: You already have `d2l-rubric-feedback` emitting an `close-feedback` event. If you put an `on-close-feedback` attribute on the `<d2l-rubric-feedback>`s in the `d2l-rubric-criteria-group` with its value being the name of a function you have defined in `d2l-rubric-criteria-group`, Polymer will automatically wire it up to handle that event, so you don't need to manually add the event listener yourself and manage its lifetime.

mpharoah-d2l: (You don't need to add anything to `<d2l-rubric-feedback>`. If you put an attribute starting with "on-" on an element in Polymer, it will automatically add an event handler for the event name matching the attribute (minus the "on-" part) for you)

ihair: I see, that's pretty cool, thanks!

PR #388, opened 2018-11-28
Replaced overall score's clear override div with a subtle button.

ihair:


d2l-rubric-overall-score.html
```diff
@@ -305,6 +305,14 @@ <h4>
 				);
 			},
 
+			_clearOverride: function() {
```
mpharoah-d2l: This shouldn't be necessary. The cell already has an action that will clear the override when clicked. You don't need to add another one to the button. (Or is the button preventing the click event from propagating?)

mpharoah-d2l: Actually, do we really want this to be a button? The whole cell is clickable, not just the text, so a button doesn't really make sense

mpharoah-d2l: Oh wait, this is overall score, not overall level. Nevermind, I was thinking of something else

ihair: The designers want it to be a subtle button

PR #395, opened 2018-11-30
US102258 add help dialog for associations

mdulat:


lang/ar.html
```diff
@@ -24,6 +24,7 @@
 			'changeScoringSuccessful': 'تم تغيير طريقة وضع الدرجات إلى {method}',
 			'clearOverride': 'مسح التجاوز',
 			'clearTotalScoreOverride': 'مسح التجاوز',
+			'clearFeedback': 'Clear Feedback',
```
emil-furniss: What is this lang term for?

mdulat: It's used in d2l-rubric-feedback.html, and it was missing in every lang file except en.

PR #396, opened 2018-12-01
Mbayly/add rubric visibility

mdgbayly:
Includes changes to support:

1. Saving rubric by calling action
2. Error handling
3. Accessibility support 

The component uses a debouncer on the save, because it's possible to hold down the arrow keys to change the selected radio button very fast, and this would lead to hundreds of save calls being made.

Also removed the attempt at utilizing a web component for the radio buttons in favour of a simple style sheet which I am going to look at moving into BrightspaceUI. Radio buttons just don't work as web components!

The rubric visibility radio buttons are implemented in a self-contained web component. This was done to for general code modularity reasons (the d2l-rubric-editor is getting pretty busy already) and to make it easier to integrate with the tooltip. We want to show a single tooltip for the combined radio buttons. Hiding/showing the tooltip on focus/blur is made easier by having the tooltip associated with a component rather than trying to manage the blur/focus for each button independently.

Because the buttons are in a child component, i had to modify the overall error handling. Previously components had code like `if e.target === this` which didn't work when the errors were generated by a child component. I've removed those guards in favour of stopping propagation of error events when they are handled by an appropriate component. This prevents us getting multiple alerts displayed at different levels which I believe was the reason for the guard anyway.


editor/d2l-rubric-editor.html
```diff
@@ -408,10 +408,8 @@
 			},
 			_handleSaveError: function(e) {
 				this._clearAlerts();
-				if (e.target === this) { // handle own error events
-					this.fire('iron-announce', { text: this.localize('rubricSavingErrorAriaAlert') }, { bubbles: true });
-					this._addAlert('error', e.detail.error.message, this.localize('rubricSavingErrorMessage'));
-				}
+				this.fire('iron-announce', { text: this.localize('rubricSavingErrorAriaAlert') }, { bubbles: true });
```
mdulat: Should this have a e.stopPropagation(); also?

mdgbayly: Technically doesn't need it because this is the top of the hierarchy.

PR #402, opened 2018-12-04
pass updateAssociations param when updating associations.

mdulat:
Goes with https://git.dev.d2l/projects/CORE/repos/lp/pull-requests/11775/overview to fix 
https://trello.com/c/csuF2VfR/13-rubrics-single-page-when-you-create-a-new-rubric-or-edit-existing-rubric-and-change-the-name-associations-get-unchecked


mdulat:
@mdgbayly we don't represent associations as properties currently, only within the action.

PR #409, opened 2018-12-06
Provide the visibility radio buttons tooltip a 'for' target

emil-furniss:
Wrapped the radio buttons in a div, so that the tooltip component can have a sensible `for` target.
Also excluded `d2l-rubric-structure-editor` from handling validation errors so those don't cause alerts to pop up.
This shouldn't affect the behavior of the component in any way.


editor/d2l-rubric-structure-editor.html
```diff
@@ -344,8 +344,10 @@
 			_handleSaveError: function(e) {
 				this._clearAlerts();
 				e.stopPropagation();
-				this.fire('iron-announce', { text: this.localize('rubricSavingErrorAriaAlert') }, { bubbles: true });
-				this._addAlert('error', e.detail.error.message, this.localize('rubricSavingErrorMessage'));
+				if (e.target === this) {
```
mdgbayly: Not clear why this is needed? Adding `e.target === this` breaks some kinds of alerts from child components? What scenario is it breaking without it?

emil-furniss: without it validation errors throw alerts, which they're not supposed to.

editor/d2l-rubric-visibility-editor.html
```diff
@@ -40,23 +40,25 @@
...
+			<label class="d2l-input-radio-label">
+				<input type="radio" id="NeverVisible" on-focus="_handleFocus" on-blur="_handleBlur" value="NeverVisible" name="visibility" on-change="_changeVisibility"></input>[[localize('rubricVisibilityNever')]]
+			</label>
+			<label class="d2l-input-radio-label">
+				<input type="radio" id="VisibleOnceFeedbackPosted" on-focus="_handleFocus" on-blur="_handleBlur" value="VisibleOnceFeedbackPosted" name="visibility" on-change="_changeVisibility"></input>[[localize('rubricVisibilityOnceFeedbackPosted')]]
+			</label>
+		</div>
 		<template is="dom-if" if="[[_visibilityInvalid]]">
-			<d2l-tooltip position="bottom">[[_visibilityInvalidError]]</d2l-tooltip>
+			<d2l-tooltip position="bottom" for="visibility-container">[[_visibilityInvalidError]]</d2l-tooltip>
```
mdgbayly: Also not clear on why this is needed as the tooltip is smart enough to handle creating a tooltip for a host component? Originally the intention was to hide/show the tooltip when the component in error was focused? Are we changing that behaviour.

editor/d2l-rubric-structure-editor.html
mdgbayly: Under what scenario? e.stopPropagation is supposed to prevent the error bubbling? Is that not working in some browsers?

editor/d2l-rubric-visibility-editor.html
emil-furniss: Yeah, we're changing the behavior. If there are errors on the page, exactly one tooltip should always be visible, even if none of the elements are in focus. For that I need to be able to pick out the appropriate tooltip from within a component, and for that I need the `for` to be set.

editor/d2l-rubric-structure-editor.html
mdgbayly: Actually maybe I'm not understanding what you are saying - let's chat after i'm done this meeting.

editor/d2l-rubric-visibility-editor.html
mdgbayly: Probably need to chat about how you are doing this as we need to make sure it works with shadow DOM too.
If you are doing anything that is programatically manipulating elements by ID that might not work with shadow DOM.

editor/d2l-rubric-structure-editor.html
mdgbayly: how are we distinguishing validation errors from other errors currently? What kind of validation errors?  Sounds like we should be preventing those validation errors from bubbling, rather than suppressing errors in this component?  This `_handleSaveError` is supposed to show alerts for errors that are not handled elsewhere e.g. if we get an API error trying to add a criterion group?

mdgbayly: Ok - Iooking at it again, the reason why these errors are getting this far is because I added the `e.stopPropagation` in the wrong place in `d2l-rubric-criteria-group-editor`. We have code in there which is skipping the showing of the alert at the criteria group level if the error is a 'validation' error. So the call to `e.stopPropagation` should just be moved to the top of that method, then errors won't get here. This was supposed to be a catch all error handler for errors that are not handled more specifically in lower level components. By restricting it to check for `this` it breaks that feature.

But stepping back a little, I kind of think we should stop showing alerts at all for `d2l-siren-entity-error`. That was our original error handling strategy for things like adding criterion and adding criteria groups where we didn't have a good way to show a more specific error bubble.  I think at that stage performSirenAction wasn't even returning a promise.

If we have specific calls to performSirenAction where we don't already have our own custom error handling, then maybe we should raise a specific catch all rubric save error in those locations that can bubble up and cause an alert to be shown. For any instance where we actually highlight a field as being in error it seems superfluous to show the error alert too. Let's discuss with Joseph.

That would simplify things as we wouldn't then have to be trying to suppress `d2l-siren-entity-error` in cases where we actually handle the error and show an error bubble.

PR #431, opened 2019-01-17
add quotation mark and changed overall feedback font

iFun:
not sure img path is correct or not. the maze.svg is not viewable in the demo page


djwester:
the maze.svg is test data and is a broken link on purpose.

PR #432, opened 2019-01-17
add quotation mark and changed overall feedback font

iFun:


iFun:
close due to polymer 3 conversion

PR #433, opened 2019-01-17
Polymer 3 Conversion

stevenao:


wongvincent:
"The proof will be in the pudding!"
(spoiler: it works sorta :P)

PR #438, opened 2019-01-21
minor fix that was triggering rubric description tooltip with no error

wongvincent:


wongvincent:
When the error is null, d2l-tooltip sees the textContent and innerText as `""` after trimming. https://github.com/BrightspaceUI/tooltip/blob/master/d2l-tooltip.js#L302 is `true` and the function is returned before `showing = true` is executed. If we pass `false`, d2l-tooltip sees `dom(this).textContent` as `undefined`, and so the function continues to that line where it sets `showing` to `true`.

PR #439, opened 2019-01-22
Fix focus not returning to rubric type dropdown button after closing LMS confirm dialog

wongvincent:
https://trello.com/c/3KLZUuyJ/79-rubrics-focus-resets-to-window-after-analytic-holistic-warning-popup-is-closed

### DIFF
--- a/editor/d2l-rubric-structure-editor.js
+++ b/editor/d2l-rubric-structure-editor.js
@@ -396,6 +396,8 @@ Polymer({
 				negativeButtonText: this.localize('changeConfirmationNo')
 			}).then(performAction, function() {
 				this.resetSelectedMenuItem(menuButton, this._rubricTypeValue);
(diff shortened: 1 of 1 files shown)


editor/d2l-rubric-structure-editor.js
```diff
@@ -396,6 +396,8 @@ Polymer({
 				negativeButtonText: this.localize('changeConfirmationNo')
 			}).then(performAction, function() {
 				this.resetSelectedMenuItem(menuButton, this._rubricTypeValue);
+			}.bind(this)).finally(function() {
```
emil-furniss: Last time I checked we couldn't use `finally` because it's not supported by IE. As a workaround in the past we've used `then(foo)` with a separate `catch(foo)`.

wongvincent: come on IE :( 👎

PR #447, opened 2019-01-23
fix edge and ie space before pt

stevenao:


mdgbayly:
I'm still a bit unclear on what is the 'recommended' way to do this with Polymer 3...

Dave B seemed to be saying just use :dir(rtl) but the polymer 3 docs he referenced say use

`:host(:dir(rtl))`

https://polymer-library.polymer-project.org/3.0/docs/devguide/style-shadow-dom#style-directional-text-with-the-dir-selector

Or am i misreading them?

PR #456, opened 2019-02-01
UI changes for US101718

dabdullahd2l:
https://rally1.rallydev.com/#/57732444928d/detail/userstory/264313782352?fdp=true


mpharoah-d2l:
I think we can close this PR since we're using Polymer 3 going forward :)

PR #457, opened 2019-02-01
UI changes for US101718

dabdullahd2l:
https://rally1.rallydev.com/#/57732444928d/detail/userstory/264313782352?fdp=true

### DIFF
--- a/d2l-rubric-feedback.js
+++ b/d2l-rubric-feedback.js
@@ -25,7 +25,6 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-feedback">
 				height: 0;
 				border: 12px solid transparent;
 				position: absolute;
-				-moz-box-sizing: border-box;
 				transition-property: border-color;
 				transition-timing-function: ease;
 				transition: border-color 0.5s;
@@ -47,7 +46,6 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-feedback">
 				border-right: 12px solid transparent;
 				border-bottom: 12px solid white;
 				z-index: 1;
-				-moz-box-sizing: border-box;
 			}
 			:host([_focus-styling]) .feedback-arrow-inner {
 				top: 3px;
--- a/d2l-rubric-feedback.js
+++ b/d2l-rubric-feedback.js
@@ -193,7 +193,7 @@ Polymer({
 		feedbackWrapper.style.cursor = 'text';
 		feedbackWrapper.style.padding = 'calc(0.5rem - 1px)';
 		feedbackWrapper.style.paddingLeft = 'calc(1rem - 1px)';
-		feedbackArrow.style.borderBottomColor = '#006fbf';
+		feedbackArrow.style.borderBottomColor = 'var(--d2l-color-celestine)';
 		feedbackArrowInner.style.top = '3px';
 	},
 
@@ -206,7 +206,7 @@ Polymer({
 			feedbackWrapper.style.padding = '0.5rem';
 			feedbackWrapper.style.paddingLeft = '1rem';
 			feedbackWrapper.style.cursor = 'default';
-			feedbackArrow.style.borderBottomColor = '#d3d9e3';
+			feedbackArrow.style.borderBottomColor = 'var(--d2l-table-border-color)';
 			feedbackArrowInner.style.top = '2px';
 		}
 	},
--- a/d2l-rubric-feedback.js
+++ b/d2l-rubric-feedback.js
@@ -56,6 +56,7 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-feedback">
 				border-radius: 5px;
 				padding: 5px;
 				border: solid transparent 1px;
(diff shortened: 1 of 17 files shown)


djwester:
There are a few linting errors to clean up. @dabdullahd2l

lang/en.js
```diff
@@ -27,6 +27,7 @@ D2L.PolymerBehaviors.Rubric.LocalizeBehavior.LangEnBehavior = {
 		'criterionAriaLabel': 'Criterion {index, number} of {count, number}',
 		'criterionDeleted': '{name} criterion deleted',
 		'criterionDescriptionAriaLabel': 'Description for criterion {criterionName}, level {levelName}',
+		'criterionFeedback': 'Criterion Feedback',
```
djwester: I think we need to add this to all the language files (just put in english for now). I will submit it for translation and then we can update that other languages once we get them back.

I think though, if we don't have the key in all the language files we might get errors if people change their language

d2l-rubric-feedback.js
```diff
@@ -24,22 +24,33 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-feedback">
 				width: 0;
 				height: 0;
 				border: 12px solid transparent;
-				border-bottom-color: var(--d2l-table-border-color);
 				position: absolute;
+				-moz-box-sizing: border-box;
```
mpharoah-d2l: Is there a reason you're using the Firefox/Gecko specific styling here instead of just `box-sizing` which works in all browsers?

d2l-rubric-feedback.js
```diff
@@ -24,22 +24,33 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-feedback">
...
-				top: 1px;
+				top: 2px;
 				left: -12px;
 				width: 0;
 				height: 0;
 				border-left: 12px solid transparent;
 				border-right: 12px solid transparent;
 				border-bottom: 12px solid white;
 				z-index: 1;
+				-moz-box-sizing: border-box;
```
mpharoah-d2l: Same here

d2l-rubric-feedback.js
```diff
@@ -136,9 +153,30 @@ Polymer({
...
+		},
 		addingFeedback: {
 			type: Boolean,
 			value: false
+		},
+		_feedbackInFocus: {
+			type: Boolean,
+			value: false
+		},
+		_focusStyling: {
```
mpharoah-d2l: With the changes to `d2l-tspan`, is this still necessary? I think you can use `d2l-tspan[_feedback-in-focus] .feedback-wrapper` instead of setting a property/attribute and using `:host([_focus-styling]) .feedback-wrapper`

dabdullahd2l: `_focusStyling` is set to true when `feedback-wrapper` is hovered as well so `_feedback-in-focus` alone won't work (it's only flipped to true when the textarea has focus). 
It's also being used to turn on the styling for blue arrow border when `feedback-wrapper` is hovered, this is because I can't turn on the styling for blue arrow border using `:hover` pseudo-selector for the arrow classes. If I use the `:hover` pseudo-selector, the blue arrow border styling only turns on when I am specifically hovering the arrow area instead of the whole `feedback-wrapper`.
Plus this selector doesn't work.

d2l-rubric-feedback.js
dabdullahd2l: removed both.

d2l-rubric-feedback.js
```diff
@@ -153,24 +190,50 @@ Polymer({
 			return;
 		}
 		elem.addEventListener('blur', this._boundBlurHandler);
+		var feedbackWrapper = dom(this.root).querySelector('.feedback-wrapper');
+		feedbackWrapper.addEventListener('focusout', this._boundFocusOutHandler);
```
mpharoah-d2l: Can you just put the `on-focusin` and `on-focusout` attributes on the element instead of manually added and removing these listeners? Or does Polymer not support annotated event listeners for these events?

d2l-rubric-feedback.js
```diff
@@ -153,24 +190,50 @@ Polymer({
...
+		this._removeFocusStylingFromFeedbackWrapper();
+	},
+
 	_blurHandler: function(event) {
 		var feedback = event.target.$.textarea.value;
 		this.saveFeedback(feedback);
 	},
 
+	_addFocusStylingToFeedbackWrapper: function() {
+		this._focusStyling = true;
```
mpharoah-d2l: This function and the one below it look like they are only called from one place. Since it's such a simple function, you can just move this into the _focusInHandler and _focusOutHandler functions

dabdullahd2l: I have them separate to differentiate between hover and focus. Using them with `on-mouseover` and `on-mouseout` attributes.

d2l-rubric-feedback.js
dabdullahd2l: Done!

PR #469, opened 2019-03-04
US102575 added outcome to rubric editor

iFun:


editor/d2l-rubric-criterion-editor.js
```diff
@@ -424,5 +463,34 @@ Polymer({
 
 	_getCellKeyRels: function() {
 		return [this.HypermediaRels.Rubrics.level, 'self'];
-	}
+	},
+
+	//hide browser outcomes button when holistic or text only or LD flag is off
```
djwester: typo in the comment (hide browse button, not hide browser button).

Which LD flag controls this?

iFun: the LD flag is rubrics-outcomes-alignments. and it used by the backend API so if LD flag is off there will be no activity usage api.  will fix the typo

editor/d2l-rubric-criterion-editor.js
```diff
@@ -424,5 +463,34 @@ Polymer({
...
-	}
+	},
+
+	//hide browse outcomes button when holistic or text only or LD flag is off
+	_canHideBrowseOutcomesButton: function(entity) {
+		const isFlagOff = entity &&
+			this.HypermediaRels.Activities &&
+			this.HypermediaRels.Activities.activityUsage &&
+			entity.getLinkByRel(this.HypermediaRels.Activities.activityUsage);
+		return !isFlagOff || this._hasOutOf || this.isHolistic;
```
mpharoah-d2l: You shoud pass `_hasOutOf` and `isHolistic` in as a parameters rather then referencing `this` so that Polymer knows it needs to compute those values first before this one

editor/d2l-rubric-criterion-editor.js
```diff
@@ -424,5 +463,34 @@ Polymer({
...
+	},
+
+	_getOutcomeRel: function() {
```
mpharoah-d2l: Same here, pass in `_hideBrowseOutcomesButton` as a parameter so that dependencies are resolved in the correct order

editor/d2l-rubric-criterion-editor.js
iFun: The _hasOutOf is always undefined in my function. The isHolistic works.

PR #470, opened 2019-03-11
Update lang terms for 'criterion feedback'

djwester:


djwester:
@dabdullahd2l we don't actually support Finish anymore (you'll notice that the file is entirely in English). It is just there for backwards compatibility at this point.

PR #477, opened 2019-03-18
check to see if outcome is empty

iFun:
hide the browser outcome button when there is not outcome to show

### DIFF
--- a/editor/d2l-rubric-criterion-editor.js
+++ b/editor/d2l-rubric-criterion-editor.js
@@ -297,12 +297,12 @@ $_documentContainer.innerHTML = /*html*/`<dom-module id="d2l-rubric-criterion-ed
 					<d2l-button-icon id="remove" hidden$="[[!_canDelete]]" icon="d2l-tier1:delete" text="[[localize('removeCriterion', 'name', entity.properties.name)]]" on-tap="_handleDeleteCriterion" type="button"></d2l-button-icon>
 				</div>
 				</div>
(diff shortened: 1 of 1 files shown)


editor/d2l-rubric-criterion-editor.js
```diff
@@ -527,13 +532,14 @@ Polymer({
...
-	_canHideBrowseOutcomesButton: function(entity, _hasOutOf, isHolistic, _isAlignmentListEmpty) {
+	// hide browse outcomes button when holistic or text only or LD flag is off or tag list is empty
+	// there is no outcome to browser
+	_canHideBrowseOutcomesButton: function(entity, _hasOutOf, isHolistic, _isAlignmentListEmpty, _isOutcomeEmpty) {
 		const isFlagOff = entity &&
 			this.HypermediaRels.Activities &&
 			this.HypermediaRels.Activities.activityUsage &&
 			entity.getLinkByRel(this.HypermediaRels.Activities.activityUsage);
-		return !isFlagOff || !_hasOutOf || isHolistic || !_isAlignmentListEmpty;
+		return !isFlagOff || !_hasOutOf || isHolistic || !_isAlignmentListEmpty || _isOutcomeEmpty;
```
mpharoah-d2l: Don't you want it to be hidden when the outcome tag list is _not_ empty? Since then it uses the plus button on the tag list

iFun: that is for !_isAlignmentListEmpty
this is when there is no outcome at all(so when user open the outcome list the list is empty)

PR #480, opened 2019-03-22
US102690 Add Outcomes Indicator

rgoessl:


demo/data/alignments/activity-usage/text-only_rcriterion_242-R-765/text-only.json
```diff
@@ -0,0 +1 @@
+{}
```
iFun: is this intentional?

rgoessl: I was too lazy to make more test data.
But it shows that if no alignments are returned for a criterion, the indicator won't be displayed

PR #490, opened 2019-04-11
fix for outcomes tooltips not disappearing on mouseout

dabdullahd2l:


editor/d2l-rubric-error-handling-behavior.js
```diff
@@ -86,8 +86,10 @@ D2L.PolymerBehaviors.Rubric.ErrorHandlingBehavior = {
 		if (cache.bubble) {
 			cache.bubble.forceShow = false;
 		}
-
-		target.forceShow = true;
+		// filter out d2l-tooltips that are not being used for errors
+		if (target.offsetParent && target.offsetParent.nodeName !== 'D2L-MULTI-SELECT-LIST-ITEM') {
```
emil-furniss: Is there a reason we can't add this filter at the very top of this function and skip this function entirely? I'm not sure about running parts of `_onBubbleShow` on non-error tooltips (setting the cached tooltip, stopping propagation etc). Something like `if( isNotErrorTooltip ) return;` on line 80?

dabdullahd2l: Moved it up.
Also found another place with the same issue. I am setting an attribute that `is-not-error` and checking for that in the if condition. Not adding another property to `d2l-tooltip` since this seems to only be an issue with rubric-editor error handling behavior. Let me know if you prefer another way.

mpharoah-d2l:
Are the tooltips that _are_ errors all defined in this repo? If so, it would be better to add and attribute or class to those instead of ones that aren't errors

emil-furniss:
> Are the tooltips that _are_ errors all defined in this repo? If so, it would be better to add and attribute or class to those instead of ones that aren't errors

Agreed. Yes, all error tooltips for the editor are defined in this repo and there aren't that many either.

PR #492, opened 2019-04-17
Rubric ui improvements

dabdullahd2l:
Fixed:
1) Bottom border for selected cell for the last criteria in a rubric
2) Clicking anywhere on the rubric-feedback component should focus the textarea and apply focus styling

### DIFF
--- a/d2l-rubric-criteria-group.js
+++ b/d2l-rubric-criteria-group.js
@@ -446,7 +446,6 @@ Polymer({
 		var selfLink = this._getSelfLink(criterionCell);
 
 		var noBottom = noBottomCells && noBottomCells[selfLink];
-		var isLastCell = criterionNum === criteria.length - 1;
 		var hasFeedback = this._hasFeedback(criteria[criterionNum], assessmentResult) || criterionNum === addingFeedback;
 
 		// A cell already has a bottom border in the following cases:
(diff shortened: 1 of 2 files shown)


d2l-rubric-criteria-group.js
```diff
@@ -93,6 +93,9 @@ $_documentContainer.innerHTML = /*html*/`<dom-module id="d2l-rubric-criteria-gro
 				box-shadow: -1px 0 0 var(--d2l-color-celestine), 0 1px 0 var(--d2l-color-celestine);
 				z-index: 1; /* Need bottom border to render over feedback cell border */
 			}
+			.criterion-cell.selected.is-last {
```
mpharoah-d2l: Could you just use the CSS selector `:last-child` instead?

dabdullahd2l: Doesn't work in this case since the parent in this case would be `d2l-tr` and so the last child would be the last cell in the row? 
I want this styling to be applied to the last cell in the column. Just tried `.criterion-cell.selected:last-child`

mpharoah-d2l: You could probably do something like `d2l-tr:last-child  .criterion-cell.selected`, but this is fine too

PR #493, opened 2019-04-22
pass in config variable for outcomesToolIntegration from LMS

dabdullahd2l:


mpharoah-d2l:
Is `reflectToAttribute` needed? You are passing it in, so that should not be required- it's more useful for computed properties

dabdullahd2l:
> Is `reflectToAttribute` needed? You are passing it in, so that should not be required- it's more useful for computed properties

Yep, not needed. Removed it.

PR #496, opened 2019-05-06
US105054 -- Show competencies in new rubrics experience

mpharoah-d2l:
Only supporting Desktop mode right now. Will add mobile view later once we have a design


d2l-rubric-criteria-group.js
```diff
@@ -497,6 +512,43 @@ Polymer({
 		return className;
 	},
 
+	_hasCompetencies: function(assessmentEntity, criterion, readOnly) {
+		return !readOnly && !!this._getCriterionCompetencies(assessmentEntity, criterion).length;
```
djwester: I feel like this function is mis-named - It is more _showCompetencies then _has.  They can be there, but we don't want to show them in read-only views.

mpharoah-d2l: I'll rename this in my next PR when I add mobile view.

PR #498, opened 2019-05-09
Show learning outcomes in mobile view

mpharoah-d2l:


d2l-rubric-alignments-indicator.js
```diff
@@ -82,6 +92,12 @@ class RubricAlignmentsIndicator extends mixinBehaviors([
 		`;
 	}
 
+	ready() {
+		super.ready();
+		this.addEventListener('blur', () => this._hasFocus = false);
```
gfrancu: Do we not care about removing these listeners on unload?

mpharoah-d2l: `ready` only fires the first time the component is attached to the DOM. We don't need to explicitly remove the listeners because when there are no more references to the element, the garbage collector will deal with it

PR #500, opened 2019-05-13
Remove orphaned “Name” label for locked rubric editor

kevinwang-d2l:


djwester:
I think these PRs are failing because we can't access sauce on a Fork

PR #505, opened 2019-05-14
Make cells into tab stops only when interactive

kevinwang-d2l:
In static views, the rubric table cells are not interactive, so they shouldn’t be focusable. This included the locked rubric view, rubric preview, and graded view.


d2l-rubric-criteria-group.js
```diff
@@ -605,6 +605,16 @@ Polymer({
...
-	}
+	},
+
+	_isStaticView: function() {
+		return this.readOnly || !this.assessmentHref;
+	},
 
+	_handleTabIndex: function() {
+		if (this._isStaticView()) {
+			return;
```
mpharoah-d2l: I would explicitly put `return undefined;` for clarity

kevinwang-d2l: thanks for the suggestion! updated

PR #508, opened 2019-05-21
US105635 -- Fixed hardcoded text

mpharoah-d2l:
LMS side: https://git.dev.d2l/projects/CORE/repos/platformtools/pull-requests/3089/overview

### DIFF
--- a/editor/d2l-rubric-criterion-editor.js
+++ b/editor/d2l-rubric-criterion-editor.js
@@ -243,7 +243,7 @@ $_documentContainer.innerHTML = /*html*/`<dom-module id="d2l-rubric-criterion-ed
 
 		<simple-overlay id="overlay" tabindex="-1" scroll-action="lock" class="with-margin scrollable"  with-backdrop>
 			<div>
-				<h2>Browse Outcomes </h2>
+				<h2>[[browseOutcomesText]]</h2>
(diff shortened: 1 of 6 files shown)


editor/d2l-rubric-criterion-editor.js
```diff
@@ -243,7 +243,7 @@ $_documentContainer.innerHTML = /*html*/`<dom-module id="d2l-rubric-criterion-ed
 
 		<simple-overlay id="overlay" tabindex="-1" scroll-action="lock" class="with-margin scrollable"  with-backdrop>
 			<div>
-				<h2>Browse Outcomes </h2>
+				<h2>[[browseOutcomesText]]</h2>
```
djwester: can't we just keep the 'Browse' and replace the 'outcomes' part of this with [[outcomesTitle]]. So something like:

`<h2> Browse [[outcomesTitle]]</h2>`

mpharoah-d2l: Because then we'd end up with things like 'Browse 標準'

djwester: well, yeah, but can't we localize Browse?

mpharoah-d2l: Word order isn't necessarily going to be consistent across languages

PR #509, opened 2019-05-21
setup d2l-rubric to use serge

djwester:


djwester:
there are new translations that should get in and we missed the serge run already.  I think this is correct?  Also, Can I get rid of the files in the old location?

d2l-rubric.serge.json
```diff
@@ -0,0 +1,21 @@
+{
+    "name": "d2l-rubric",
+    "source_dir": "lang",
+    "output_file_path": "lang/%LANG%.json",
```
rgoessl: Shouldn't this bt `build/lang/%LANG%.json`

rgoessl: Never mind

sarahboyd:
> there are new translations that should get in and we missed the serge run already. I think this is correct? Also, Can I get rid of the files in the old location?

I think serge only outputs json files, so we might need to change the files in`/lang` to be json instead of javascript and then do something like Dawood wanted to do for activity-alignments  (https://github.com/Brightspace/d2l-activity-alignments/pull/79)

mpharoah-d2l:
Looks like this is how its done for webcomponents:
https://github.com/BrightspaceHypermediaComponents/d2l-sequence-navigator/blob/master/localize-behavior.js

The language files are stored as JSON, then a script is run that autogenerates a javasript file for localization

djwester:
I have updated the `/lang` files to be `.json`

@mpharoah-d2l it looks like the link you shared has the translation hard coded in the function.  I'll look at how other components are doing it and see if there are any good 'generators' for the lang files that we can use, but for now I'm thinking we get this PR in so that we can kick off Serge to send the translations before it gets too late in the release cycle.

lang/ko.json
```diff
@@ -0,0 +1,135 @@
+import '@polymer/polymer/polymer-legacy.js';
```
rgoessl: What's wrong with this file?

lang/tr.json
```diff
@@ -0,0 +1,135 @@
+import '@polymer/polymer/polymer-legacy.js';
```
rgoessl: Not json

lang/ko.json
djwester: committed before saving - should be good now

PR #511, opened 2019-05-27
mobile grading rubric style improvements

kevinwang-d2l:
Capture2


kevinwang-d2l:
made some updates after getting feedback from rebecca

PR #513, opened 2019-05-31
change tab order of grading rubric 

kevinwang-d2l:
- Selecting a level in the rubric should be like selecting an option from a set of radio buttons. 
- The sequence of tab stops in the rubric should be as follows:


d2l-rubric-criteria-group.js
```diff
@@ -617,5 +639,38 @@ Polymer({
 			return undefined;
 		}
 		return 0;
+	},
+
+	_handleInvisibleFeedbackFocusin: function(event) {
+		var criterionNum = event.model.get('criterionNum');
+		var elem = dom(this.root).querySelector('#addFeedback' + this._getRowIndex(criterionNum));
+		elem.className = 'feedback-button-focused';
```
mpharoah-d2l: A slightly nicer way to do this is `elem.classList.add( 'feedback-button-focused'` and `elem.classList.remove( 'feedback-button-focused'`

kevinwang-d2l: thanks for the advice, changed accordingly

PR #528, opened 2019-07-11
WCAG US108465

kevinwang-d2l:
1. stop save-status from interrupting others in screen reader
2. set focus to add feedback button after clearing feedback
3. add aria-label to add feedback button 
4. aria-hidden on tooltip


d2l-rubric-criteria-group.js
```diff
@@ -367,8 +367,14 @@ Polymer({
...
 	},
 
-	_closeFeedback: function() {
+	_closeFeedback: function(event) {
 		this._addingFeedback = -1;
+		const criterionNum = +event.currentTarget.dataset.criterion;
+		var elem = dom(this.root).querySelector('#invisible-addFeedback' + this._getRowIndex(criterionNum));
+		this.async(function() {
+			elem.focus();
+		}.bind(this), 500);
```
djwester: Why do we need the wait here?

PR #535, opened 2019-08-12
[rubric] [ui] Render shell component w/ button to launch OU picker

stevenao:


editor/d2l-rubric-editor.js
```diff
@@ -343,6 +345,10 @@ $_documentContainer.innerHTML = /*html*/`<dom-module id="d2l-rubric-editor">
 							<d2l-link on-click="_openHelpDialog" href="">[[localize('helpAssociations')]]</d2l-link>
 						</div>
 					</div>
+					<div id="make-rubric-available-container" hidden$="[[!_canShare]]">
+						<label>[[localize('makeRubricAvailableHeader')]]</label>
```
mdulat: Is this accessible?

build/lang/en.js
```diff
@@ -31,7 +31,7 @@ D2L.PolymerBehaviors.Rubric.LocalizeBehavior.LangEnBehavior = {
 		'criterionNameAriaLabel': 'Criterion name',
 		'criterionOutOf': 'Criterion {name} is out of {value} points',
 		'criterionPlaceholder': 'Click to edit criterion',
-		'dashOutOf': '\u2014 / {outOf}',
+		'dashOutOf': '— / {outOf}',
```
mdulat: Why change these from unicode?

stevenao: the lang:build tool did it automatically. I am going to ask in the channel.

editor/d2l-rubric-editor.js
stevenao: I think label probably doesn't need aria help. Anyways, I think it is more of a placeholder rather than having anything functional yet.

mdulat: I just mean is it even read by the screen reader? Maybe it doesn't matter just yet, but should be checked once it's more than a placeholder.

build/lang/en.js
stevenao: "—" === '\u2014'
true

if you look at other lang json files, they are using the string presentation actually, so I guess it is just a different way of writing it.

PR #536, opened 2019-08-21
fixing rubric editor not saving properly when input is in focus

kevinwang-d2l:
`on-input` event occurs immediately after modification, unlike the `on-change` event, which occurs when the element loses focus

### DIFF
--- a/editor/d2l-rubric-criterion-editor.js
+++ b/editor/d2l-rubric-criterion-editor.js
@@ -261,7 +261,7 @@ $_documentContainer.innerHTML = /*html*/`<dom-module id="d2l-rubric-criterion-ed
 		<div style="display:flex; flex-direction:column;">
 			<div style="display:flex">
 				<div class="cell col-first criterion-name" hidden$="[[isHolistic]]">
-					<d2l-input-textarea id="name" aria-invalid="[[isAriaInvalid(_nameInvalid)]]" aria-label$="[[localize('criterionNameAriaLabel')]]" disabled="[[!_canEdit]]" value="[[entity.properties.name]]" placeholder="[[_getNamePlaceholder(localize, displayNamePlaceholder)]]" on-change="_saveName">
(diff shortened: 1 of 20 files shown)


editor/d2l-rubric-criterion-editor.js
```diff
@@ -261,7 +261,7 @@ $_documentContainer.innerHTML = /*html*/`<dom-module id="d2l-rubric-criterion-ed
 		<div style="display:flex; flex-direction:column;">
 			<div style="display:flex">
 				<div class="cell col-first criterion-name" hidden$="[[isHolistic]]">
-					<d2l-input-textarea id="name" aria-invalid="[[isAriaInvalid(_nameInvalid)]]" aria-label$="[[localize('criterionNameAriaLabel')]]" disabled="[[!_canEdit]]" value="[[entity.properties.name]]" placeholder="[[_getNamePlaceholder(localize, displayNamePlaceholder)]]" on-change="_saveName">
+					<d2l-input-textarea id="name" aria-invalid="[[isAriaInvalid(_nameInvalid)]]" aria-label$="[[localize('criterionNameAriaLabel')]]" disabled="[[!_canEdit]]" value="[[entity.properties.name]]" placeholder="[[_getNamePlaceholder(localize, displayNamePlaceholder)]]" on-change="_saveName" on-input="_saveName">
```
djwester: Does this mean we are saving with every keystroke?

kevinwang-d2l: yea, it's pretty much equivalent to `keydown`

djwester: hmm, that might lead to a lot of network traffic, since we would need to call the API every time we save right?

kevinwang-d2l: hmm good point, guess I should add `debounce `....

kevinwang-d2l:
I'm getting the same test failures on the master branch... @djwester @mpharoah-d2l  not sure what is going on... 🤔

djwester:
I wonder if those tests were broken by the changes @stevenao made.  Looks like things were committed with a red build last week and the tests didn't even run.

Steven can you take a look at this?

PR #542, opened 2019-09-23
DE36087 - fix rubric input autosave

xiangqiyan:


xiangqiyan:
Gonna have to make some more changes. Seems like this broke adding levels when rich text is enabled. Also if the api call to update the rubric takes longer than 0.5s to complete then there could still be problems with new input changes being overwritten.

PR #547, opened 2019-10-01
DE36245 - Fix feedback not saving on ipad

xiangqiyan:


xiangqiyan:
I'm pretty sure this issue is caused by the rubric being embedded in an iframe in the assess rubric popup. On iOS clicking the close button doesn't cause the feedback input to blur and I think it's because the button is part of a different window.

PR #561, opened 2019-10-28
Localization updates

apalaniuk:
- Add `fr-fr` to automated build and reprocess files to reduce language-checkr errors
- Move localization scripts as `lang` prefix was causing `language-checkr` to analyze contained JSON files


build/lang/fi.js
```diff
@@ -77,7 +77,7 @@ D2L.PolymerBehaviors.Rubric.LocalizeBehavior.LangFiBehavior = {
 		'pointsAreRequired': 'Point value is required',
 		'pointsMinimum': '{number} {number, plural, one {point} other {points}} minimum',
 		'pointsSaveFailed': 'Saving points failed',
-		'percentage': '\{number\} %',
```
apalaniuk: change: looks like this was a bug in the translation

PR #563, opened 2019-10-29
Rubric assessment error and autosave fix

xiangqiyan:


djwester:
We need to get this in so that we can test and ship a hotfix. Please merge even with broken tests and follow up with another PR later to fix the tests.

PR #577, opened 2019-11-13
[20.19.11/12] DE36773 -- Fixed trailing spaces being removed from rubric title while still typing

mpharoah-d2l:


djwester:
I think Xiangqi just added some of the code being removed for making it so we don't lose data if the internet connection goes out. Do these changes still support that workflow?

mpharoah-d2l:
If the internet goes out between you clicking on the field and you clicking out of it, the changes will be lost, but that's pretty standard behaviour

xiangqiyan:
I think this will cause inputs to be lost if you click out of it then click back and change the text before the save completes

djwester:
"standard behaviour" that snhu seems to violently object to

mpharoah-d2l:
> 
> 
> I think this will cause inputs to be lost if you click out of it then click back and change the text before the save completes

It will save the original value when you clicked out. If you change it before it saves, the new value will get saved when you click out again

xiangqiyan:
> > I think this will cause inputs to be lost if you click out of it then click back and change the text before the save completes
> 
> It will save the original value when you clicked out. If you change it before it saves, the new value will get saved when you click out again

Wouldn't it overwrite the new value with the original value when the save completes due to the update?

apalaniuk:
This could result in lost data in some cases, eg. if the user just closes the tab/browser as the next action after input.

editor/d2l-rubric-editor.js
```diff
@@ -909,5 +923,8 @@ Polymer({
 		if (isLocked) {
 			import('../d2l-rubric.js');
 		}
+	},
+	_getShownRubricName: function(isFocused, enteredName, actualName) {
+		return isFocused ? enteredName : actualName;
```
xiangqiyan: If the user saves by clicking out of the input, it will display the old value until the save completes. And if they focus the input again before the save finishes, it will cause the old value to be displayed until the save finishes and the user clicks out of the input again. Might cause issues if the user's connection is slow.
I think we could either get rid of saving while input is changing but keep the input changed flag and pending save counter so it doesn't update while the user is typing again or just don't update the property with the entity value after saving since the only modification seems to be trimming the whitespace.

mpharoah-d2l: Good catch. Fixed it to not revert to showing the old value while a save is pending

editor/d2l-rubric-editor.js
```diff
@@ -728,7 +728,7 @@ Polymer({
 		}
 
 	},
-	_nameFocusHandler: function(e) {
+	_nameFocusHandler: function() {
 		this._enteredRubricName = this._rubricName;
```
xiangqiyan: I think _enteredRubricName should only be set if the name is not invalid and there are no pending saves otherwise it will clear any changes made if the save request fails and the user clicks the input again to retry.

editor/d2l-rubric-editor.js
```diff
@@ -924,7 +924,7 @@ Polymer({
 			import('../d2l-rubric.js');
 		}
 	},
-	_getShownRubricName: function(isFocused, enteredName, actualName) {
-		return isFocused ? enteredName : actualName;
+	_getShownRubricName: function(isFocused, pendingSaves, enteredName, actualName) {
+		return (isFocused || pendingSaves > 0) ? enteredName : actualName;
```
xiangqiyan: enteredName should also be shown if _nameInvalid is true.

editor/d2l-rubric-level-editor.js
```diff
@@ -417,7 +417,7 @@ Polymer({
 			deleteButton.removeAttribute('disabled');
 		});
 	},
-	_getDisplayedValue: function(hasFocus, isInvalid, enteredValue, actualValue) {
-		return (hasFocus || isInvalid) ? enteredValue : actualValue;
+	_getDisplayedValue: function(hasFocus, pendingSaves, enteredValue, actualValue) {
```
xiangqiyan: enteredValue should also be shown if isInvalid is true

mpharoah-d2l: Should it? Do we actually want to keep invalid values after leaving the field?

xiangqiyan: it will be marked as invalid if the save request fails, even if for valid values, so I think we should keep the entered value

xiangqiyan:
There are other components with score inputs:
description-editor, overall-level-editor, criterion-editor, overall-level-editor
and the criteria group name and overall level name have the whitespace issues as well...

editor/d2l-rubric-autosaving-input.js
```diff
@@ -0,0 +1,110 @@
...
+		};
+	}
+
+	_onFocus(e) {
+		this._enteredValue = e.target.value;
+		this._isFocused = true;
+	}
+
+	_onBlur(e) {
+		if (this.editing || !this.pendingSaves && this.invalid) {
```
apalaniuk: This probably just has to flush the debouncer, no?

mpharoah-d2l: If you paste text, or drag it in, it won't generate an input event, so those changes would be lost. This probably should cancel the debouncer though before saving

PR #580, opened 2019-11-22
DE36885 -- Fix scores visually reverting their value after losing focus

mpharoah-d2l:


djwester:
I think this will need to get into 20.19.12 and master.

PR #582, opened 2019-12-02
Styling fix for rubric assessment and code refactor for rubric editor pending saves

xiangqiyan:


editor/d2l-rubric-siren-autosave-action-behavior.js
```diff
@@ -0,0 +1,30 @@
...
+
+	performAutosaveAction: function(action, fields, pendingPropPath) {
+		if (!action) {
+			return;
+		}
+
+		this.set(pendingPropPath, this.get(pendingPropPath) + 1);
+
+		return this.performSirenAction(action, fields).finally(function() {
+			this.set(pendingPropPath, this.get(pendingPropPath) - 1);
```
mpharoah-d2l: What does this do? It doesn't seem like anything cares about the value of this.get(pendingPropPath)

xiangqiyan: This will increment and decrement the pending saves property values corresponding to each autosaving input before and after the siren request so that I don't need to do it in each individual component. So I can just call updateInput() if the call was successful and it have the correct pending saves value.

PR #583, opened 2019-12-03
DE36955 - fix rubric assessment table cells in firebox

xiangqiyan:


d2l-rubric-criteria-group.js
```diff
@@ -345,6 +348,16 @@ Polymer({
 		'_updateFeedbackDisplay(_criteriaEntities, assessmentResult)'
 	],
 
+	_rowHeaderDomChange: function() {
+		// set styling to have the criteria-row-header-container be the same height as the table cell in firefox
+		if (navigator.userAgent.toLowerCase().indexOf('firefox') > -1) {
```
xiangqiyan: This issue happened after I made the change to have the add feedback button show at the bottom of the cell. I couldn't find a way to style the table that works for both chrome and firefox so i did this as a workaround.

mpharoah-d2l: What about Edge?

xiangqiyan: edge behaves the same as chrome

mpharoah-d2l: Doing a case sensitive check for `Firefox/` is safer (https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent#Browser_Name)

PR #588, opened 2019-12-09
Add shared info alert to the top of the rubric-editor for shared rubrics

wongvincent:


wongvincent:
> There's probably a reason I don't recall, but why is the `build` folder in git?

I think serge needed the files to be in `.json`, which is what the `lang` folder has the files as.
And the files outputted by serge were `.js`, and needed to be committed somewhere.

I think we could probably change the `lang` folder files to all be `.js`. Then, update the serge file to match `js` files and use the `parse_js` parser_plugin. Then serge should be able to handle it and we can update all the `build/lang` references to just `lang` and remove the `build/lang` folder.

Someone can try this when we've got more time to play around with the serge stuff.

PR #591, opened 2019-12-11
Fix incorrect accesses of input Shadow DOM

apalaniuk:
Some functions are currently expecting Polymer convenience methods to exist on external web components in order to access their Shadow DOM, which have since been upgraded to LitElement. For now, access it through the spec-compliant method - probably in the future these should expose a more standard interface themselves.

Currently causes autosave for editable scores to fail, as well as a few other UX issues.

### DIFF
--- a/d2l-rubric-editable-score.js
+++ b/d2l-rubric-editable-score.js
@@ -250,9 +250,9 @@ Polymer({
 	focus: function() {
 		var elem = this.$['text-area'];
 		elem.focus();
-		var inputElem = elem.$$('input');
+		var inputElem = elem.shadowRoot.querySelector('input');
 		if (inputElem && this._largeScreen) {
-			elem.$$('input').select();
+			inputElem.select();
 		}
 	},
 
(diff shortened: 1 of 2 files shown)


d2l-rubric-editable-score.js
```diff
@@ -288,7 +288,8 @@ Polymer({
 		if (event.relatedTarget && event.relatedTarget.id === 'clear-button') {
 			return;
 		}
-		var innerInput = event.target.$$('input');
+		var innerInput = event.target.shadowRoot.querySelector('input');
```
apalaniuk: Just eyeballed this, but checking their component, the validity check doesn't exist anymore. So, this is almost certainly not sufficient, just the initial error.

editor/d2l-rubric-criteria-groups-editor.js
```diff
@@ -183,7 +183,7 @@ Polymer({
 	_refocus: function() {
 		var allGroups = dom(this.root).querySelectorAll('d2l-rubric-criteria-group-editor');
 		var lastGroup = allGroups[allGroups.length - 1];
-		lastGroup.$$('d2l-input-text').$$('input').select();
+		lastGroup.$$('d2l-input-text').shadowRoot.querySelector('input').select();
```
apalaniuk: Seems like we should just refocus this?

PR #592, opened 2019-12-11
Add LOA Overlay

rgoessl:


d2l-rubric-criteria-group.js
```diff
@@ -388,6 +462,58 @@ Polymer({
...
+			this._levelsReversed = true;
+		} else {
+			this._levelsReversed = false;
+		}
+
+		this._loaLevels = loaLevelEntities;
+	},
+
+	_resolveRubricLevel: function(rubricLevelHref) {
+		if (!this._levels || !this._levels.length) {
```
mpharoah-d2l: This should take in `_levels` as a parameter to ensure that any changes to `this._levels` are resolved before `_resolveRubricLevel` is called

d2l-rubric-criteria-group.js
```diff
@@ -388,6 +462,58 @@ Polymer({
...
+
+			if (this._getSelfLink(rubricLevel) === rubricLevelHref) {
+				return rubricLevel;
+			}
+		}
+
+		return null;
+	},
+
+	_resolveLoaLevel: function(loaLevelHref) {
```
mpharoah-d2l: Same here. This should take _loaLevels as a paremter

rgoessl: Should I do this for all the functions that use properties?

editor/d2l-rubric-criteria-editor.js
```diff
@@ -292,6 +311,11 @@ Polymer({
 		}
 	},
 
+	_getRubricLevelsLink: function(entity) {
+		var link = entity && entity.getLinkByRel('https://rubrics.api.brightspace.com/rels/levels');
```
gfrancu: Do we not want this in HypermediaRels

rgoessl: Yes, I'll make the contants/lang terms later

d2l-rubric-criteria-group.js
mpharoah-d2l: Any function that's called from the HTML template (except event handlers) should take in all properties it uses as arguments instead of using `this` so that Polymer knows the dependency graph so it can order things properly when there are changes.

mpharoah-d2l: If a property never changes after the element is created, or if you are in an event handler, then it's safe to use `this`, but otherwise it should take it in as a property

editor/d2l-rubric-criterion-editor.js
```diff
@@ -500,6 +531,115 @@ Polymer({
...
+		const styles = [];
+
+		if (loaLevels && loaLevels.length) {
+			const rubricLevelHref = this._getRubricLevelLink(cell);
+			const rubricLevelEntity = this._resolveRubricLevel(rubricLevelHref);
+			const loaLink = rubricLevelLoaMapping[rubricLevelHref]
+				? rubricLevelLoaMapping[rubricLevelHref].loaLevel
+				: this._getLoaLevelLink(rubricLevelEntity);
+			const loaLevelEntity = this._resolveLoaLevel(loaLink);
+
```
mpharoah-d2l: You should pass loadLevels and rubricLevels into _resolveLoaLevel and _resolveRubricLevel respectively to ensure they are updated before _getCellStyle. I think in this case they will be anyways, but it's best to be explicit

rgoessl: I'm changing everything to be explicit

PR #594, opened 2019-12-13
DE36621 - Fix edge html editor paste doubling

xiangqiyan:


xiangqiyan:
There seems to be an issue with d2l-html-editor where using the paste plugin for tinymce causes the doubling issue on edge. I changed it to use the powerpaste plugin instead, but didn't set the powerPasteEnabled property for d2l-html-editor so that options such as paste_as_text would stay the same as before.

PR #599, opened 2019-12-19
Fix assessment overlay styling when resizing and scrolling and when levels are reversed

xiangqiyan:


xiangqiyan:
Since d2l-table doesn't support colspan and d2l-tspan doesn't work with the table's horizontal scrolling, I just made it so the width of the loa levels are updated when resizing and scrolling.

d2l-rubric-criteria-group.js
```diff
@@ -490,7 +500,7 @@ Polymer({
 		const loaLevelEntities = loaLevelEntity.getSubEntitiesByClass('level-of-achievement');
 
 		const lastRubric = this._resolveRubricLevel(this._levels, this._getRubricLevelLink(loaLevelEntities[loaLevelEntities.length - 1]));
-		const lastRubricIndex = this._getRubricLevelIndex(lastRubric);
+		const lastRubricIndex = this._getRubricLevelIndex(this._sortedLevels, lastRubric);
```
rgoessl: Good catch

d2l-rubric-criteria-group.js
```diff
@@ -930,22 +938,95 @@ Polymer({
...
+			const tableRect = table.getBoundingClientRect();
+
+			const firstRow = this.root.querySelector('.d2l-table-row-first');
+			const firstHeaderRect = firstRow.firstChild.getBoundingClientRect();
+			const startLabel = this.root.querySelectorAll('.loa-label')[0];
+			const startlabelTextDiv = startLabel.querySelector('div');
+			let startLabelWidth = 0;
+			if (firstHeaderRect.left < tableRect.left && firstHeaderRect.right > tableRect.left) {
+				// achievement label is clipped but still visible
+				startLabelWidth = firstHeaderRect.right - tableRect.left;
```
rgoessl: What is this case? Why is margin adjustment needed?

xiangqiyan: When you zoom in on the rubric, eventually a scroll bar shows up. You can scroll to make any of the table cells appear clipped. But because d2l-tspan doesn't have this scrolling, I'm settting the width of the loa levels to make it appear as if it's scrolling. Adding the margin lets the text inside look like it's scrolled properly.

d2l-rubric-criteria-group.js
```diff
@@ -930,22 +938,99 @@ Polymer({
 		return !this._isStaticView() && !feedbackInvalid;
 	},
 
-	_onLoaResize: function() {
-		const firstRow = this.root.querySelectorAll('.d2l-table-row-first')[0];
+	_setLoaCellsWidth: function() {
+		this.debounce('set-loa-cells-width', () => {
```
xiangqiyan: I wasn't sure if I should add this debounce or not. I was thinking there might be performance issues for large rubrics, but adding this makes the resizing not look as smooth.

rgoessl: I think it's a good idea

PR #600, opened 2019-12-19
Add slider behaviour for mobile criteron level selector

apalaniuk:


d2l-rubric.js
```diff
@@ -411,6 +411,14 @@ Polymer({
 		e.detail.opened
 			? this.setAttribute('compact-expanded', '')
 			: this.removeAttribute('compact-expanded');
+
+		this.dispatchEvent(new CustomEvent('d2l-rubric-compact-expanded-changed', {
```
apalaniuk: Unrelated, but reducing PR overhead: since consumer can't access inner-accordion state, and the current accordion event doesn't bubble/have desired info, bubble another event ourselves.

d2l-rubric-criterion-mobile.js
```diff
@@ -470,7 +469,8 @@ Polymer({
 	},
 
 	_isLevelSelected: function(levelIndex, selected) {
-		return levelIndex === selected;
+		return levelIndex === selected
+			|| (!(selected >= 0) && levelIndex === 0);
```
jstefaniuk-d2l:
```suggestion
			|| (selected < 0 && levelIndex === 0);
```

d2l-rubric-criterion-mobile.js
```diff
@@ -470,7 +469,7 @@ Polymer({
 	},
 
 	_isLevelSelected: function(levelIndex, selected) {
-		return levelIndex === selected;
+		return levelIndex === selected || (!(selected >= 0) && levelIndex === 0);
```
jstefaniuk-d2l:
```suggestion
		return levelIndex === selected || (selected < 0 && levelIndex === 0);
```

apalaniuk: Was concerned about some weird upstream behaviour setting to `undefined`, but you're right, this is doing a non-obvious implicit check ("is number") that I should make explicit. :+1:

PR #607, opened 2020-01-14
DE36452 -- Updated d2l-rubric component to use the new more performant assessments API

mpharoah-d2l:
Updated to use the new more performant API
Requires this LMS branch: https://git.dev.d2l/projects/CORE/repos/lms/pull-requests/6498/overview


d2l-rubric-criterion-mobile.js
```diff
@@ -404,18 +361,8 @@ Polymer({
...
-			return;
-		}
-		element.classList.remove('slide-from-right');
-		element.classList.remove('slide-from-left');
-		element.classList.add(animation);
-	},
-
 	_hideLeftChevron: function(selected) {
-		return selected === 0 || selected === -1;
+		return selected < 0;
```
rgoessl: This is different than the original condition, is that intended?

mpharoah-d2l: Yeah, it should be <=. Good catch

PR #616, opened 2020-01-22
Add focus states to criterion cells

max-hao-d2l:
[US112880](https://rally1.rallydev.com/#/detail/userstory/361150934368?fdp=true): Accessible focus states for grading rubric
This PR adds focus states to criterion cells when grading a rubric.
Event listeners were added for `focusin` and `focusout` for the invisible radio buttons within the rubric add or remove the `focused` classname. to the table cells.
We also need to refocus the radio buttons after a criterion cell is selected, because selecting a cell refreshes all of the classnames for the cells, which clears the `focused` classname.


max-hao-d2l:
The focusing of the cells follow the behaviour of the hidden radio buttons, which differs from the behaviour of the table cells themselves. When selecting a table cell via the arrow keys, the corresponding radio button gets focused before the API call, and the cell ends up rendering as "selected" after the API call, which ends up with a split second disparity between the "focused" and "selected" states for the cells. 
Don't know if it is acceptable for the requirements, any input would be appreciated.

d2l-rubric-criteria-group.js
```diff
@@ -838,13 +840,16 @@ Polymer({
...
+			this._refocusCriterionCell(event);
+		});
 		this._addingFeedback = -1;
 		this.editingScore = -1;
 	},
 
 	_handleKey: function(event) {
-		if (event.keyCode === 13 || event.keyCode === 32) { // enter or space key
+		var criterionCell = event.model.get('criterionCell');
+		if (event.keyCode === 13 || (this._isSelected(criterionCell, this.cellAssessmentMap) === true && event.keyCode === 32)) { // enter or space key
```
max-hao-d2l: There was an issue with the space key accidentally selecting the cell twice when the cell is not selected (it calls both `_handleKey` and `_handleTap`, which ultimately results in calling `_handleTap` twice). The issue is fixed here.

d2l-rubric-criteria-group.js
```diff
@@ -140,6 +140,9 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-criteria-group">
 				border-radius: 0.3rem;
 				border: 1px solid var(--d2l-color-celestine);
 			}
+			.criterion-cell.focused {
```
rgoessl:
```suggestion
			.criterion-cell:focus {
```

You can use the `:focus` CSS selector instead. Then the class and event handlers aren't necessary.

d2l-rubric-criteria-group.js
```diff
@@ -149,11 +152,16 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-criteria-group">
 				box-shadow: -2px 0 0 var(--d2l-color-celestine); /* left border */
 				border-width: 2px;
 			}
-			.criterion-cell.selected.has-bottom,
-			.criterion-cell.assessable.has-bottom:focus {
+			.criterion-cell.selected.focused {
```
rgoessl: `:focus`

d2l-rubric-criteria-group.js
```diff
@@ -149,11 +152,16 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-criteria-group">
...
+				box-shadow: -2px 0 0 var(--d2l-color-celestine), 0 0 0 4px inset rgba(0, 111, 191, 0.4);
+			}
+			.criterion-cell.selected.has-bottom {
 				box-shadow: -2px 0 0 var(--d2l-color-celestine), -2px 2px 0 var(--d2l-color-celestine), 0 2px 0 var(--d2l-color-celestine);
 				z-index: 1; /* Need bottom border to render over feedback cell border */
 			}
+			.criterion-cell.selected.has-bottom.focused {
```
rgoessl: `:focus`

max-hao-d2l:
Ended up using the `focus-within` CSS selector in place of toggling a `focused` classname. `focus-within` is compatible with all desktop browsers except for IE11 ([source](https://caniuse.com/#feat=css-focus-within)).

PR #617, opened 2020-01-22
US112903: Telemetry Performance 

lvyuan95-d2l:
Adding telemetry performance on:
- Assessing API performance(Tap cell)
- Rubric rendering
- Adding new criteraion/level API performance


lvyuan95-d2l:
Not sure why is CI failed, do you guys have any idea?

djwester:
CI says `this.perfMark is not a function`

telemetry-behavior.js
```diff
@@ -9,26 +9,19 @@ window.D2L.PolymerBehaviors.Rubric = window.D2L.PolymerBehaviors.Rubric || {};
...
-
-		eventBody.setAction('Open')
+	logViewRubricEvent: function({ id, isMobile = false }, telemetryData) {
+		const eventBody = new window.d2lTelemetryBrowserClient.EventBody();
+		eventBody.addCustom('isMobile', isMobile)
+			.setAction('View')
 			.setObject(encodeURIComponent(id), 'Rubric', id)
-			.addCustom('rubricMode', telemetryData.rubricMode || 'unkown')
-			.addCustom('originTool', telemetryData.originTool || 'unkown');
+			.addCustom('rubricMode', telemetryData.rubricMode || 'unknow')
```
mpharoah-d2l: Traded one misspelling for another

lvyuan95-d2l: oops

d2l-rubric-criteria-group.js
```diff
@@ -551,6 +551,16 @@ Polymer({
...
 	},
 
+	_onCriterionCellDomChanged: function() {
+		const self = this;
+		this.perfMark('rubricRenderEnd');
+		this.perfMark('rubricLoadEnd');
+		this.debounce('criterionCellDomChangedEvent', function() {
+			self.logRubricRenderedEvent('rubricRenderStart', 'rubricRenderEnd', self.telemetryData);
+			self.logRubricLoadedEvent('rubricLoadStart', 'rubricLoadEnd', self.telemetryData);
+		}, 500);
```
mpharoah-d2l: This looks like it will give inaccurate performance metrics since its' waiting 500ms. Probably want to set the timeout to 0 so that it sends on the first frame that no more changes are queued

lvyuan95-d2l: I'm not 100% sure but I think `window.performance.measure` calculates period between two marks, so when `window.performance.measure` is being called doesn't matter? As the end mark is before the `debounce`

mpharoah-d2l: Oh, I see. You still might want to use a shorter debounce period, otherwise if you do multiple actions within 0.5 seconds of each other, only the last one will get logged.

lvyuan95-d2l: 100ms sounds good?

PR #621, opened 2020-01-27
Fix out of order points error issue

max-hao-d2l:
[DE34471](https://rally1.rallydev.com/#/detail/defect/309752498680?fdp=true): Out of order points error message does not go away if you correct the point order by editing a cell that does not have the error associated with it

Levels can only be changed one at a time through the API. When there is an out-of-order points error, the offending level does not get saved. If the error is corrected by editing a level that is not the offending level, the error message remains and the offending level still does not get saved. We can fix this by attempting to save levels with errors whenever there is a successful save by another level.

### DIFF
--- a/editor/d2l-rubric-levels-editor.js
+++ b/editor/d2l-rubric-levels-editor.js
@@ -248,11 +248,13 @@ Polymer({
 		this.fire('iron-announce', { text: this.localize('addLevelAppend', 'name', this._getLastLevelName()) }, { bubbles: true });
 	},
 	_onSavePoints: function(event) {
-		// Upon successful save, we attempt to save other levels that have previously had errors saving (ex. out of bounds errors)
+		// Upon successful save, we attempt to save other levels that have previously had errors saving
 		var levels = Array.from(dom(this.root).querySelectorAll('d2l-rubric-level-editor'));
 		var saveIndex = levels.findIndex((level) => {
 			return level.entity.properties.id === event.id;
 		});
+
+		// To prevent false out of bounds errors, we call savePointsAfterError in an order starting from the levels adjacent to the saved level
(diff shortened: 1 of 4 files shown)


editor/d2l-rubric-levels-editor.js
```diff
@@ -246,5 +246,18 @@ Polymer({
...
 	_onAppendFocus: function() {
 		this.fire('iron-announce', { text: this.localize('addLevelAppend', 'name', this._getLastLevelName()) }, { bubbles: true });
+	},
+	_onSavePoints: function(event) {
+		// Upon successful save, we attempt to save other levels that have previously had errors saving (ex. out of bounds errors)
+		var levels = Array.from(dom(this.root).querySelectorAll('d2l-rubric-level-editor'));
+		var saveIndex = levels.findIndex((level) => {
+			return level.entity.properties.id === event.id;
+		});
+		for (let i = saveIndex + 1; i < levels.length; i++) {
```
rgoessl: Replace 2 loops with 1
```
for (let i = 0; i < levels.length; i++) {
	if (i !== saveIndex) {
		levels[i].savePointsAfterError();
	}
}
```

max-hao-d2l: We'll need to order the levels starting from the levels adjacent to the saved level. Going from left to right will not work in all cases.
ex. 10,8,6,4 -> 10,8,2,4 -> 10,6,2,4 -> 10,6,2,0
If going from left to right, the last change is the only successful change, and the rubric levels will be 10,8,6,0 on the server. Going from 10,8,6,0 to 10,6,6,0 will result in an error if going from left to right.

max-hao-d2l:
The same defect exists for overall level as well. I will get around to fixing that for this PR.

PR #622, opened 2020-01-28
Use attribute for perf logging

rgoessl:
Performance telemetry now respects new LD flag


lvyuan95-d2l:
Just wanted to make sure, do we want `performanceTelemetryFlag` only controls the performance log or all logs(including event, e.g. `logViewRubricEvent`). If it's the former I think we only need to put the `if not enabled: return` part into `_logAndDestroyPerformanceEvent`

PR #624, opened 2020-01-31
update translations

djwester:


djwester:
`loaSliderLabel` doesn't have translations because we changed the term after they were committed (we have manual translation for French though so we should be ok to release this. 

Also, ain't nobody got time for Danish or Arabic translations.  Following up on the serge_status channel in slack as to why we didn't get these.

djwester:
@rgoessl it looks like you new translation stuff used double quotes instead of single and spaces instead of tabs - linter is throwing a fit.

PR #627, opened 2020-02-03
Fix some UI tests

rgoessl:


test/components/d2l-rubric-criteria-groups-editor.js
```diff
@@ -37,7 +37,7 @@ suite('<d2l-rubric-criteria-groups-editor>', function() {
 				window.D2L.Siren.EntityStore.clear();
 			});
 
-			test.skip('adds criteria group', function(done) {
+			test.skip('adds criteria group', function(done) { // Hangs even though done is called, no errors
```
djwester: probably trying to check things before the nested shadow doms have fully rendered. The 'waiting' in these tests kinda sucks with nested components. I think other tests had to put in their own custom built waiting mechisms to make sure things are fully rendered before we do the checks

djwester:
We are skipping a lot of tests, but I guess this is ok for now - I think we need a spike at some point in the future to see what kind of testing framework or strategy we can take that will allow us to more easily test deeply nested components like we have in here.

PR #632, opened 2020-02-05
Load the telemetry endpoint from the html root element

mpharoah-d2l:
The telemetry endpoint is already added to every page by putting it in the dataset of the root HTML element:

https://search.d2l.dev/xref/lms/lp/framework/web/D2L.LP.Web/UI/Telemetry/TelemetryEndpointHtmlElementAttributeProvider.cs

https://search.d2l.dev/xref/lms/lp/framework/web/D2L.Web/ui/page/d2lPageControl.cs#953

Changed the component to load from this instead of relying on the value being passed through to the component directly.

### DIFF
--- a/d2l-rubric.js
+++ b/d2l-rubric.js
@@ -392,7 +392,7 @@ Polymer({
 			this._telemetryData = {
 				rubricMode: this.dataset.rubricMode,
 				originTool: this.dataset.originTool,
-				endpoint: this.dataset.telemetryEndpoint,
+				endpoint: window.document.documentElement.dataset.telemetryEndpoint,
 				performanceTelemetryEnabled: this.performanceTelemetryFlag
 			};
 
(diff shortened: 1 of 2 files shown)


lvyuan95-d2l:
I think the idea of having to pass the endpoint to the component is because we can have "feature flag" for telemetry, e.g. https://search.d2l.dev/xref/lms/platformTools/rubrics/D2L.PlatformTools.Rubrics/Views/AssessedRubric/AssessedRubric.Renderer.cs?r=8eeee10a#112

mpharoah-d2l:
We already have the `performance-telemetry-flag` feature flag though

lvyuan95-d2l:
Right, but that's for telemetry performance tracking right, there's also another flag to control telemetry overall, see https://learningoutcomes.slack.com/archives/CANSREZKJ/p1580235010028700

lvyuan95-d2l:
Could you please update the `demo/index.html` to reflect that change as well?

PR #637, opened 2020-02-15
Avoid legacy Polymer 1.x syntactic sugar

apalaniuk:
Discourage additional use of `fire`, `async`, `iron-announce` by removing them in favour of platform/@brightspace-ui/core APIs. (Not exhaustive, eg. didn't touch `debounce`).

- [x] ~Wait until tests are running again~ (they're running, but reporting is broken, I think)
- [x] Wait until current release is cut


d2l-rubric-editable-score.js
```diff
@@ -445,13 +445,24 @@ Polymer({
...
+				score: score,
+			};
+
+			if (outOf) {
+				detail.outOf = outOf.toString();
+			}
+
+			this.dispatchEvent(new CustomEvent('d2l-rubric-total-score-changed', {
+				detail: detail,
+				bubbles: true,
```
apalaniuk: `fire` implies both `bubbles` and `composed`, so setting by default in all uses

editor/d2l-rubric-alignment-editor.js
```diff
@@ -131,8 +130,12 @@ Polymer({
 				this.toggleBubble('_alignmentInvalid', false, 'alignment-bubble');
 				var fields = [{ 'name': 'alignment', 'value': this._debounceSelected }];
 				this.performSirenAction(action, fields).then(function() {
-					this.fire('d2l-rubric-alignment-saved');
-					this.fire('d2l-loa-overlay-refresh');
+					['d2l-rubric-alignment-saved', 'd2l-loa-overlay-refresh'].forEach(evt => {
```
apalaniuk: Probably `d2l-rubric-loa-overlay` could just be aware of "alignment saved" events and refresh rather than this being aware of it

editor/d2l-rubric-level-editor.js
```diff
@@ -303,12 +307,17 @@ Polymer({
...
 				this.performAutosaveAction(action, fields, '_pendingPointsSaves').then(function() {
-					this.fire('d2l-rubric-level-points-saved');
+					this.dispatchEvent(new CustomEvent('d2l-rubric-level-points-saved', {
+						bubbles: true,
+						composed: true,
+					}));
 					this._updatePoints(this.entity, false);
 
-					const savePointsEvent = new CustomEvent('save-points');
-					savePointsEvent.id = this.entity.properties.id;
```
apalaniuk: Odd pattern, and `id` should be under `detail` - updated event consumer to match

editor/d2l-rubric-loa-overlay.js
```diff
@@ -1,929 +1,938 @@
...
-		return this._getSelfLink(this._getFixedLoaLevel(loaLevels, reversed)) === this._getSelfLink(loaLevel);
-	}
-
-	_reloadEntity() {
-		D2L.Siren.EntityStore.fetch(this.href, this.token, true);
-	}
-}
-
-customElements.define('d2l-rubric-loa-overlay', RubricLoaOverlay);
+import { PolymerElement, html } from '@polymer/polymer';
```
apalaniuk: Looks like `CRLF` => `LF` (which is defined by the `.editorconfig` so, I guess was previously saved without a plugin that respected that)

editor/d2l-rubric-overall-level-editor.js
```diff
@@ -258,13 +262,20 @@ Polymer({
...
+					}));
 					this._updateRangeStart(this.entity, false);
 
 					var link = this.entity.getLinkByRel('self');
 					if (link) {
-						const saveRangeStartEvent = new CustomEvent('save-range-start');
-						saveRangeStartEvent.href = link.href;
+						const saveRangeStartEvent = new CustomEvent('save-range-start', {
+							detail: {
+								href: link.href
```
apalaniuk: [Updated event consumer to match](https://github.com/Brightspace/d2l-rubric/pull/637/files#diff-c89d6edfbcb9d736318e8461dd42ed5bR343)

d2l-rubric-criteria-group.js
```diff
@@ -565,7 +565,7 @@ Polymer({
 		this._addingFeedback = -1;
 		const criterionNum = +event.currentTarget.dataset.criterion;
 		var elem = dom(this.root).querySelector('#invisible-addFeedback' + this._getRowIndex(criterionNum));
-		this.async(function() {
+		setTimeout(function() {
```
ctwomey1: This is what is happening under the hood it looks like. https://github.com/Polymer/polymer/blob/master/lib/legacy/legacy-element-mixin.js#L823, probs better to do this unless we need the micro task jiggery

editor/d2l-rubric-editor.js
```diff
@@ -5,14 +5,14 @@ Creates and edits a rubric
 @demo demo/d2l-rubric-editor.html
 */
 import '@polymer/polymer/polymer-legacy.js';
```
ctwomey1: possibly not needed anymore?

d2l-rubric-criteria-group.js
apalaniuk: For nonzero delays, it will be a `setTimeout` effectively (https://github.com/Polymer/polymer/blob/master/lib/utils/async.js#L66). A 500ms delay is suspect here, but meh. Don't think there's anything where performing the task before processing any other events is critical/intended, but if so, we could add a similar helper to core and it can do the same thing (but using `queueMicrotask`)

apalaniuk:
I'll wait until after this release to merge.

PR #649, opened 2020-03-30
US114948 - Remove overall-score-flag

max-hao-d2l:
[US114948](https://rally1.rallydev.com/#/detail/userstory/378529980860?fdp=true): LD Flag Cleanup - rubrics-overall-score

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -30,7 +30,7 @@
 	<script type="module" src="../../@polymer/iron-demo-helpers/demo-pages-shared-styles.js"></script>
 	<script type="module" src="../../@polymer/iron-demo-helpers/demo-snippet.js"></script>
 	<script type="module" src="../../d2l-fetch-siren-entity-behavior/d2l-fetch-siren-entity-behavior.js"></script>
-	<script type="module" src="./fetch-siren-entity-whitelist.js"></script> 
+	<script type="module" src="./fetch-siren-entity-whitelist.js"></script>
 	<script type="module" src="../d2l-rubric.js"></script>
 	<script type="module" src="../d2l-rubric-title.js"></script>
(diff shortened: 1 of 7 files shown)


demo/index.html
```diff
@@ -120,14 +120,6 @@ <h3>Assessed Rubric with multiple criteria groups</h3>
 	<demo-snippet>
 		<d2l-rubric read-only="" href="data/rubrics/organizations/multiple-groups/203.json" token="foozleberries" assessment-href="data/rubrics/organizations/multiple-groups/203/assessment/10.json"></d2l-rubric>
 	</demo-snippet>
-	<h3>Rubric with Overall Score</h3>
```
mpharoah-d2l: We still want this test, right? But without the flag passed in?

max-hao-d2l: Fixed

test/components/d2l-rubric-text-only.js
```diff
@@ -24,13 +24,5 @@ suite('<d2l-rubric-text-only>', function() {
 			outOfContainer = dom(element.root).querySelector('.out-of-container');
 			expect(outOfContainer.attributes).to.have.ownProperty('hidden');
 		});
-
-		test('Overall Score section is not rendered if the rubric has no overall score', function(done) {
```
mpharoah-d2l: Why was this test removed?

rgoessl: The fixture wasn't passed the flag, so I can see it being useless now

rgoessl: Though you're right, there are many more conditions for displaying overall all, this should be re-added

max-hao-d2l: Holistic rubrics should be the only case where the overall level does not show up. I've added a test to check if the overall level does not show up for holistic rubrics.

PR #652, opened 2020-04-06
Add set width to hierarchical overlay

max-hao-d2l:
When collapsing/expanding a node in outcomes hierarchy, the width can change depending on the length of the text of the non-hidden nodes. Adding a set width of 800px (similar to the MVC dialog) fixes this. `simple-overlay` deals with the max-width of the compoent.

### DIFF
--- a/editor/d2l-rubric-criterion-editor.js
+++ b/editor/d2l-rubric-criterion-editor.js
@@ -93,6 +93,10 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-criterion-editor">
 				flex-direction: row;
 			}
 
+			#hierarchicaloverlay {
+				width: 50%;
+			}
+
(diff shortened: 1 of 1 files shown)


editor/d2l-rubric-criterion-editor.js
```diff
@@ -93,6 +93,10 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-criterion-editor">
 				flex-direction: row;
 			}
 
+			#hierarchicaloverlay {
+				width: 800px;
```
lvyuan95-d2l, line 97: Will this work for a smaller screen? e.g. 400px width window

max-hao-d2l: Yes, it does work. `simple-overlay` sets a max width and max-height to the component depending on the window size.

PR #657, opened 2020-04-23
Fix tests

apalaniuk:
Not so much fix as skip broken tests/remove broken dependencies so that other regressions aren't hidden.


test/components/d2l-rubric-holistic.js
```diff
@@ -3,15 +3,15 @@
...
+suite.skip('<d2l-rubric-holistic>', function() {
 
 	var element, sandbox;
 
 	suiteSetup(function(done) {
 		sandbox = sinon.sandbox.create();
 		element = fixture('holistic-rubric');
 		/* global getLoadedElement */
-		element = getLoadedElement(element, 'data/rubrics/organizations/holistic/226.json', done);
+		element = getLoadedElement(element, 'data/rubrics/organizations/holistic-percent/226.json', done);
```
apalaniuk, line 14: I think this is what was intended as `holistic` doesn't exist in test data, though it still doesn't pass, so skipping for now.

package.json
```diff
@@ -46,7 +46,6 @@
     "polymer-cli": "^1.9.7",
     "python-shell": "^0.5.0",
     "wct-browser-legacy": "^1.0.1",
-    "web-component-tester-istanbul": "^0.10.0",
```
apalaniuk, line 49: This is broken, as is `wct-istanbul` and `wct-istanbub`. I think the only path here is to migrate off of `web-component-tester` to eg. Karma.

PR #663, opened 2020-05-05
Added prop for hiding title dropdown

pureooze:
We need this for FACE and other pages where we don't have access to legacy and the user does not need to be able to use the Preview/Stats dropdown


editor/d2l-rubric-editor.js
```diff
@@ -228,17 +228,19 @@ const $_documentContainer = html `
...
-				<d2l-dropdown-menu>
-					<d2l-menu id="title-menu">
-						<d2l-menu-item text="[[localize('statistics')]]" disabled="[[!_hasStatisticsLink]]" on-d2l-menu-item-select="_handleSelectStatistics">
-						</d2l-menu-item>
-						<d2l-menu-item text="[[localize('preview')]]" disabled="[[!_hasPreviewLink]]" on-d2l-menu-item-select="_openPreviewDialog">
-						</d2l-menu-item>
-					</d2l-menu>
-				</d2l-dropdown-menu>
-			</d2l-dropdown>
+			<template is="dom-if" if="{{!titleDropdownHidden}}">
```
pureooze, line 231: Just added a conditional template around the dropdown so we can hide it in https://github.com/BrightspaceHypermediaComponents/activities/pull/864

pureooze:
Tried adding/removing of levels and other options locally, seemed to be fine 👍

PR #665, opened 2020-05-06
Removed some apt addons in travis.yml

pureooze:


.travis.yml
```diff
@@ -1,8 +1,6 @@
 language: node_js
 node_js:
   - "11"
-dist: trusty
-sudo: required
```
pureooze, line 5: According to https://docs.travis-ci.com/user/reference/overview/#linux sudo is enabled by default and the dist will default to Xenial

PR #668, opened 2020-05-09
US115595: Improve Telemetry for Rubrics Preview

lvyuan95-d2l:
- Refactored so `telemetryData` is available globally, no need to pass through components anymore.
- When logging `RubricLoadEvent`, check if both rubric/assessment entity has been received.
- Tested with 'delayed' assessment calls (by mocking API and adding `Thread.sleep()` in hypermedia API)


telemetry-behavior.js
```diff
@@ -11,17 +11,29 @@ window.D2L.PolymerBehaviors.Rubric = window.D2L.PolymerBehaviors.Rubric || {};
...
+	_loadedEntity: new Set(),
+	_data: {
+		sessionId: null
+	},
+
+	setTelemetryData: function(telemetryData) {
+		if (!telemetryData || !telemetryData.endpoint) return;
+
+		Object.assign(this.telemetryData, telemetryData);
+		this._data.sessionId = this.getUUID();
```
mpharoah-d2l, line 24: It looks like this will get a random UUID per component rather than per page. You'll want to move the sessionId and getUUID call to outside of the behaviour so its only called once

lvyuan95-d2l: I think `setTelemetryData` only gets called on `ready` in the outmost component (`d2l-rubruc.js`/`d2l-rubric-editor.js`) so I guess it should be fine?

mpharoah-d2l: Ah, it's because all instances share the same object reference since you use Object.assign instead of `=`.

PR #675, opened 2020-06-03
fix for DE34203 Add Criterion Button is Hard to Find due to Multiple Levels

rebeccazhou0727:
moved the footer containing add criteria outside the scroll wrapper


rebeccazhou0727:
We will skip the Web Component test for this ticket

PR #677, opened 2020-06-11
added loading message to screen reader when adding a new level

rebeccazhou0727:


djwester:
we actually don't need to update the files in /build until we have the translations back

editor/d2l-rubric-criteria-group-editor.js
```diff
@@ -326,7 +326,8 @@ Polymer({
 		if (action) {
 			const uuid = this.getUUID();
 			this.perfMark(`criterionAddedStart-${uuid}`);
-
+			// a new criterion is created before this sentence is finished
+			announce(this.localize('levelLoading', 'name', 'criterion'));
```
djwester, line 330: Don't we need to pass in the criterion name here? I think this is hardcoding the name to 'criterion' isn't it?

djwester: ahh, nevermind! I see what you are doing here. You are using the same 'level loading' lang term for a new criterion and for a new new criterion group.  looks good :)

PR #686, opened 2020-08-06
bypassing d2l-input-text validation

dlockhart:
Context: d2l-input-text is about to have validation enabled, which will suddenly start checking for min/max/required and a few other attributes. Since we're not sure how this will impact places that weren't expecting validation to suddenly start happening, we're disabling it.


dlockhart:
I'm guessing merging this isn't going to update & release.... seems to be manual.

djwester:
Yes, update and release of this repo is manual.

dlockhart:
@djwester Build is failing due to a syntax error in the Finnish lang file. I can fix that (a string just needs to be escaped), but Finnish isn't even a language we need to support and the file is in English anyway... so I can also just remove it. Thoughts?

djwester:
I don't exactly remember the details, but I think we need to have the file there even though it only has English in it, otherwise we end up with some errors. 

Also, once that is fixed I think there might still be a test failure (about holistic rubrics) that you don't need to worry about for now. This repo needs a bit of love soon :)

PR #688, opened 2020-09-08
DE40414 - Holistic Rubrics with No Score crash mobile view

JordanCorner:
[DE40414](https://rally1.rallydev.com/#/detail/defect/424405440612?fdp=true) - We encountered this error in FACE since we use the mobile view there. It was escalated into this defect.

This fix updates the isHolistic property in `d2l-criteria-group-mobile` to more correctly be "IsHolistic" instead of what it was before which would have been more appropriately named "IsPercentage".

If this has potential to cause harm since we are changing what this IsHolistic means please advise, or let me know of anything else that I haven't thought of as I haven't touched much rubrics before, though I did have the guidance of Kieran.


kieranderson (approved):
LGTM. Matching on rubric type rather than percentage for holistic.

djwester (approved):

PR #689, opened 2020-09-15
DE40503 - Preserve feedback focus state when cell is assessed

max-hao-d2l:
Evaluating a cell and adding feedback before the API call has returned causes the feedback box to lose state. There now exists a check upon the return of the API call to see if the user is currently adding feedback.
https://rally1.rallydev.com/#/detail/defect/432474477752?fdp=true


d2l-rubric-criteria-group.js
```diff
@@ -925,6 +928,11 @@ Polymer({
 		event.target.classList.remove('feedback-button-focused');
 	},
 
+	_handleFeedbackTextFocus: function(event) {
```
max-hao-d2l, line 931: Fixes the case where the feedback textbox is already open, shouldn't have an impact on functionality.

rgoessl (approved):

PR #700, opened 2021-02-08
Standards now have text to convey what criterion they belong to

MitchelPaulin:
Added an aria-label, screen reader will now describe what standards are aligned to what criterion, tested on NVDA


d2l-rubric-alignments-indicator.js
```diff
@@ -106,6 +109,13 @@ class RubricAlignmentsIndicator extends mixinBehaviors([
 		this.addEventListener('focus', () => this._hasFocus = true);
 	}
 
+	_getCriterionText(criterionName){
+		if(criterionName && criterionName.length > 0){
+			return "Standards aligned to " + criterionName;
```
rgoessl: Must use a lang term for this

rgoessl (changes requested):

d2l-rubric-alignments-indicator.js
MitchelPaulin: I asked @fionaffzhang  and she said we don't need to add lang terms for screen reader only text.

fionaffzhang: I thought we were discussing language translation? I could be wrong. 
Also the word "standard" is custom-specified so I suppose that needs to be a lang term

MitchelPaulin: Sounds good, I'll add a lang term

lang/en.json
```diff
@@ -106,6 +106,7 @@
 	"scoresVisibilityVisible": "Scores are visible to students",
 	"shareRubricSaveFailed": "Changing org unit availability failed",
 	"statistics": "Statistics",
+	"standardsAligned": "Standards aligned to {name}",
```
rgoessl, line 109: You need to use a select like this to get the configured outcomes term: https://github.com/Brightspace/d2l-outcomes-overall-achievement/blob/master/lang/en.js#L36

rgoessl (changes requested):

lang/en.json
MitchelPaulin: I'm not sure what you mean, what word here needs to be configured?

rgoessl: `Standards` is a configurable term. It could be Outcomes, Objectives, Expectations... based on what client selects

lang/en.json
```diff
@@ -106,7 +106,7 @@
 	"scoresVisibilityVisible": "Scores are visible to students",
 	"shareRubricSaveFailed": "Changing org unit availability failed",
 	"statistics": "Statistics",
-	"standardsAligned": "Standards aligned to {name}",
+	"standardsAligned": "{outcome, select, competencies {competencies} expectations {expectations} objectives {objectives} outcomes {outcomes} standards {standards} other {standards}} aligned to {name}",
```
rgoessl, line 109:
```suggestion
	"standardsAligned": "{outcome, select, competencies {Competencies} expectations {Expectations} objectives {Objectives} outcomes {Outcomes} standards {Standards} other {Standards}} aligned to {name}",
```

These should be capitalized (will need to re-run lang scripts)

d2l-rubric-alignments-indicator.js
```diff
@@ -113,7 +113,7 @@ class RubricAlignmentsIndicator extends mixinBehaviors([
 
 	_getCriterionText(criterionName) {
 		if (criterionName && criterionName.length > 0) {
-			return this.localize('standardsAligned', 'name', criterionName);
+			return this.localize('standardsAligned', 'outcome', this.outcomesTitleText, 'name', criterionName);
```
rgoessl, line 116: I'm not sure if `outcomesTitleText` will give the expected value here. It's value comes from C# here https://search.d2l.dev/xref/lms/platformTools/rubrics/D2L.PlatformTools.Rubrics/Views/Default/RubricRenderer.cs?r=d9c7c68a#67 which I think is meant to be human readable (possibly already translated) version of the outcome term; whereas the `select` statement in the lang file expects it to exactly match one of the enum values from here: https://search.d2l.dev/xref/lms/le/lo/D2L.LE.LO/Web/Desktop/Controls/OutcomesMasteryViewTableRenderer.cs?r=88059323#58

Unfortunately you might have to add a new attribute to the rubric component, similar to the example above, to get it to work.

rgoessl (changes requested):

rgoessl (approved):

devpow112 (approved):
Feel free to ignore the language checker failures they are all false positives.

PR #701, opened 2021-02-08
Switch to using BrightspaceUI/core button components

devpow112:
This is in service of US116199.


MitchelPaulin (dismissed):

devpow112:
Abandoning. Need to fix tests first.

PR #702, opened 2021-02-08
Clean-up tests

devpow112:
This fixes a few tests and removes some that were broken by https://github.com/Brightspace/d2l-rubric/pull/675. Since CI is failing on these every time anyway removing a broken test seems reasonable. This reduces our tests from 88 to 85 (we lost 3 tests around criteria addition that were broken anyways) .


djwester (approved):
Hopefully we can get to acutally cleaning these up into useful tests eventually, but certainly getting rid of always failing tests is ok :)

PR #707, opened 2021-02-09
Fixed some broken lang terms

MitchelPaulin:


build/lang/ko.js
```diff
@@ -102,7 +102,7 @@ D2L.PolymerBehaviors.Rubric.LocalizeBehavior.LangKoBehavior = {
 		'nameSaveFailed': '이름 저장 실패',
 		'newAssociationLabel': '다음에 새 연결 허용',
 		'numberAndPercentage': '{number}%',
-		'numberAndPoints': '{number} {number, plural, =1 {point} 기타 {points}}',
```
MitchelPaulin, line 105: 기타 might be the translation for "other", will need to get this re-translated if that is the case

rgoessl (approved):

PR #711, opened 2021-02-11
US116199 - Adjust accessibility for Add Feedback button

devpow112:
For an `Add Feedback` button under a criterion named `Criterion 1` this will read: `clickable add feedback for criterion Criterion 1 button`.

Tested with **Chrome**, **Edge** and **Firefox** using **NVDA**.

### DIFF
--- a/lang/fr.json
+++ b/lang/fr.json
@@ -5,7 +5,7 @@
     "addedOverallLevelAppend": "Un nouveau niveau de note globale a été ajouté après {name}",
     "addedOverallLevelPrepend": "Un nouveau niveau de note globale niveau a été ajouté avant {name}",
     "addFeedback": "Ajouter une rétroaction",
-    "addFeedbackFor": "Add feedback for criterion {criterionName}",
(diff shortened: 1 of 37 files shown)


lang/fr.json
```diff
@@ -5,6 +5,7 @@
     "addedOverallLevelAppend": "Un nouveau niveau de note globale a été ajouté après {name}",
     "addedOverallLevelPrepend": "Un nouveau niveau de note globale niveau a été ajouté avant {name}",
     "addFeedback": "Ajouter une rétroaction",
+    "addFeedbackFor": "Add feedback for criterion {criterionName}",
```
MitchelPaulin, line 8: Do we need to provide manual translations for the french version?

devpow112: We can translation for this should come in next week. I can google translate on for french now.

devpow112: Done in 64228f5

MitchelPaulin (approved):

PR #716, opened 2021-02-17
Rzhou/us115651 notify users of rubric structure

rebeccazhou0727:
After talking to Fiona, we decided that if there is only 1 criteria group, this information is only displayed in reading mode. If there is more than 1 criteria group in a rubric, we would then give the rubric structure information again when the keyboard is focused on the criteria group name input text field.

Rally Ticket: https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F385560857244

PR #722, opened 2021-02-18
Try doubling test timeout

devpow112:
Default test timeout is 90 seconds and rubrics tests are super slow.


devpow112:
@rebeccazhou0727 this will probably conflict with #716.

rebeccazhou0727 (approved):

PR #732, opened 2021-02-26
Version bump

mpharoah-d2l:


devpow112 (approved):

devpow112:
Actually @mpharoah-d2l  this isn't needed. You can just release 3.7.49 as it was bumped but not yet released.

PR #747, opened 2021-03-05
Fix dialog width issues

devpow112:
The dialog width was not being set correctly. This resulted in a thinner dialog then desired and the background rubric behind the dialog getting shifted every time the dialog was shown.

**Before**
before

**After**
after

### DIFF
--- a/editor/d2l-rubric-criterion-editor.js
+++ b/editor/d2l-rubric-criterion-editor.js
@@ -106,10 +106,6 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-criterion-editor">
 				flex-direction: row;
 			}
 
-			#hierarchicaloverlay {
-				width: 800px;
-			}
-
 			#outcometag:not([hidden]){
 				border-left: 1px solid var(--d2l-color-galena);
 				margin-right: 50px;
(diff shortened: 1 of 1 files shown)


rebeccazhou0727 (approved):
Interesting.. Making it inline style fixes the problem 😮

devpow112:
> Interesting.. Making it inline style fixes the problem 😮

It's actually a property on the component. Inline styling using old HTML stuff like `width` doesn't work on components. They added the new component a while back and I'm guessing we just never noticed to switch it. Width used to not work at all for dialogs.

editor/d2l-rubric-criterion-editor.js
```diff
@@ -275,7 +271,7 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-criterion-editor">
 			</d2l-select-outcomes>
 		</d2l-dialog>
 
-		<d2l-dialog title-text="[[browseOutcomesText]]" id="hierarchicaloverlay">
+		<d2l-dialog title-text="[[browseOutcomesText]]" width="800" id="hierarchicaloverlay">
```
rgoessl, line 274: Is the `id` necessary still?

devpow112: I'll double check but I think it's used to determine which dialog to mark as `open`.

mpharoah-d2l (approved):

editor/d2l-rubric-criterion-editor.js
devpow112: Yeah it's used to determine what to mark as `open`
https://github.com/Brightspace/d2l-rubric/blob/master/editor/d2l-rubric-criterion-editor.js#L976

PR #755, opened 2021-03-15
Bump version

devpow112:
In preparation for release. To release telemetry changes.


djwester (approved):

djwester:
> @djwester do we want the telemetry changes in cert as well?

Yeah, I think that would be good. Thanks for following up on this!

PR #756, opened 2021-03-22
removing Polymer style mixins

dlockhart:
Before we can migrate `d2l-table` from Polymer to Lit, we need to remove usages to its Polymer style mixins -- which aren't supported in Lit.

### DIFF
--- a/d2l-rubric-overall-score.js
+++ b/d2l-rubric-overall-score.js
@@ -123,20 +123,6 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-overall-score">
 			}
 
 			d2l-scroll-wrapper {
-
-				--d2l-scroll-wrapper-h-scroll: {
-					border-left: 1px dashed var(--d2l-color-mica);
-					border-right: 1px dashed var(--d2l-color-mica);
-				};
-
-				--d2l-scroll-wrapper-left: {
-					border-left: none;
-				};
-
-				--d2l-scroll-wrapper-right: {
-					border-right: none;
-				};
-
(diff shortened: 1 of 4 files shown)


d2l-rubric-overall-score.js
```diff
@@ -123,20 +123,6 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-overall-score">
...
-					border-right: 1px dashed var(--d2l-color-mica);
-				};
-
-				--d2l-scroll-wrapper-left: {
-					border-left: none;
-				};
-
-				--d2l-scroll-wrapper-right: {
-					border-right: none;
-				};
```
dlockhart, line 138: Table has been updated so that when the overflow buttons are visible, it automatically gets a dashed mica border -- so this isn't needed anymore.

editor/d2l-rubric-editor-cell-styles.js
```diff
@@ -13,32 +13,17 @@ $_documentContainer.innerHTML = /*html*/`<dom-module id="d2l-rubric-editor-cell-
...
-			};
-
-			--d2l-scroll-wrapper-right: {
-				overflow-y: hidden;
-				border-right: none;
-				margin-right: unset;
-			};
-
+			--d2l-scroll-wrapper-overflow-border-color: var(--d2l-color-galena);
+		}
```
dlockhart, line 17: The below CSS accomplishes the same thing the mixins above were trying to do -- apply some left/right space when the overflow area is scrolled in different positions so that the edges line up with the sticky row at the bottom.

devpow112 (dismissed):

devpow112 (approved):

PR #757, opened 2021-03-22
removing redundant attributes

dlockhart:
This is just setting these to the default values. These attributes weren't ever used, so the plan is to remove them in a future change to `d2l-table` prior to the Lit migration.

### DIFF
--- a/editor/d2l-rubric-criteria-group-editor.js
+++ b/editor/d2l-rubric-criteria-group-editor.js
@@ -103,7 +103,7 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-criteria-group-edito
 			</div>
 		</template>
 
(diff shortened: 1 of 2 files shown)


devpow112:
Test failure was for a flaky test 😞. Re-ran tests.

devpow112:
@dlockhart I'm guessing we can just lump this together with the version bump from the previous PR?

dlockhart:
> I'm guessing we can just lump this together with the version bump from the previous PR?

Yeah for sure, although this code isn't actually doing anything so I didn't include a version bump here. But yeah, either way.

devpow112 (approved):

devpow112:
> > I'm guessing we can just lump this together with the version bump from the previous PR?
> 
> Yeah for sure, although this code isn't actually doing anything so I didn't include a version bump here. But yeah, either way.

Hopefully automate this someday using semantic-release, but for now... manual it is :P

PR #760, opened 2021-03-25
removing unnecessary overrides

dlockhart:
We're [changing scroll-wrapper](https://github.com/BrightspaceUI/table/pull/234/files) so that the action button colours are no longer configurable -- they just use nice defaults that should work for everyone.

So setting these variables is no longer necessary.

### DIFF
--- a/d2l-rubric-overall-score.js
+++ b/d2l-rubric-overall-score.js
@@ -122,11 +122,6 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-overall-score">
 				flex-direction: row;
 			}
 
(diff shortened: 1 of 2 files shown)


devpow112 (approved):

devpow112:
I have some translations coming in for this so I'll take care of releasing it along with those changes

dlockhart:
Thanks! It technically should wait for [this PR in table](https://github.com/BrightspaceUI/table/pull/234) to merge and release, which should happen within the hour anyway.

dlockhart:
Ok good to go now!

PR #771, opened 2021-04-30
replacing uses of table variables with inline styles

dlockhart:
The Polymer version of table currently exposes all of its CSS variables globally on the HTML element:
https://github.com/BrightspaceUI/table/blob/master/d2l-table-shared-styles.js

This isn't something we normally do or like doing, since typically the variables are meant for the component itself. I'd also like to remove and simplify a lot of these variables while porting this to Lit.

So this change replaces usages of the variables with the underlying static values used by table.


editor/d2l-rubric-loa-overlay.js
```diff
@@ -98,7 +98,7 @@ class RubricLoaOverlay extends mixinBehaviors([
                     display: flex;
                     flex-direction: row;
                     justify-content: flex-start;
-                    padding: var(--d2l-table-cell_-_padding);
+                    padding: 0.5rem 1rem;
```
dlockhart, line 101: This one is interesting. It was relying on how Polymer style mixins work internally -- so the `d2l-table-cell` mixin set a `padding` value, which creates a `--d2l-table-cell_-_padding`. This would have broken a couple months ago when I [removed all the style mixins](https://github.com/BrightspaceUI/table/pull/226/files#diff-0df310981dd8ac5c4b98ac2cad6b36bbe2921c9898ce7ed2968a067a7514e6aeL24).

Callum-Mitchell (approved):

editor/d2l-rubric-criteria-group-editor.js
```diff
@@ -81,9 +81,9 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-criteria-group-edito
 				padding: 0.45rem;
 				flex: 1 1 auto;
 				border: 1px solid var(--d2l-color-galena);
-				border-bottom-left-radius: var(--d2l-table-border-radius);
```
svanherk, line 84: Minor: Seems like we don't need the `import 'd2l-table/d2l-table-shared-styles.js';` import in this file anymore?

editor/d2l-rubric-criteria-editor.js
```diff
@@ -45,9 +45,9 @@ $_documentContainer.innerHTML = /*html*/`<dom-module id="d2l-rubric-criteria-edi
 				padding: 0.45rem;
 				flex: 1 1 auto;
 				border: 1px solid var(--d2l-color-galena);
-				border-bottom-left-radius: var(--d2l-table-border-radius);
```
svanherk, line 48: Likewise for here and a few other files... I'm thinking `import 'd2l-table/d2l-table-shared-styles.js';` won't be needed anywhere now, because `d2l-table` includes `d2l-table-style.js`?

svanherk (approved):

editor/d2l-rubric-criteria-editor.js
dlockhart: Yeah good call -- I've removed all the `d2l-table-shared-styles.js` imports. It imported our colours, so in files that also used colours but didn't import them explicitly I've added that.

editor/d2l-rubric-criterion-editor.js
```diff
@@ -14,7 +14,6 @@ import '@brightspace-ui/core/components/button/button-subtle.js';
 import '@brightspace-ui/core/components/dropdown/dropdown-more.js';
 import '@brightspace-ui/core/components/dropdown/dropdown-menu.js';
 import 'd2l-tooltip/d2l-tooltip.js';
-import 'd2l-table/d2l-tspan.js';
```
dlockhart, line 17: Wasn't used in this file.

svanherk (approved):

PR #780, opened 2021-05-20
mobile rubric style updates

semibran:
Rally: [US126933](https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F521003693996)

Follow-up from #779 

Part of the inline grading rubric revamp - removes default selection tile, improving clarity with regards to which level is selected

PR #781, opened 2021-05-25
mobile rubric keyboard navigation

semibran:
Rally: [US126939](https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F521015653580)

Follow-up from #780, part of the inline grading rubric revamp

Streamlines rubric interface for keyboard/screen reader users by reducing the number of focusable elements. Offloads a good chunk of interactivity to a hidden range-type input in order to move arrow key handling, tabulation, and ARIA props as close to the metal as possible.


d2l-rubric-levels-mobile.js
```diff
@@ -400,34 +448,61 @@ Polymer({
 		}
 	},
 
+	_getLevelText: function(index) {
+		const item = this.criterionCells[index];
+		const levelName = item && item.properties && item.properties.levelName;
+		const levelDescription = item && item.getSubEntityByClass(this.HypermediaClasses.text.description).properties.text;
```
Callum-Mitchell, line 454: This could throw an exception if the appropriate subentity can't be found -- can we reasonably guarantee its existence?

d2l-rubric-levels-mobile.js
```diff
@@ -341,14 +394,21 @@ Polymer({
 		this.levelEntities = entity.getSubEntitiesByClass(this.HypermediaClasses.rubrics.level);
 	},
 
-	_getLevelClassName: function(index, selected, hovered, criterionCells, cellAssessmentMap) {
+	_getLevelCount: function(levelEntities) {
+		return levelEntities.length - 1;
```
Callum-Mitchell, line 398: Might be worth adding a comment explaining why we're subtracting 1 here (I'm assuming it's to exclude an overall level entity or similar from the total).

Callum-Mitchell (commented):
Couple of nitpicks, otherwise LGTM!

d2l-rubric-levels-mobile.js
semibran: I'm not completely sold on throwing an exception, but some more robust ref checking is warranted for sure

d2l-rubric-levels-mobile.js
semibran: It's actually to handle an off-by-one indexing gimmick - like a scale with 4 levels needs to be indexed from 0 to 3. I guess that could be a bit clearer.

Callum-Mitchell (dismissed):

Callum-Mitchell (approved):

PR #782, opened 2021-05-26
mobile rubric selection tile animation

semibran:
Rally: [US126941](https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F521016659012)

Follow-up from #781, part of the inline grading rubric revamp

Animates the movement of the level slider's thumb.


d2l-rubric-levels-mobile.js
```diff
@@ -387,6 +374,19 @@ Polymer({
...
+	},
+
+	_getThumbClassName: function(selected, focused) {
+		return 'level-thumb'
+			+ (focused >= 0 && focused === selected ? ' focused' : '');
+	},
+
+	_getThumbStyles: function(selected) {
+		return `transform: translateX(calc(${selected * 100}% - ${selected * 2}px))`;
+	},
```
semibran, line 388: still thinking about a way to clean up the calculations here. The thumb uses the width of the first level to determine x-position, which needs to be offset by 2 pixels because of the level border

Callum-Mitchell (approved):

PR #784, opened 2021-05-27
mobile rubric arrow button interactions

semibran:
Rally: [US126937](https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F521010078100)

Follow-up from #782, part of the inline grading rubric revamp

Fixes selected criterion levels in place for preview/graded mode and improves overall screen reader usability.

PR #786, opened 2021-05-31
mobile rubric arrow button interactions (v2)

semibran:
Rally: [US126937](https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F521010078100)

Follow-up from #784 (merge error)
- Uses `<d2l-offscreen>` instead of `.sr-only` class
- Fixes tabulation defects by simplifying selection logic


MitchelPaulin (approved):
Make sure the lang terms get update for all languages (especially french) but other than that LGTM

PR #787, opened 2021-05-31
switching to use core and alphabetizing imports

dlockhart:
No functional changes here, just doing some cleanup ahead of switching rubrics to use native tables instead of the `d2l-table`/`d2l-th`/etc. elements.

This does two things:
1. For web components which have been moved to core, use those instead of the old individual modules.
2. Alphabetize all the imports and add a linting rule to keep those alphabetized. Otherwise it's impossible to check if something you need is already included.


.eslintrc.yaml
```diff
@@ -4,6 +4,7 @@ rules:
   eqeqeq: [2, "always", {"null": "ignore"}]
   no-useless-escape: 0
   object-shorthand: [2, "never"]
+  sort-imports: [2, {"ignoreCase": true}]
```
dlockhart, line 7: When this gets migrated to Lit, the eslint config it uses includes this.

d2l-rubric-overall-score.js
```diff
@@ -172,7 +172,7 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-overall-score">
 					></d2l-rubric-competencies-icon>
 				</template>
 			</h3>
-			<d2l-scroll-wrapper show-actions="" role="group" aria-labelledby="overall-grouping-label">
```
dlockhart, line 175: The `show-actions` attribute is no longer needed -- actions are shown by default.

demo/rubric-editor.html
```diff
@@ -18,18 +18,6 @@
...
-	<script type="module" src="/node_modules/@polymer/promise-polyfill/promise-polyfill.js"></script>
-	<script src="/node_modules/whatwg-fetch/fetch.js"></script>
-
-	<script type="module" src="/node_modules/@polymer/iron-demo-helpers/demo-pages-shared-styles.js"></script>
-	<script type="module" src="/node_modules/@polymer/iron-demo-helpers/demo-snippet.js"></script>
-	<script type="module" src="/node_modules/d2l-typography/d2l-typography.js"></script>
-	<script type="module" src="/node_modules/d2l-fetch-siren-entity-behavior/d2l-fetch-siren-entity-behavior.js"></script>
-	<script type="module" src="./fetch-siren-entity-whitelist.js"></script>
-	<script type="module" src="../editor/d2l-rubric-editor.js"></script>
-	<script type="module" src="../d2l-rubric-title.js"></script>
```
dlockhart, line 31: These imports are duplicated and were causing a lot of duplicate component registration errors.

editor/d2l-rubric-criteria-group-editor.js
```diff
@@ -103,7 +104,7 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-criteria-group-edito
 			</div>
 		</template>
 
-		<d2l-scroll-wrapper id="scroll-wrapper" show-actions check-scroll-delta-value="1">
```
dlockhart, line 106: `check-scroll-delta-value` is gone.

editor/d2l-rubric-dropdown-menu-behavior.js
```diff
@@ -1,4 +1,6 @@
+import '@brightspace-ui/core/components/menu/menu-item-radio.js';
```
dlockhart, line 1: This mixin creates `menu-item-radio` elements, so the import for those should be here.

editor/d2l-rubric-editor-header.js
```diff
@@ -73,16 +73,6 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-editor-header">
...
-
-			d2l-alert {
-				margin: 0;
-				margin-left: var(--d2l-rubric-editor-start-gutter-width);
-				margin-right: var(--d2l-rubric-editor-end-gutter-width);
-			}
-			:dir(rtl) d2l-alert {
-				margin-right: var(--d2l-rubric-editor-start-gutter-width);
-				margin-left: var(--d2l-rubric-editor-end-gutter-width);
-			}
```
dlockhart, line 85: There are no alerts used in this component.

package.json
```diff
@@ -39,28 +39,19 @@
...
-    "d2l-alert": "github:BrightspaceUI/alert#semver:^4",
-    "d2l-colors": "github:BrightspaceUI/colors#semver:^4",
     "d2l-dnd-sortable": "github:Brightspace/dnd-sortable#semver:^3",
-    "d2l-dropdown": "github:BrightspaceUI/dropdown#semver:^7",
     "d2l-fetch": "github:Brightspace/d2l-fetch#semver:^2",
     "d2l-fetch-siren-entity-behavior": "github:Brightspace/d2l-fetch-siren-entity-behavior#semver:^6",
     "d2l-html-editor": "github:Brightspace/d2l-html-editor#semver:^2",
     "d2l-hypermedia-constants": "^6",
-    "d2l-icons": "github:BrightspaceUI/icons#semver:^6",
     "d2l-inputs": "github:BrightspaceUI/inputs#semver:^2",
```
dlockhart, line 47: Inputs, typography and table are still needed for "Polymer" reasons.

devpow112:
Looks like the test failure is related to something with `d2l-icon`
```
Failed to execute 'define' on 'CustomElementRegistry': the name "d2l-icon" has already been used with this registry
    <unknown> at /components/d2l-table/node_modules/ at brightspace-ui/core/components/icons/icon.js:72:0
```

dlockhart:
> Looks like the test failure is related to something with d2l-icon

Yeah started looking into this last night but ran out of steam. It's going to be related to importing things from `node_modules/@brightspace-ui/core` AND using the `@brightspace-ui/core` way and then getting duplicate registrations of the same web components. I'll sort it out this morning.

.eslintrc.yaml
svanherk: Maybe something we should just add to the Polymer config now. I've done this in a few Polymer repos too, it's an easy small step towards migration when there isn't time to actually migrate yet.

dlockhart: Yeah I can do that, although it'll require repos to upgrade to that new version which is likely going to be quite the undertaking. Also I noticed the version of eslint that this repo is using doesn't actually catch non-alphabetical imports within the groupings, so not a huge gain really.

dlockhart:
Ok looks like the tests are all sorted now!

editor/d2l-rubric-structure-editor.js
```diff
@@ -5,27 +5,27 @@ Creates and edits a rubric structue
...
-import 'd2l-alert/d2l-alert.js';
-import '../d2l-rubric-loading.js';
-import '../localize-behavior.js';
-import '../d2l-rubric-entity-behavior.js';
 import 'd2l-polymer-siren-behaviors/store/siren-action-behavior.js';
-import './d2l-rubric-editor-header.js';
+import 'd2l-typography/d2l-typography-shared-styles.js';
 import './d2l-rubric-criteria-groups-editor.js';
-import './d2l-rubric-overall-levels-editor.js';
-import 'd2l-menu/d2l-menu-item-radio.js';
```
svanherk, line 21: Assuming this being removed from here matches up with https://github.com/Brightspace/d2l-rubric/pull/787#discussion_r642682416 👍

dlockhart: Yeah exactly, moving it to the mixin where it's actually used.

svanherk (approved):

devpow112 (approved):

PR #789, opened 2021-06-02
mobile rubric assessment saving

semibran:
Rally: [DE43477](https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fdefect%2F603621606540)

Implements a 1 second debounce timeout at the criterion level to avoid unnecessary post requests. Also blocks regressive cache priming for save responses from taking effect on the frontend.

### DIFF
--- a/d2l-rubric-criterion-mobile.js
+++ b/d2l-rubric-criterion-mobile.js
@@ -358,18 +358,32 @@ Polymer({
 		}
 	},
 
+	_isCachePrimed: function() {
+		return !!this._levelEntities;
+	},
+
+	_isCachePriming: function(assessmentCriterionEntity) {
+		return !!assessmentCriterionEntity.rel;
+	},
+
+	_isCriterionSaving: function(assessmentCriterionEntity) {
+		return this._save
+			&& this._save.entity
+			&& this._save.entity.getLink().href !== assessmentCriterionEntity.getAction().href;
+	},
+
+	_canUpdateAssessment: function(assessmentCriterionEntity) {
+		return !this._isCachePrimed() ||
+			!this._isCachePriming(assessmentCriterionEntity)
+			&& this._isCriterionSaving(assessmentCriterionEntity)
+			&& !--this._save.ops;
+	},
+
 	_selectAssessedLevel: function(cells, cellAssessmentMap, assessmentCriterionEntity) {
 		if (!cells
-		|| !cellAssessmentMap
-		|| !Object.keys(cellAssessmentMap).length
+		|| !cellAssessmentMap || !Object.keys(cellAssessmentMap).length
 		|| !assessmentCriterionEntity
-		|| this._levelEntities && (
-			assessmentCriterionEntity.rel
-			|| !this._save
-			|| !this._save.entity
-			|| this._save.entity.getLink().href !== assessmentCriterionEntity.getAction().href
-			|| --this._save.ops
-		)) {
+		|| !this._canUpdateAssessment(assessmentCriterionEntity)) {
 			return;
 		}
 		if (this._save && !this._save.ops) {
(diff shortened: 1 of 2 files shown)


d2l-rubric-criterion-mobile.js
```diff
@@ -273,6 +273,8 @@ Polymer({
 			value: null
 		},
 
+		_save: Object,
```
semibran, line 276: expand into separate fields?

d2l-rubric-criterion-mobile.js
```diff
@@ -328,10 +330,51 @@ Polymer({
...
+		|| !cellAssessmentMap
+		|| !Object.keys(cellAssessmentMap).length
+		|| !assessmentCriterionEntity
+		|| this._levelEntities && (
+			assessmentCriterionEntity.rel
+			|| !this._save
+			|| !this._save.entity
+			|| this._save.entity.getLink().href !== assessmentCriterionEntity.getAction().href
+			|| --this._save.ops
+		)) {
```
semibran, line 372: For every criterion, this entire handler needs to run:
- Once on page load to adjust levels to match partially assessed criteria
- Once after every save action response is received

The original code calls this handler several times more than necessary. I didn't dig too deeply to find the root cause of why this is occurring, but I did note some behavioral patterns that get filtered out in the above conditional.

**On page load:**
- `_levelEntities` is null
- Handler fires multiple times with undefined/null `cellAssessmentMap` and `assessmentCriterionEntity`
- The actual cache prime yields an actual `Entity` instance through `assessmentCriterionEntity`

**On save action response:**
- `_levelEntities` is populated
- Handler fires twice for each criterion in succession, plus the actual response for the criterion that was selected
- The first "round" of criteria updates yields an empty `cellAssessmentMap` (`{}`)
- The second round includes a `rel` property under `assessmentCriterionEntity` to indicate it's a `cache-primer`

It looks like there's a number of ways to map selection entities to assessment entities to ensure only relevant criteria are changed, but matching hrefs appears to be the simplest route.

semibran, line 372: For every criterion, this entire handler needs to run:
- Once on page load to adjust levels to match partially assessed criteria
- Once after every save action response is received

The original code calls this handler several times more than necessary. I didn't dig too deeply to find the root cause of why this is occurring, but I did note some behavioral patterns that get filtered out in the above conditional.

**On page load:**
- `_levelEntities` is null
- Handler fires multiple times with undefined/null `cellAssessmentMap` and `assessmentCriterionEntity`
- The actual cache prime yields an actual `Entity` instance through `assessmentCriterionEntity`

**On save action response:**
- `_levelEntities` is populated
- Handler fires twice for each criterion in succession, plus the actual response for the criterion that was selected
- The first "round" of criteria updates yields an empty `cellAssessmentMap` (`{}`)
- The second round includes a `rel` property under `assessmentCriterionEntity` to indicate it's a `cache-primer`

There's probably a number of ways to map selection entities to assessment entities to ensure only the relevant criteria are changed, but it looks like matching hrefs is the simplest way to do it.

d2l-rubric-criterion-mobile.js
```diff
@@ -328,10 +330,51 @@ Polymer({
...
+			this._save = {
+				timeout: timeout,
+				entity: null,
+				ops: 0
+			};
+		}
+	},
+
+	_selectAssessedLevel: function(cells, cellAssessmentMap, assessmentCriterionEntity) {
+		if (!cells
```
MitchelPaulin, line 362: This is a very complicated if statement, I recommend putting this in its own function and naming it `is{ThingThisShouldCheck}` to make the intent clear. To me at least its very hard to decipher what this check is doing

MitchelPaulin (approved):

PR #790, opened 2021-06-02
initial draft work to switch to native tables

dlockhart:
**Note:** this is an incomplete draft PR and should not be merged as-is.

While attempting to convert this repo over from using the `d2l-table`/`d2l-tr`/`d2l-td`/etc. Polymer elements over to the Lit `<d2l-table-wrapper>` + native `<table>` approach, I ran into several roadblocks that involve pretty significant changes to how rubrics works (for the better though).

Therefore, I don't feel comfortable making these changes, nor do I have enough time to do them properly -- so I'm leaving this draft PR up as a first pass and will add some inline comments about the roadblocks should someone want to tackle them one day (maybe during a future Lit migration!).

### DIFF
--- a/d2l-rubric-criteria-group.js
+++ b/d2l-rubric-criteria-group.js
@@ -3,13 +3,7 @@ import '@brightspace-ui/core/components/colors/colors.js';
 import '@brightspace-ui/core/components/offscreen/offscreen.js';
 import '@polymer/polymer/polymer-legacy.js';
 import 'd2l-hypermedia-constants/d2l-hypermedia-constants.js';
-import 'd2l-table/d2l-table.js';
-import 'd2l-table/d2l-td.js';
-import 'd2l-table/d2l-tr.js';
-import 'd2l-table/d2l-thead.js';
-import 'd2l-table/d2l-tbody.js';
-import 'd2l-table/d2l-th.js';
-import 'd2l-table/d2l-tspan.js';
+import 'd2l-table/d2l-table-wrapper.js';
 import 'd2l-typography/d2l-typography-shared-styles.js';
 import 'fastdom/fastdom.js';
 import 's-html/s-html.js';
@@ -39,22 +33,19 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-criteria-group">
 				position: relative;
 				--column-count: 0;
 			}
-			d2l-table[type="default"] d2l-td.out-of {
+			d2l-table-wrapper[type="default"] td.out-of {
 				text-align: left;
 				vertical-align: top;
 				min-width: 0;
 				pointer-events: none;
 			}
-			d2l-table[type="default"] d2l-td.out-of.assessable {
+			d2l-table-wrapper[type="default"] td.out-of.assessable {
 				pointer-events: auto;
 				max-width: 75px;
 			}
-			d2l-td.out-of.assessable:hover {
+			td.out-of.assessable:hover {
 				color: var(--d2l-color-celestine);
 				cursor: pointer;
-			}
-			d2l-th  {
-				text-align: center;
 			}
 			d2l-rubric-editable-score {
 				min-width: 70px;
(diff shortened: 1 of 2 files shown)


d2l-rubric-criteria-group.js
```diff
@@ -266,149 +254,150 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-criteria-group">
...
 								</div>
+							</th>
+						</template>
+						<template is="dom-if" if="[[_hasOutOf(entity)]]">
+							<th class="out-of">
+								<div class="level-name">[[localize('criterionScore')]]</div>
+							</th>
+						</template>
+					</tr>
+					<!--
```
dlockhart, line 279: This is the first problem, which I've commented out here.`<d2l-tspan>` is being used to "span" the whole row, and then JavaScript runs to adjust the size of the fake "cells" inside.

Screen Shot 2021-06-02 at 11 40 46 AM

With native tables this should be _much_ easier, since normal `<td>`s and `colspan` can be used and no JS would be required to size anything.

d2l-rubric-criteria-group.js
```diff
@@ -266,149 +254,150 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-criteria-group">
...
-									<input hidden="[[_isStaticView()]]" on-keypress="_handleKey" name="[[criterionNum]]" type="radio" checked="[[_isSelected(criterionCell, cellAssessmentMap)]]" id="criterion-cell-input[[criterionNum]]_[[cellNum]]">
-								</d2l-offscreen>
-							</d2l-td>
-						</template>
-						<template is="dom-if" if="[[_hasOutOf(entity)]]">
-							<d2l-td class$="[[_getOutOfClassName(criterion, criterionResultMap, readOnly)]]">
-								<d2l-rubric-editable-score
-									id="score-inner[[criterionNum]]"
+						</tr>
+						<!--
```
dlockhart, line 366: Here's the next issue -- this offscreen button is being placed in between rows, but `<tr>`s are the only valid child here so this is no longer possible. I'd suggest an alternate solution where the button maybe moves into the last cell of the previous row or something similar.

d2l-rubric-criteria-group.js
```diff
@@ -266,149 +254,150 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-criteria-group">
...
+								on-click="_handleAddFeedback"
+								on-focusin="_handleInvisibleFeedbackFocusin"
+								on-focusout="_handleInvisibleFeedbackFocusout"
+								description="[[_localizeAddFeedbackButtonDescription(criterion)]]"
+								hidden="[[!_showAddFeedback(criterion, criterionResultMap, criterionNum, _addingFeedback, _savingFeedback.*, _feedbackInvalid.*)]]"
+								data-criterion$="[[criterionNum]]"
+							></d2l-button-subtle>
+						</d2l-offscreen>
+						<template is="dom-if" if="[[_displayFeedback(criterion, criterionResultMap, criterionNum, _addingFeedback, _savingFeedback.*, _feedbackInvalid.*)]]" restamp="true">
+							<d2l-tspan id="feedback[[criterionNum]]" role="cell" focused-styling$="[[_isFocusedStyling(_feedbackInvalid.*, criterionNum)]]">
```
dlockhart, line 379: There's another spanning row here to show feedback:

Screen Shot 2021-06-02 at 11 45 05 AM

This should be similarly easy to just convert to a `<td colspan="all-the-cells">`, but getting that little bubble tail to show up will be a bit of extra work.

test/components/default-rubric.js
```diff
@@ -10,6 +10,7 @@ suite('Preview Mode', function() {
 		element.href = '/rubrics/default-rubric';
 		/* global waitForLastSirenChangeEvent */
 		return waitForLastSirenChangeEvent(element, 'd2l-siren-entity-changed', 'Criterion 3').then(function() {
+			// TODO: needs fixing
```
dlockhart, line 13: There are a few places like this where tests or code are searching for the old elements. Much of it likely isn't needed anymore, or can just be switched to look for the native table elements.

MitchelPaulin:
Converted to `d2l-table-wrapper` in the new Lit version of rubrics https://github.com/Brightspace/d2l-rubric/pull/1421

PR #791, opened 2021-06-03
popout rubric radio a11y

semibran:
Rally: [US116198](https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F388950348748)

Uses radio button roles and radio groups to match screen reader output with the design spec.

### DIFF
--- a/d2l-rubric-criteria-group.js
+++ b/d2l-rubric-criteria-group.js
@@ -312,7 +312,12 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-criteria-group">
 			</d2l-thead>
 			<d2l-tbody>
 				<template is="dom-repeat" items="[[_criteriaEntities]]" as="criterion" index-as="criterionNum">
-					<d2l-tr aria-rowindex$="[[_getRowIndex(criterionNum)]]" aria-owns$="[[_getFeedbackID(criterion, criterionResultMap, criterionNum)]]">
+					<d2l-tr
+						role="radiogroup"
+						aria-labelledby$="criterion-name[[criterionNum]]"
+						aria-rowindex$="[[_getRowIndex(criterionNum)]]"
+						aria-owns$="[[_getFeedbackID(criterion, criterionResultMap, criterionNum)]]"
+					>
 						<template is="dom-if" if="[[_showRowHeaders(rubricType)]]" on-dom-change="_rowHeaderDomChange">
 							<d2l-td class="criteria" role="rowheader">
 								<div class="criteria-row-header-container">
@@ -329,7 +334,7 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-criteria-group">
 											></d2l-rubric-competencies-icon>
 										</template>
 										<div class="criterion-name">
-											<span>
+											<span id="criterion-name[[criterionNum]]">
 												[[criterion.properties.name]]
 											</span>
 										</div>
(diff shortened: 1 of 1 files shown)


d2l-rubric-criteria-group.js
```diff
@@ -904,22 +914,53 @@ Polymer({
...
 
+		this._focusCriterionCell(event);
 		this._addingFeedback = -1;
 		this.editingScore = -1;
 	},
 
+	_getRadioSibling: function(radio, delta) {
+		const radios = radio.parentNode.querySelectorAll('[role=radio]');
+		const index = Array.prototype.indexOf.call(radios, radio);
+		return radios[index + delta];
```
semibran, line 929: there's most likely a way to cache entire radiogroups on load that I'm not aware of

d2l-rubric-criteria-group.js
```diff
@@ -904,22 +914,53 @@ Polymer({
...
-		if (event.keyCode === 13 || (this._isSelected(criterionCell, this.cellAssessmentMap) === true && event.keyCode === 32)) { // enter or space key
-			this._handleTap(event);
+		switch (event.code) {
+			case 'Enter':
+			case 'Space':
+				return this._handleTap(event);
+			case 'ArrowLeft':
+				return this._handlePrev(event);
+			case 'ArrowRight':
+				return this._handleNext(event);
```
semibran, line 963: Native radio buttons use up/down arrows for prev/next respectively as well but I figured we didn't need those here since rubric levels are always horizontal. I considered using up/down for movement between criteria but that's outside the scope of this ticket

d2l-rubric-criteria-group.js
```diff
@@ -984,10 +1025,9 @@ Polymer({
 
 	_focusCriterionCell: function(event) {
 		var elem = dom(this.root).querySelector('#criterion-cell-input' + event.model.get('criterionNum') + '_' + event.model.get('cellNum'));
-		fastdom.mutate(function() {
-			// Refocus criterion cell input
+		if (elem) {
 			elem.focus();
-		}.bind(this));
+		}
```
semibran, line 1030: I would prefer to use fastdom here but it looks like it was blocking criterion cell focus from occurring at all until I removed it. Haven't tested if that's an isolated issue

d2l-rubric-criteria-group.js
devpow112: Lol, there probably is. I'm not aware of what.s the right way to do this either. For this it should be fine though.

d2l-rubric-criteria-group.js
```diff
@@ -904,22 +914,53 @@ Polymer({
...
+			this._selectRadio(nextRadio);
+		}
+	},
+
 	_handleKey: function(event) {
-		var criterionCell = event.model.get('criterionCell');
-		if (event.keyCode === 13 || (this._isSelected(criterionCell, this.cellAssessmentMap) === true && event.keyCode === 32)) { // enter or space key
-			this._handleTap(event);
+		switch (event.code) {
+			case 'Enter':
```
devpow112, line 957: I used to using `keyCode` I'm assuming `code` is also consistent across browsers

devpow112 (approved):
Looks good to me :tada:

d2l-rubric-criteria-group.js
semibran: yeah I can imagine there's some discussion to be had about this, `keyCode` is deprecated but has better compatibility, while `code` is more readable and more future-proof. On second thought it actually might be better to use `KeyboardEvent.key` in this context but 90% of the time there really shouldn't be any discernable difference

PR #794, opened 2021-06-04
mobile rubric grading status

semibran:
Rally: [US124152](https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F484015978896)

Adds more detail to the status blurb in the inline grading accordion:
- Ungraded rubrics will display "Not scored"
- Partially-graded rubrics display the number of graded criteria
- Graded rubrics display the points : total points ratio (which is the current behavior for everything), but only if the rubric in question is points-based; points-based rubrics just display e.g. "3 / 3 criteria scored"

Requires binding the criterion assessment map all the way up to the `d2l-rubric` level in order to resolve assessment entity hrefs in the adapter.


d2l-rubric-adapter.js
```diff
@@ -11,7 +12,8 @@ import { mixinBehaviors } from '@polymer/polymer/lib/legacy/class.js';
  * owning the specific rendering logic.
  */
 window.customElements.define('d2l-rubric-adapter', class RubricAdapter extends mixinBehaviors([
-	D2L.PolymerBehaviors.Rubric.LocalizeBehavior
+	D2L.PolymerBehaviors.Rubric.LocalizeBehavior,
+	window.D2L.Hypermedia.HMConstantsBehavior
```
devpow112, line 16: Don't think you need the `window.`

devpow112 (approved):

d2l-rubric-adapter.js
semibran: You're right, though `window.D2L` appears like 60+ times in rubrics for whatever reason so it'd be a pain to try and weed everything out :disappointed:

PR #799, opened 2021-06-08
Allow edit rubric title header to be hidden

annabelsegalld2l:
This will allow us to render the title outside the d2l-rubric-editor component, and hide the built-in title. 
The title-hidden tag is used in https://github.com/BrightspaceHypermediaComponents/activities/pull/1769


editor/d2l-rubric-editor.js
```diff
@@ -455,6 +457,10 @@ Polymer({
 			type: Boolean,
 			value: false
 		},
+		titleHidden: {
+			type: Boolean,
+			value: false
```
annabelsegalld2l, line 462: Will default to false, so the title will be shown unless the title-hidden tag is applied.

editor/d2l-rubric-editor.js
```diff
@@ -231,7 +231,9 @@ const $_documentContainer = html `
 			}
 		</style>
 		<d2l-rubric-editor-header id="rubric-header">
-			<div slot="title">[[localize('editRubric')]]</div>
+			<template is="dom-if" if="{{!titleHidden}}">
```
svanherk, line 234: So Polymer supported both one way data-binding with `[[ ]]` (ex. `if="[[!titleHidden]]"`) and two-way data-binding with `{{ }}` (ex. `if="{{!titleHidden}}"`).  In one-way data-binding, the parent changes the data and the child gets notified.  In two-way data binding, the child could also change the data and the parent would get notified.  Two-way binding is powerful, but has issues and Lit doesn't support.  We knew Lit was where we were heading so we stopped trying to use two-way data binding altogether, and instead notify the parent of child data changes through events.  But you'll still see uses of two-way data binding in older components.

The good news is that 90% of the time two-way binding isn't even needed, and was probably spread through copy/paste and looking at examples.  So long story short lol... you should be able to change this to `[[!titleHidden]]` without any issues, and that'll make a future Lit migration easier.

svanherk: (more info if interested: https://polymer-library.polymer-project.org/3.0/docs/devguide/data-binding)

annabelsegalld2l: Thanks for the catch! It worked fine with the one-way binding.

svanherk (dismissed):

mdulat (approved):

mdulat (approved):

PR #803, opened 2021-06-09
fix: mobile rubric parallel actions

semibran:
> EDIT: On hold for now until `rubrics-inline-grading` feature flag is lifted, or global assessment handlers can be disambiguated properly

Runs siren actions in parallel instead of in series.

Requires some extra blocking conditions to the reassessment handler since it looks like polymer is flagging the criterion assessment entity twice on (most) every action complete..... so essentially the handler fires once with the old entity and once with the new entity. This behavior doesn't occur when performing changes in series, so it might have something to do with the number of active queries


devpow112 (approved):

mpharoah-d2l (changes requested):
I don't think we want to completely remove the promise queue. That was added to ensure a consistent ordering of requests, and without it, we may get unexpected behaviour and failing requests

semibran:
Closed for now, too many blockers on this

PR #806, opened 2021-06-09
changed cachee to masquee

joshvanwijk:
as per tELO request


devpow112:
@joshvanwijk to get CI to pass you will need to run the steps here to build the language files.
https://github.com/Brightspace/d2l-rubric#modifying-a-lang-term

PR #819, opened 2021-06-14
Change rubric editor from using MVC to web-component dialogs

annabelsegalld2l:
This should not affect the functionality of the edit rubrics page or the dialogs. The only difference between the MVC dialogs and the new web component dialogs is the missing 'x' at the top right of the dialog, as seen here: 
dialog-confirm


editor/d2l-rubric-dialog-behavior.js
```diff
@@ -25,30 +22,5 @@ D2L.PolymerBehaviors.Rubric.DialogBehavior = {
 			default:
 				return 'd2l-dialog-inline-body-confirm';
 		}
-	},
-	openConfirm: function(settings) {
```
annabelsegalld2l, line 29: This function was just used by the four files I changed, so it should be safe to remove.

editor/d2l-rubric-criterion-editor.js
```diff
@@ -286,6 +286,13 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-criterion-editor">
 			</d2l-select-outcomes-hierarchical>
 		</d2l-dialog>
 
+		<d2l-dialog-confirm id="deleteCriterionDialog"
```
annabelsegalld2l, line 289: The criterion editor, level editor, overall level editor, and structure editor file changes are very similar. Instead of calling the MVC inline dialog to appear, open a d2l-dialog-confirm instead.

editor/d2l-rubric-criterion-editor.js
```diff
@@ -890,11 +902,13 @@ Polymer({
 		if (!action) return;
 		var deleteButton = e.currentTarget;
 		deleteButton.setAttribute('disabled', '');
-		this.openConfirm({
-			title: this.localize('deleteCriterionConfirmationTitle'),
-			secondaryMessage: action.title || this.localize('deleteCriterionConfirmationText'),
-			positiveButtonText: this.localize('deleteConfirmationYes'),
-			negativeButtonText: this.localize('deleteConfirmationNo')
+		var dialog = this.shadowRoot.getElementById('deleteCriterionDialog');
+		new Promise(function(resolve, reject) {
```
annabelsegalld2l, line 906: I tried to replicate the logic of the openConfirm function here. If the dialog closes with the successful action ('delete'), resolve. If the dialog is cancelled, then reject.

test/components/d2l-rubric-criterion-editor.js
```diff
@@ -161,15 +161,15 @@ suite('<d2l-rubric-criterion-editor>', function() {
...
 
 			test('generates event if deleting fails', function(done) {
-				fetch = sinon.stub(window.d2lfetch, 'fetch');
-				var promise = Promise.resolve({
-					ok: false,
-					json: function() {
-						return Promise.resolve(JSON.stringify({}));
-					}
+				element.addEventListener('d2l-dialog-open', function() {
+					element.$$('#deleteCriterionDialog').dispatchEvent(new CustomEvent(
```
annabelsegalld2l, line 165: Previously, this test was creating a stub for the openConfirm promise. Instead, we can listen to the open event, and trigger a close immediately with the successful action (which will resolve the promise).

editor/d2l-rubric-criterion-editor.js
svanherk: This is another case where you should be able to use one-way binding `[[ ]]` in all of your changes across the four files (instead of `{{ }}`).

editor/d2l-rubric-criterion-editor.js
svanherk: They don't seem to `catch` to handle the rejection if the delete is canceled.  Does that result in an uncaught exception in the console?  Did it before?

svanherk: Also, it looks like before `openConfirm` was returning the promise and now you're not, but it _seems_ like it doesn't matter (the value isn't used for anything)

editor/simple-overlay.js
```diff
@@ -1,43 +0,0 @@
-/**
```
svanherk, line 1: 🎉

editor/d2l-rubric-criterion-editor.js
annabelsegalld2l: I think the value is used, since the criterion is only deleted in the resolve() case... in the cancel case, I can't see any errors in the console popping up, event though it is a reject (no errors before either).

svanherk: Ah I see, I missed the second function in `then` for `onRejected` that just removes the `disabled` attribute

devpow112 (approved):

PR #825, opened 2021-06-15
Try and fix issues between rubrics and consistent eval

devpow112:
Seems to be related to the locale work at least partially. Reverting for now to eliminate possible issues.


devpow112:
It's weird cause this works in practice (lms, demos) but not for consistent eval repo tests. I don't have the desire to figure out that repos issues at the moment though

MitchelPaulin (approved):

semibran (approved):

PR #830, opened 2021-06-15
fix: translation updates

semibran:
Adds the "scoreOutOfAria" langterm for screen reading criterion level scores.

Also renames `{score}` to `{level}` under the `"levelSelected"` field for accuracy.


devpow112:
The stuff in `{}` is used by code so you'll need tp update the location these are used as well to use the new replacement term `level`.
https://github.com/Brightspace/d2l-rubric/blob/master/d2l-rubric-levels-mobile.js#L527

devpow112:
NM looks like you're doing that as part of https://github.com/Brightspace/d2l-rubric/pull/829. Why not include these changes in that PR?

devpow112 (approved):

PR #836, opened 2021-06-16
US129172 - Remove active state

devpow112:
Didn't bump the version. We can bump it when other stuff goes out.


assessment-helpers/criterion-assessment.js
```diff
@@ -84,6 +84,10 @@ class CriterionAssessmentHelper {
 		return entity && entity.hasActionByName('update-critierion-assessment');
 	}
 
+	canSelectCriterionCell(entity) {
+		return entity && entity.hasActionByName('select-criterion-cell');
+	}
+
```
semibran, line 90: This method already exists under CriterionCellAssessmentHelper, is there a reason to include it twice?

devpow112: Nope, just missed it. Will switch it over.

semibran:
this isn't super high priority at the moment, but it might not be necessary to create a distinction between `.assessed` and `.selected` at all. Especially if we're trying to prevent the empty tile state from occurring

devpow112:
> this isn't super high priority at the moment, but it might not be necessary to create a distinction between `.assessed` and `.selected` at all. Especially if we're trying to prevent the empty tile state from occurring

There were some weirdness with the student preview mode when I tried to just reply on selected. We can revisit them later for now this keeps them separate since I couldn't figure it out

semibran (approved):

PR #838, opened 2021-06-16
US129172 - Allow deselection

devpow112:
Did this off your since I used `selectLevelAt`. Tested with master merged as well and this works with the removal of active state stuff.


d2l-rubric-levels-mobile.js
```diff
@@ -395,13 +395,25 @@ Polymer({
...
-		if (this.focused === level.dataIndex
-		&& this.selected === level.dataIndex) {
+		if (this.focused === level.dataIndex &&
+			this.selected === level.dataIndex) {
 			return;
 		}
-		this._focusLevel(level.dataIndex);
+
+		if (this.selected === level.dataIndex) {
+			this.selected = undefined;
```
semibran, line 404: I've typically been setting selected/hovered/focused to -1 when they're not in use, though I think the default value is undefined so this probably doesn't break anything

d2l-rubric-levels-mobile.js
```diff
@@ -395,13 +395,25 @@ Polymer({
...
+		}
+
+		this.focusSlider();
+
 		if (!this.readOnly) {
-			this._getCriterion().selectCell(() => this.cellAssessmentMap[level.dataset.cellHref]);
+			const levelCellHref = level.dataset.cellHref;
+			const cellAssessment = this.cellAssessmentMap[levelCellHref];
+
+			this._getCriterion().selectCell(() => cellAssessment);
```
semibran, line 416: I'd be careful about streamlining the entity getter like this..... it looks like the cell assessment map is subject to change for some reason, so I've run into some issues with this kind of thing while working with selection logic before. Maybe you're a bit more familiar than me about why that might be the case.

d2l-rubric-levels-mobile.js
```diff
@@ -451,7 +463,9 @@ Polymer({
 	},
 
 	_handleThumbClick: function() {
-		this.focusSlider();
+		if (this.selected >= 0) {
+			this.selectLevelAt(this.selected);
+		}
```
semibran, line 468: May or may not conflict with #823

d2l-rubric-levels-mobile.js
devpow112: K will revert

d2l-rubric-levels-mobile.js
devpow112: This worked in my testing. Let me try `-1` and see how it goes.

devpow112:
Tests timed out. I'm gonna increase our test timeout

d2l-rubric-levels-mobile.js
devpow112: Seemed to work the same with `-1` but feels safer. Changed.

d2l-rubric-levels-mobile.js
devpow112: Reverted

semibran (approved):

PR #842, opened 2021-06-17
feat: switch to using the new HTML editor for rubric editing

Callum-Mitchell:


devpow112:
Make sure when you merge this that the merge commit title looks something like this `feat: switch to using the new HTML editor for rubric editing`. Since none of your commits include a prefix since they predate the auto release stuff you'll want this to trigger a release

semibran (approved):
LGTM 👍

devpow112:
This unfortunately was missing the `feat:` so no release was done 😞.

PR #847, opened 2021-06-17
Overriden score fixes in graded mode

MitchelPaulin:
Overridden score now shows in graded mode and overridden tooltip does not pop up in graded mode


d2l-rubric-criterion-mobile.js
```diff
@@ -479,18 +479,12 @@ Polymer({
...
-	_getVisibleLevelPoints: function(selected, hovered, focused, levels, cells) {
+	_getVisibleLevelPoints: function(selected, hovered, focused, levels) {
 		const level = this._resolveSelection(selected, hovered, focused, levels);
-		if (!level) {
-			return null;
+		if (level) {
+			return level.properties.points;
 		}
-		// check for overrides
-		let points = level.properties.points;
```
MitchelPaulin, line 488: For me this code just doesn't work at all. This never would properly return the overridden score to the component. The overriden score however is saved in `_score` so you can just return that. Not sure if there is some reason this was here, but everything seems to function without it...

d2l-rubric-editable-score.js
```diff
@@ -195,7 +195,6 @@ Polymer({
 		},
 		score: {
 			type: Number,
-			value: -1,
```
MitchelPaulin, line 198: Bug from previous PR. If there was no grade sometimes you would get `-1/4` instead of ` /4`

d2l-rubric-editable-score.js
```diff
@@ -297,6 +296,9 @@ Polymer({
 	_scoreChanged: function() {
 		if (this._shouldShowOverrideState(this.scoreOverridden, this.readOnly, this.hovered) && !this.totalScore) {
 			this.score = this.overriddenScore;
+			this.overriddenStyling = true;
```
MitchelPaulin, line 299: Fixes the small styling issue Brandon pointed out

d2l-rubric-editable-score.js
semibran: Observed same bug and same fix 🙂

semibran (approved):

devpow112 (approved):

PR #850, opened 2021-06-18
DE44044 - Remove tab trap

devpow112:
This removes the tab stop on entering the override element and re-adds it on exiting. Also removed uses of `var` in the code I was changing.


devpow112:
Not sure if this will work. Still trying to test on our janky Mac set-up.

d2l-rubric-editable-score.js
```diff
@@ -368,31 +370,52 @@ Polymer({
 		this._scoreModified = true;
 	},
 
+	_focusHandler: function(e) {
+		if (!e || !e.target) {
+			return;
+		}
```
semibran, line 376: I don't think it's possible for the event to come up undefined in this case

devpow112: Fair, I'm just used to null checking everything. I can remove it if we want.

semibran: the things i'd give to setup optional chaining in this repo.......

semibran (approved):

d2l-rubric-editable-score.js
devpow112: Lol someday...

PR #861, opened 2021-06-23
Not Scored no longer shows up if the score was overriden

MitchelPaulin:
overriden-no-level
overriden-with-level

https://rally1.rallydev.com/#/57732444928ud/custom/373260458992?detail=%2Fdefect%2F602359390443


semibran (approved):

devpow112 (dismissed):

devpow112:
@MitchelPaulin make sure to bump the version. Based on the way we were doing versioning version bump isn't needed when going into the `release/20.21.7` branch only `master`.

PR #862, opened 2021-06-23
DE44044 - Fix editable score tab trap

semibran:
don't ask questions, just hit approve........


d2l-rubric-editable-score.js
```diff
@@ -367,6 +367,10 @@ Polymer({
 	},
 
 	_blurHandler: function(event) {
+		if (!event.sourceCapabilities) {
```
devpow112, line 370: Really!? That's it? How did you figure that out (I realize you said don't ask question but I must 😛)

devpow112 (approved):

devpow112:
Oh also should probably version bump

PR #863, opened 2021-06-23
DE44145 - Disable feedback input on clear

devpow112:


semibran (approved):

d2l-rubric-feedback.js
```diff
@@ -73,10 +73,13 @@ $_documentContainer.innerHTML = /*html*/`<dom-module id="d2l-rubric-feedback">
 				border: solid transparent 1px;
 				cursor: pointer;
 			}
-			.clear-feedback-button.is-focused,
-			.clear-feedback-button:hover {
+			.clear-feedback-button:not([disabled]).is-focused,
```
devpow112, line 76: Styling is needed since the item is a `d2l-icon` not a `d2l-button-*` type. Tried switching to a `d2l-button-icon` but the styling is problematic as it isn't allowed to be as small as we need. Resizing it doesn't work as the size of the icon button is hard-coded and didn't seem to be overridable.

d2l-rubric-feedback.js
```diff
@@ -359,8 +369,14 @@ Polymer({
 	},
 
 	_clearFeedbackHandler: function() {
+		if (this._disabled) {
```
devpow112, line 372: Event source is not a button so it can't really be disabled. Doing this check to make sure multiple click in the disabled state aren't processed.

d2l-rubric-feedback.js
```diff
@@ -433,10 +449,13 @@ Polymer({
 	},
 
 	_handleVisibleFocusin: function() {
-		var elem = dom(this.root).querySelector('#clear-feedback-invisible');
-		fastdom.mutate(function() {
-			elem.focus();
-		}.bind(this));
+		if (this._disabled) {
```
devpow112, line 452: Event source is not a button so it can't really be disabled. Doing this check to make sure multiple click in the disabled state aren't processed.

PR #874, opened 2021-06-30
US124186 - Overall score criterion in compact view

semibran:
Rally: [US124186](https://rally1.rallydev.com/#/57732444928d/custom/373260458992)

There's a lot going on here but it essentially boils down to three major refactoring goals:
1. Move action handling and arrow buttons into levels component wholesale
2. Create an `overall-score-mobile` component that shares the levels component with the criterion component
3. Create a `total-score` component for reuse just before the overall score component in compact view

More details on design and behavior in [the spec](https://xd.adobe.com/view/a9f04c8b-d935-4a84-b082-d1eb9abadd33-fca3/)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "d2l-rubric",
-  "version": "3.14.3",
+  "version": "3.15.0",
```
MitchelPaulin, line 3: The release process for rubrics was recently updated https://github.com/Brightspace/d2l-rubric#versioning--releasing, you should no longer need to update package.json manually

d2l-rubric-levels-mobile.js
```diff
@@ -508,6 +705,16 @@ Polymer({
 		}
 	},
 
+	_getVisibleLevel: function(selected, hovered, focused, saved) {
+		return hovered !== -1
```
MitchelPaulin, line 709: I think a simple if/else if chain here would be much more readable

d2l-rubric-levels-mobile.js
```diff
@@ -462,15 +554,107 @@ Polymer({
...
+		const assessmentAction = assessmentEntity && assessmentEntity.getAction();
+		return cellLink && assessmentAction && cellLink.href === assessmentAction.href;
+	},
+
+	_getAssessmentHref: function(assessmentEntity) {
+		return this._getSelfLink(assessmentEntity) || this._getSelfLink(this.entity);
+	},
+
+	_canUpdateAssessment: function(assessmentEntity) {
+		return assessmentEntity && (
```
MitchelPaulin, line 611: Can this be broken up into multiple statements? As is this is hard to read and the `!--this._action.ops` means this has side effect which could be even more confusing

semibran: i'm getting deja vu......

d2l-rubric-levels-mobile.js
```diff
@@ -706,13 +709,18 @@ Polymer({
...
+		if (hovered !== -1) {
+			return hovered;
+		}
+		if (focused !== -1) {
+			return focused;
+		}
+		if (selected !== -1 || !this._overallScore) {
+			return selected;
+		}
+		if (saved !== -1) {
```
MitchelPaulin, line 721: Should this not return saved by default to keep with the original logic? Does this work fine if execution falls all the way through to the bottom?

MitchelPaulin:
To use the new release system make sure that at least one commit is prefixed with `feat:` for a minor release. https://github.com/Brightspace/d2l-rubric#versioning--releasing

d2l-rubric-levels-mobile.js
semibran: Yeah it's a bit tough to explain but the idea is that the behavior should be different for criteria levels vs. overall score levels. With criteria you want changes to be reflected immediately for perceived performance even if they aren't saved, but the behavior when selecting a cell for overall score levels is different since it functions via overrides - even if you deselect a level, you don't want the selection tile to disappear which is why there's the saved level fallback.

Basically without that extra check, overall score would still work fine but criteria would feel laggy (in theory..... criteria don't actually use this function yet though but I'll address that in a future PR 😓)

d2l-rubric-levels-mobile.js
```diff
@@ -895,5 +1074,53 @@ Polymer({
 		setTimeout(() => {
 			this._preventClick = false;
 		});
-	}
+	},
+
+	_getIteratorIndex: function(selected, focused) {
+		return focused !== -1
```
MitchelPaulin, line 1080:
```suggestion
		return focused !== -1 ? focused : selected;
```

MitchelPaulin (dismissed):

semibran:
CI was breaking because of some stray lines

MitchelPaulin (approved):

PR #880, opened 2021-07-07
Added cache invalidation to level editor

MitchelPaulin:
Same issue and same fix that was addressed in https://github.com/Brightspace/d2l-rubric/pull/875, this one for the level name


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "d2l-rubric",
-  "version": "3.15.0",
+  "version": "3.15.1",
```
devpow112, line 3: You don't need to do this manually anymore thanks to @imcdonald-d2l. If you revert this and just make sure your merge commit includes `fix: ` it will auto increment the version and make a release.

devpow112: This is getting implemented on all our repos so it will make releasing much nicer.

devpow112 (approved):

PR #881, opened 2021-07-07
US129370 - Switch to serge JS

devpow112:


djwester (approved):
Can we just do a quick sanity check on an LMS build after this is merged to make sure the translations are still lining up ok?

devpow112:
> Can we just do a quick sanity check on an LMS build after this is merged to make sure the translations are still lining up ok?

Yup that's the plan. I tested locally with demos but definitely want to do more.

PR #898, opened 2021-07-14
test: convert over to karma (US129369)

### LINKED ISSUE #723: Investigate flaky test
It seems like even after increasing the timeout that tests will sometime hang. Enable verbose test execution output and see if we can figure out which test is hanging an why.

devpow112:
This converts over existing tests to use `karma` and more modern frameworks for unit testing. It also expands our testing platforms from just **Chrome** on **Linux** running on the CI runner to **Firefox**, **Safari**, **Chrome** and **Edge** on a combination of **Windows** and **MacOS** using **SauceLabs**. This also fixes local tests which had stopped working due to issues with `polymer test` (couldn't install/detect needed stuff for Chrome locally anymore). Test validity was not taken into account much as this is just meant to be a straight conversion.

Resolves #723 (hopefully 🤞).


devpow112:
Abandoning for now

PR #899, opened 2021-07-15
fix: Brightspace editor creation fixes

Callum-Mitchell:
Fixes 2 problems with the new editor in the creation page:
- Rubric description now has a larger height, preventing click difficulties
- Criterion level description/initial feedback cells: editor now expands to fill full cell height (with a set minimum height to avoid an unusably small edit window)


devpow112 (approved):

devpow112:
None of the commits had `fix:` so it didn't trigger a release 😞

PR #905, opened 2021-07-20
De44409 override score not appearing correctly

MitchelPaulin:
https://rally1.rallydev.com/#/57732444928ud/custom/373260458992?detail=%2Fdefect%2F604059071761

Fixes problem where overridden score would not show up and fixes the problem where you need to wait for the web request to succeed before the override state is cleared. This optimistically clears the override state which makes things feel more responsive.


devpow112 (approved):
Looks good to me :tada:

PR #907, opened 2021-07-21
US129838 - Make locked rubrics partially editable client-side

semibran:
Rally: [US129838](https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F603673038897)

[Design spec](https://xd.adobe.com/view/4307335f-803d-4a0a-ba88-1165f7c03bf6-1654/screen/1c3ac2fc-9b85-46b3-995c-63c0eb9d41b2/)

Adds the `isPartiallyEditing` attribute to the rubric editor to make points, outcomes, and rubric structure read-only on the frontend. Backend stuff gets handled in [Brightspace/lms#12620](https://github.com/Brightspace/lms/pull/12620)

Review with whitespace diff off


editor/d2l-rubric-editor.js
```diff
@@ -848,8 +855,7 @@ Polymer({
 			const isReadOnly = entity.hasClass('readOnly');
 			this._isLocked = isLocked;
 			this._isShared = isShared;
-			this._isReadOnly = isReadOnly || isLocked || isShared;
-
+			this._isReadOnly = this._isLocked || isShared || isReadOnly;
```
semibran, line 858: the read-only prop can be true by default if the rubric is locked, then the alert stuff can just set this prop to false manually. That'll trigger a redraw of the editor component with the partially editable view toggled on

test/components/static-data/rubrics/organizations/text-only/199.json
```diff
@@ -7,21 +7,6 @@
...
-          {
-            "class": [
-              "required"
-            ],
-            "type": "text",
-            "name": "name",
-            "value": "Text Only Rubric"
-          }
-        ]
-      },
```
semibran, line 24: Removed to pass CI, but this test rubric is in an impossible state anyways (writable but missing update actions) which makes some of these tests not so useful.

PR #909, opened 2021-07-22
fixing initial feedback formatting and feedback borders for inline grading

gfrancu:


devpow112:
Just an FYI, you need to have `fix:` in one of your commits to trigger an automatic release :). You can do this on merge by changing the title of the merge commit to have `fix:` at the very front.

image

devpow112 (approved):

MitchelPaulin (approved):

PR #914, opened 2021-07-23
US129837 -- Implemented alert and confirmation dialog for switching to partial edit mode

mpharoah-d2l:
When the `locked` attribute is present, but `readOnly` is not, the alert now has an Edit Text button with a dialog confirmation. Confirming it sets `_textEditable` to `true` while `_structureEditable` remains `false`. This is not fully functional yet, since it requires additional changes to make the rubric actually respond to these properties.


devpow112:
Just a reminder that you'll need to have a commit with either `fix:` or `feat:` for the auto release stuff to do it's thing
https://github.com/Brightspace/d2l-rubric#versioning--releasing
For this it's probably good for a `feat:` since it's a new feature. You can take care of this when you merge by changing the title of the merge commit to have `feat:` at the beginning.
image

semibran:
> For this it's probably good for a `feat:` since it's a new feature. You can take care of this when you merge by changing the title of the merge commit to have `feat:` at the beginning.

We can probably merge this stuff onto a working branch and then merge to master as one feature since it's reliant on the partial edit view

MitchelPaulin (approved):

PR #917, opened 2021-07-26
fix: score rendering

devpow112:
We were trying to set a value that affected rendering in the middle of a render already taking place. Wait till current update is complete then change value to issue proper render.


d2l-rubric-editable-score.js
```diff
@@ -393,28 +393,24 @@ Polymer({
...
+	_changeInputHandler: function() {
 		this._scoreModified = true;
 	},
 
-	_inputHandler: function() {
-		this._scoreModified = true;
-	},
-
-	_blurHandler: function(e) {
+	_blurHandler: async function(e) {
```
mpharoah-d2l, line 413: Do async methods work for event handlers?

devpow112:
NM this doesn't work fully still

d2l-rubric-editable-score.js
devpow112: They do but this causes other issues. Just abandoning for now as I look into how to make polymer behave 😞

PR #923, opened 2021-07-27
fix: possible issue with BSI updates

devpow112:
Make `@brightspace-ui/htmleditor` inline with BSI `package.json`. Also adjust `@brightspace-ui-labs/accordion` to use major range for updates.


semibran (approved):
so the autoupdate stuff is broken just cause some deps are still locked into a specific commit/version?

devpow112:
> so the autoupdate stuff is broken just cause some deps are still locked into a specific commit/version?

Yeah, we aren't actually the problem but it's what is causing the issue. We need to mimic what's in the BSI `package.json` to make sure to not have multiple versions of the same production dependencies installed.

PR #930, opened 2021-07-29
fix: Added functionality to switch to old html edtior in rubric create experience

MitchelPaulin:
https://rally1.rallydev.com/#/57732444928ud/custom/373260458992?detail=%2Fdefect%2F605369054319

https://github.com/Brightspace/lms/pull/12857


devpow112:
One thing to note is when backporting this to the `3.18.x` branch which represents cert you'll want to have your commit be `fix:` instead of `feat:`. Having it be `feat:` will cause the release to fail due to how semantic release maintenance branches work. Maintenance branches can only have patch style release increments.

editor/d2l-rubric-text-editor.js
```diff
@@ -106,6 +134,18 @@ Polymer({
 		}
 	},
 
+	_shouldUseBrightspaceEditor: function(useNewBrightspaceEditor, richTextEnabled) {
+		return useNewBrightspaceEditor && richTextEnabled;
+	},
+
+	_shouldUseTextEditor: function(useNewBrightspaceEditor, richTextEnabled) {
+		return !useNewBrightspaceEditor && !richTextEnabled;
+	},
```
Callum-Mitchell, line 143: Seems like if the new editor flag is on, but rich text is disabled, none of the editors will be included. I think we should always use the text editor when rich text being disabled, regardless of the new editor flag's value.

Callum-Mitchell (approved):

devpow112 (approved):

PR #942, opened 2021-08-05
DE44534 - Readd popout rubric assessment clearing

semibran:
[Rally](https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fdefect%2F605059888323)

Fixes a regression that got bundled in with the inline rubrics revamp perf improvements. Basically criterion cells in the popout view stopped being deselectable since I assumed (incorrectly) that deselecting radio buttons was undesired behavior.


devpow112 (approved):
Looks good to me :tada:

PR #950, opened 2021-08-06
fix: clear override need to click multiple times (DE44564)

devpow112:
Change `focus`/`keydown`/`click` handler away from the entire component and only on the score part of the control. This removes the click handler from getting fired when you click the button. Fixes [DE44564](https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fdefect%2F605237900265).

| **Before** | **After** |
|---|---|
|before|after|


semibran (approved):
bless you for the before/after gifs 😌

PR #951, opened 2021-08-06
DE44658 - Display MathML in rubrics

semibran:
[Rally](https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fdefect%2F605849569009)

Switches from `s-html` to [`d2l-html-block`](https://github.com/BrightspaceUI/core/tree/master/components/html-block) in the inline and expanded rubric views to support MathML, since that's one of the richtext features that got introduced with the brightspace editor changes.


semibran:
Very minor regression here btw since `<style>` blocks don't work properly in the new html component, probably because of shadow root stuff. So there's a bit of extra vertical margins with eg. paragraphs and lists. I'll make a follow-up PR for it if I figure it out

d2l-rubric-criterion-mobile.js
```diff
@@ -149,6 +149,7 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-criterion-mobile">
 			selected="{{_selected}}"
 			hovered="{{_hovered}}"
 			focused="{{_focused}}"
+			level-visible="{{_currentLevel}}"
```
semibran, line 152: Low priority but I've been working on and off on porting all this selected/hovered/focused logic into the levels-mobile component wholesale over the past couple of months so we don't have to repeat it in both the criterion and overall score components. Plan is to abstract it away with a field like `levelVisible`, `visibleLevel`, or `currentLevel` (leaning towards this last one since it has the least syllables)..... though we'll probably end up finishing it during the lit refactor. Don't wanna move too fast and end up breaking stuff 😕

d2l-rubric-criterion-cell.js
```diff
@@ -81,21 +81,23 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-criterion-cell">
...
-			}	
+			}
 		</style>
 		<div class$="[[_getCellClassName(entity)]]" aria-label$="[[_getEmptyLabel(entity)]]">
 			<div class="points custom-points" hidden="[[!_hasCustomPoints(entity)]]">
 				[[_localizeCustomPoints(entity)]]
 			</div>
-			<s-html class="criterion-description-richtext" html="[[_getDescription(entity)]]"></s-html>
+			<d2l-html-block>
+				<template id="criterion-description-html"></template>
```
imcdonald-d2l, line 91: Could we not do the following instead of using `shadowRoot.querySelector` to populate html? 🤔 
```
<template id="criterion-description-html">[[_getDescriptionHtml(entity)]]</template>
```

semibran: Lol forgot to comment on this. Basically there's some extra oddities when using `html-block` with lit integration - usually you would be able to just [use unsafeHTML inside the template](https://github.com/BrightspaceUI/core/tree/master/components/html-block) so you wouldn't have to muck around with innerHTML stuff. But lit-html isn't compatible with polymer so in order to render non-standard HTML (mathjax) we need to use a workaround. Pretty sure I got it from [`BrightspaceHypermediaComponents/sequences`](https://search.d2l.dev/xref/BrightspaceHypermediaComponents/sequences/components/d2l-sequences-content-module.js?r=08d09249)

imcdonald-d2l: Good to know, thanks. Can't wait till we can convert to lit so we don't have these odd issues!

But yes, I worked in sequences land when I was on team phoenix when I started at D2L.

d2l-rubric-criterion-mobile.js
```diff
@@ -170,12 +171,11 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-criterion-mobile">
...
 				</div>
-				<div hidden="[[!_getVisibleLevelDescription(_selected, _hovered, _focused, _criterionCells)]]" class="criterion-description">
+				<div hidden="[[!_getCurrentLevelDescription(_currentLevel, _criterionCells)]]" class="criterion-description">
 					<p class="criterion-description-html">
-						<s-html
-							class="criterion-description-html"
-							html="[[_getVisibleLevelDescription(_selected, _hovered, _focused, _criterionCells)]]"
-						></s-html>
+						<d2l-html-block>
+							<template id="criterion-description-html"></template>
```
imcdonald-d2l, line 177: Looks like this file is missing the `observer` function:
```
_onEntityChanged: function(entity) {
		if (!entity) return;
		const descHtml = this._getDescription(entity);
		const descWrap = this.shadowRoot.querySelector('#criterion-description-html');
		if (descWrap) {
			descWrap.innerHTML = descHtml;
		}
	},
```

imcdonald-d2l: I take that back... just seeing `_updateCurrentDescription` now...

d2l-rubric-criterion-mobile.js
```diff
@@ -240,6 +240,12 @@ Polymer({
 			value: -1
 		},
 
+		_currentLevel: {
```
imcdonald-d2l, line 243: Where is this property being initialized?

semibran: if you mean where it gets updated that'd be [this guy](https://github.com/Brightspace/d2l-rubric/pull/951/files#diff-8b7d5e263ed2b68912182a7d69e9b46d83efe190e706c9707aa02227b306bdc3R152)...... agreed that upwards data binding is a pain to parse sometimes. Just polymer things :shrug: I mean I guess we could go ahead and use events for this but that might be a bit more trouble than it's worth.

imcdonald-d2l: Yes, I meant updating 😅  Totally forgot about data binding in polymer. Thanks for clarifying.

No need for events, just re-learning polymer again. 😆

imcdonald-d2l (approved):

PR #953, opened 2021-08-09
fix: handle zero scores in hovered criteria

semibran:
Addresses minor defect while testing [Brightspace/lms#13122](https://github.com/Brightspace/lms/pull/13122) - hovering over a level with a score of 0 wouldn't display the actual score of the hovered level

Before | After
-|-
rubric-zero-hovering|rubric-zero-hovering-patch


d2l-rubric-criterion-mobile.js
```diff
@@ -343,9 +343,11 @@ Polymer({
 		const hoveredScore = hoveredLevel && hoveredLevel.properties.points;
 
 		const selectedLevel = levels && levels[selected];
-		const levelScore = selectedLevel && selectedLevel.properties.points;
+		const selectedScore = selectedLevel && selectedLevel.properties.points;
 
-		this._editableScore = hoveredScore || this._score || levelScore;
+		this._editableScore = hoveredScore != null
+			? hoveredScore
+			: (this._score || selectedScore);
```
semibran, line 350: no nullish coalescing moment

imcdonald-d2l:
Do you mind providing a screenshot and/or gif of this working? Would be nice so I have context on where in the UI this is happening 🙂

d2l-rubric-criterion-mobile.js
MitchelPaulin: I wonder how many bugs in js could be avoided if `0 == false` was false. One day we will completely have moved past the `#define true 1` days of C

MitchelPaulin (approved):

d2l-rubric-criterion-mobile.js
MitchelPaulin: Alternatively something like `this._editableScore = [hoveredScore, this._score, selectedScore].find(x => x != null && x != undefined);` might be more expressive, but its up to you for whatever syntax you like better.

semibran:
> Do you mind providing a screenshot and/or gif of this working? Would be nice so I have context on where in the UI this is happening 🙂

Updated description with gifs, though in hindsight just using screenshots might've been easier lol

devpow112 (approved):

d2l-rubric-criterion-mobile.js
semibran: Yeah I guess there's a part of me that would like to use something more expressive but the speed golfing side of me doesn't like the array/fn creation overhead per-redraw for something this small. Even though it's honestly moot from a perf standpoint lol. If we really wanted to be super accurate I'd probably just chain ternaries like this:

```js
this._editableScore =
  hoveredScore != null ? hoveredScore
  : this._score != null ? this._score
  : selectedScore != null ? selectedScore
  : null
```

That'd work provided eslint doesn't whine about indentation. I'm not sure it's necessary to go this hard in this scenario though.

Also minor point of clarification (might've just slipped your mind) but equality is weird in JS..... `x == null` is an idiom for matching either `x === null` or `x === undefined` but not `x === 0`. In this scenario the most concise syntactic equivalent to the above code would be `hoveredScore ?? this._score ?? selectedScore`, but we can't use experimental syntax because polymer locks in our transpiler pipeline 😞

MitchelPaulin: `hoveredScore ?? this._score ?? selectedScore` would be best for sure, but yeah, we can't use `??` :(

PR #955, opened 2021-08-10
fix: empty stack trace being logged to Kibana

MitchelPaulin:
This is hot fixing a fix to the logging for stack traces in 20.21.7 to hopefully get the stack trace info asap.


editor/d2l-rubric-structure-editor.js
```diff
@@ -369,7 +369,7 @@ Polymer({
 			}.bind(this)).catch(function(err) {
 				this.dispatchEvent(new CustomEvent('d2l-rubric-editor-save-error', {
 					detail: {
-						message: err.messagem
```
MitchelPaulin, line 372: Unrelated to the stack trace problem but a typo I found when updating this file

devpow112 (approved):

mpharoah-d2l (dismissed):

devpow112:
This is pre auto release so the release creation is manual. We should probably follow he old convention for consistency. So this would be `v3.13.8-20.21.7-hotfix-3`

MitchelPaulin:
> 
> 
> This is pre auto release so the release creation is manual. We should probably follow the old convention for consistency. So this would be `v3.13.8-20.21.7-hotfix-3`.

As in this should be the name of the branch?

devpow112:
> > This is pre auto release so the release creation is manual. We should probably follow the old convention for consistency. So this would be `v3.13.8-20.21.7-hotfix-3`.
> 
> As in this should be the name of the branch?

Sorry wasn't clear. That should be the tag name when you manually create the release.

devpow112:
Don't forget to do 20.21.8 as well which is the `3.18.x` branch.

PR #961, opened 2021-08-11
update to use correct classes

djwester:
@semibran I think the recent changes you put it made it so that you are using a different method now to get the descriptions which means that DE44659 (https://rally1.rallydev.com/#/57732444928ud/custom/373260458992?detail=%2Fdefect%2F605850175193) is still broken

looking at the PR you fixed this with last time (https://github.com/Brightspace/d2l-rubric/pull/946), I suspect we just need to make the same updates in this new method. 

I also updated it in the legacy view so that it won't break if we switch back to it.

I haven't tested this locally (local BSI being what it is right now), so let me know if you have any concerns with this.


d2l-rubric-criterion-mobile-legacy.js
```diff
@@ -331,7 +331,7 @@ Polymer({
 	},
 
 	_hasDescription: function(criterionCell) {
-		var description = criterionCell.getSubEntityByClass(this.HypermediaClasses.text.description);
+		var description = criterionCell.getSubEntityByClass(this.HypermediaClasse.rubrics.description);
```
semibran, line 334:
```suggestion
		var description = criterionCell.getSubEntityByClass(this.HypermediaClasses.rubrics.description);
```

semibran (commented):
Yup this was an oversight on my part. Thanks for picking up on it

semibran (approved):

PR #966, opened 2021-08-12
fix: use new editor in more places (DE44657)

devpow112:
We missed a few spots when pluming through `use-new-brightspace-editor`. This was causing the old editor to get loaded as well as the new one and v4 and v5 do not play well together. It's mainly death, lots of death, if they are both loaded on the same page. 
[White-space diff off](https://github.com/Brightspace/d2l-rubric/pull/966/files?w=1).

Corresponding FACE change: https://github.com/BrightspaceHypermediaComponents/activities/pull/1934.


editor/d2l-rubric-html-editor.js
```diff
@@ -60,7 +60,7 @@ Polymer({
 			style="[[_getEditorStyle(_editorActive)]]"
 			type="[[_getType(editorType)]]"
 			html$={{value}}
-			no-border
+			no-border$={{!border}}
```
devpow112, line 63: This was needed cause we were disabling the border for all HTML boxes but the one below the rubrics in the options section needed it's border.

editor/d2l-rubric-editor.js
```diff
@@ -377,8 +377,10 @@ const $_documentContainer = html `
 								pending-saves="[[_pendingDescriptionSaves]]"
 								on-text-changed="_saveDescription"
 								rich-text-enabled="[[_richTextAndEditEnabled(richTextEnabled,_canEditDescription)]]"
+								use-new-brightspace-editor="[[useNewBrightspaceEditor]]"
 								html-editor-type="full"
-								html-editor-height="12rem">
+								html-editor-height="12rem"
+								border>
```
devpow112, line 383: This is where it needed a boarder

editor/d2l-rubric-html-editor.js
semibran: Nothing like a little styles in HTML :relaxed:

editor/d2l-rubric-html-editor.js
```diff
@@ -48,8 +48,8 @@ Polymer({
 				margin: 2px;
 			}
 
-			d2l-htmleditor:focus-within,
-			d2l-htmleditor:hover {
+			d2l-htmleditor[no-border]:focus-within,
+			d2l-htmleditor[no-border]:hover {
```
semibran, line 52: Was trying to figure out why there was a border definition under a selector with no-border. But I think I figured it out lol.

semibran (approved):

editor/d2l-rubric-html-editor.js
devpow112: Yeah it's a bit confusing but when the HTML editor has no boarder we add our own custom border.

PR #967, opened 2021-08-13
F17798 - Partial editor cleanup

semibran:
Addresses the four key defects identified as a result of the new partial editor feature work:

### 1. Disabled group copy/delete options
Before | After
-|-
image | image

### 2. Updated dialog text to match updated spec
Before | After
-|-
image | image

### 3. Prevented editing of custom points
Before | After
-|-
image | image

### 4. Fixed error when opening locked holistic rubrics
Before | After
-|-
image | image


editor/d2l-rubric-criterion-editor.js
```diff
@@ -485,7 +497,7 @@ Polymer({
 		},
 		_canEdit: {
 			type: Boolean,
-			computed: '_canEditCriterion(entity)',
+			computed: '_canEditCriterion(entity, partialEditMode)',
```
imcdonald-d2l, line 500: You've added this parameter to `_canEditCriterion()` but the actual function definition needs the additional parameter.

semibran: Good catch, this is unused + just leftover from testing 😞

imcdonald-d2l (approved):

devpow112:
@semibran just and FYI this stuff did not make the cut-off for 20.21.9 so it would need to be manually updated in the 20.21.9 branch in BSI (when it exists) and then the LMS updated. You did make the first release for 20.21.10 though https://github.com/Brightspace/brightspace-integration/releases/tag/v20.21.10-1.

semibran:
> just and FYI this stuff did not make the cut-off for 20.21.9 so it would need to be manually updated in the 20.21.9 branch in BSI (when it exists, currently does not) and then the LMS updated. You did make the first release for 20.21.10 though https://github.com/Brightspace/brightspace-integration/releases/tag/v20.21.10-1.

feels great to be first at something 😐:tada:

PR #968, opened 2021-08-13
US129837 - Tweak locked editable rubric alert button text

semibran:
Just a minor nit that Andrew brought up after demo this afternoon. Last time I tried changing langterm keys it got annoying, but I'm guessing stuff might've changed with the new setup?


MitchelPaulin:
You may want to re run `npm run format` so the trailing `s` gets removed from the other lang term files as well

semibran:
> You may want to re run `npm run format` so the trailing `s` gets removed from the other lang term files as well

I actually did, tried again and it looks like it doesn't actually change anything. Not sure why. I'll just do it manually

MitchelPaulin (approved):

devpow112:
> > You may want to re run `npm run format` so the trailing `s` gets removed from the other lang term files as well
> 
> I actually did that at first, tried again and it looks like it doesn't actually change anything. Not sure why. I'll just do it manually

Looks like the till we use (message format validator) has some bugs and doesn't currently work in this scenario. It will only add completely new terms or remove items not I'm en.js it won't update existing ones in any way

devpow112 (approved):

PR #963, opened 2021-08-11
De44686 inline rubric does not update levels when closing popout rubric

MitchelPaulin:

PR #969, opened 2021-08-16
fix: Deselecting problem if levels were selected very quickly

MitchelPaulin:
Follow up to https://github.com/Brightspace/d2l-rubric/pull/963. The previous PR re-introduced an issue where selecting many levels in succession would cause some to deselect and pick their previous selection, this was caused by the removal of  `_isCriterionSaving`.

Upon further investigation `_isCriterionSaving` always returned false. So this was solving the deselecting issue by essentially never allowing `_selectAssessedLevel` to execute... . This PR keeps the previous solution of removing `_isCriterionSaving` but has a different solution to the deselecting problem. 

`assessmentMap` is shared across all levels, so an update to any level was triggering execution in all level components. Selecting a new cell would cause the selection to update, then the map would be updated. Selecting many cells however would publish many version of the map since they were all being updated during selection, sometimes an outdated version of the map would be used in the execution of `_selecteAssessedLevel` and the cell would be set to what it was before. 

Removing `assessmentMap` from the observer has seemed to resolve the problem of executing `_selectAssessedLevel` unnecessarily, which has fixed the deselecting problem, though its hard to say definitively since the deselecting bug is intermittent.


semibran (approved):
Yeah I'm pretty sure this'll work in theory but we'll just have to see how it plays out in test.

devpow112 (approved):

PR #974, opened 2021-08-18
fix: preselected overall level when grading

MitchelPaulin:
https://rally1.rallydev.com/#/57732444928ud/search?detail=%2Fdefect%2Ff782c700-a238-4ea0-b5d7-118a5c3995d2&keywords=Inline%20grading

The main problem was that `if (negativeNumber)` is true in JS but `if (0)` isn't. Sometimes there would be negative ops, which would evaluate to true when in reality we wanted it to be false


semibran (approved):
Lol god knows why this is happening. If action ops are decremented below zero that probably means assessments are still coming through multiple times for the same level group. I'd think removing the assessment map from the observer would've fixed that, but maybe we need some more checks on the assessment entity..... basically I think adding this check is good but it doesn't address the root of the issue (siren callbacks giving false positives for assessment change events). I feel like siren callbacks are still a bit of a black box to me though so unless we can figure out how to fix these observers I wouldn't be surprised if we had these same kinds of defects pop up even once the lit refactor gets done.

PR #975, opened 2021-08-19
Added action for visual diff test

MitchelPaulin:


.github/workflows/visual-diff.yml
```diff
@@ -0,0 +1,21 @@
+name: Visual Diff
+on: pull_request
+jobs:
+  build:
+    name: Visual Diff
+    timeout-minutes: 10
+    runs-on: ubuntu-latest
+    steps:
+      - uses: Brightspace/third-party-actions@actions/checkout
+      - uses: Brightspace/third-party-actions@actions/setup-node
```
devpow112, line 10: Can you switch this to use  `lts/*` for node version and add the cache attribute. Just help with maintenance
https://github.com/Brightspace/d2l-rubric/blob/master/.github/workflows/ci.yml#L16

test/visual_diff/sanity.visual-diff.js
```diff
@@ -0,0 +1,9 @@
+const puppeteer = require('puppeteer');
```
devpow112, line 1: Can you switch this to be an ES module. Basically just have `import` instead of `require`.
https://github.com/BrightspaceUI/core/blob/main/components/calendar/test/calendar.visual-diff.js

MitchelPaulin: Sure thing. This isn't done, I just wanted to see CI run and make sure the empty test passed

devpow112: Ah ok. Sorry I guess I jumped the gun on that one 😄.

PR #976, opened 2021-08-19
Add logic to auto focus HTML editor - DE44817

imcdonald-d2l:
To focus feedback cells in full-size grading view you have to tab into html-editor and then hit enter. This fix auto-focuses the html editor by taking advantage of the newly added `toggleEditing()` function. (See [readme](https://github.com/BrightspaceUI/htmleditor/blob/main/README.md) for details)

[Rally ticket](https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fdefect%2F606965843651)

### Add new feedback
Before | After 
--- | --- 
feedback-not-auto-focused | feedback-auto-focused

### Focus existing feedback 
Before | After 
--- | --- 
feedback-not-auto-focused-2 | feedback-auto-focused-2


d2l-rubric-feedback.js
```diff
@@ -400,6 +405,13 @@ Polymer({
 		}
 	},
 
+	_focusFeedbackHtmlFromAddFeedbackBtn: function(focus) {
+		if (focus) {
+			setTimeout(() => this._focusHtmlEditor(), 100);
```
imcdonald-d2l, line 410: Have to add timeout to give the editor some time to initialize before focusing it.

d2l-rubric-feedback.js
```diff
@@ -331,6 +331,10 @@ Polymer({
 		_disabled: {
```
imcdonald-d2l, line 331: `d2l-rubric-feedback`  component only exists in  `d2l-rubric-criteria-group` and `d2l-rubric-criteria-group-mobile` so this shouldn't affect other views. Please correct me if I'm wrong though!

d2l-rubric-feedback.js
```diff
@@ -412,6 +424,14 @@ Polymer({
 		const title = this.localize('criterionFeedback');
 		const extra = this._feedbackHeadingExtra;
 		announce(`${title}${extra}`);
+		this._focusHtmlEditor();
```
imcdonald-d2l, line 427: This focuses already existing feedback when tabbing focus.

d2l-rubric-feedback.js
```diff
@@ -412,6 +424,14 @@ Polymer({
...
+	},
+
+	_focusHtmlEditor: function() {
+		if (this.useNewBrightspaceEditor) {
+			const htmLEditor = this._getEditor();
+			if (htmLEditor) htmLEditor.toggleInlineEditing(true);
```
semibran, line 433:
```suggestion
			const htmlEditor = this._getEditor();
			if (htmlEditor) htmlEditor.toggleInlineEditing(true);
```

d2l-rubric-feedback.js
semibran: super minor nit but I think just `setTimeout(this._focusHtmlEditor, 100)` would work here? Unless that runs into binding issues or something

d2l-rubric-feedback.js
semibran: It'd just be `criteria-mobile` not `criteria-group-mobile` but yeah. Looks like there's [a stray import in the editor](https://search.d2l.dev/xref/Brightspace/d2l-rubric/editor/d2l-rubric-criterion-editor.js?r=62e9546a#29) that doesn't get used, that should be the only other place it gets loaded in.

d2l-rubric-feedback.js
imcdonald-d2l: This isn't changing anything... did you mean to type something different? 🤔

d2l-rubric-feedback.js
imcdonald-d2l: Yeah, we need the callback function here or we lose the reference to `this`.

d2l-rubric-feedback.js
imcdonald-d2l: Yes, my bad. You are correct. Removed the stray import 👍🏻

d2l-rubric-feedback.js
imcdonald-d2l: lol, nevermind. I see, I capitalized the `L`

d2l-rubric-criteria-mobile.js
```diff
@@ -176,6 +181,7 @@ Polymer({
 		this._addingFeedback = criterionNum;
 		fastdom.mutate(function() {
 			dom(this.root).querySelector('#feedback' + criterionNum).focus();
+			this._focusFeedbackHtmlEditor = true;
```
imcdonald-d2l, line 184: This adds the focus logic to the "Add Feedback" button in the assignments in-line grading view.

editor/d2l-rubric-html-editor.js
```diff
@@ -160,7 +160,7 @@ Polymer({
 
 	_focusHandler: function() {
 		this._editorActive = true;
-		this._getEditor().focus();
+		this._getEditor().toggleInlineEditing(true);
```
imcdonald-d2l, line 163: The auto-focuses the html-editor for all other instances of the editor. (create rubric experience for example)

d2l-rubric-criteria-mobile.js
```diff
@@ -174,9 +179,7 @@ Polymer({
 	_handleAddFeedback: function(event) {
 		var criterionNum = event.model.get('criterionNum');
 		this._addingFeedback = criterionNum;
-		fastdom.mutate(function() {
-			dom(this.root).querySelector('#feedback' + criterionNum).focus();
-		}.bind(this));
+		this._focusFeedbackHtmlEditor = true;
```
imcdonald-d2l, line 182: This adds the focus logic to the "Add Feedback" button in the assignments in-line grading view.

semibran (dismissed):

imcdonald-d2l:
Not working as expected in in-line grading experience so doing some investigation. 🔍

imcdonald-d2l:
> Not working as expected in in-line grading experience so doing some investigation. 🔍

Issue was that I wasn't setting `notify: true` on property and needed an increase in timeout to successfully focus on editor when clicking on "Add Feedback"

semibran (dismissed):
forcing a delay still feels messy (more so with the magic time delay), is this really the only option?

MitchelPaulin:
Is the large change to the package.json intentional?

imcdonald-d2l:
> Is the large change to the package.json intentional?

Not entirely. Running `npm ci` wasn't getting the html editor update so I had to nuke `package-lock` and re-run `npm i`.

imcdonald-d2l:
> forcing a delay still feels messy (more so with the magic time delay), is this really the only option?

Agreed this is a messy approach. I'll investigate a bit more this morning to see if there is an alternative to the timeout.

imcdonald-d2l:
So after further investigation I noticed that html editor component has a `initializationComplete` promise property which seemed promising at first (pun intended):

image

However, when trying to implement it as follows the promise was not resolving when I hoped it would. It would only resolve after tabbing after hitting the "Add Feedback" button and the html editor was focused manually. Any other thoughts on other approaches instead of using the existing timeout?

```
_focusFeedbackHtmlFromAddFeedbackBtn: async function(focus) {
	if (focus) {
                 // need timeout of 1 to let component render first before getting it
		setTimeout(async () => {
			const htmlEditor = this._getEditor();
			if (htmlEditor && this.useNewBrightspaceEditor) {
				await htmlEditor.initializationComplete;
				this._focusEditor();
			}
		}, 1);
		this.focusFeedbackHtmlEditor = false;
	}
},
```

devpow112 (approved):

semibran:
> Any other thoughts on other approaches instead of using the existing timeout?

Wasn't able to comment on this before it got merged but just to be anal yeah I'd just use a rAF here or try seeing if a zero timeout still works. Not worth filing a separate PR for obviously but if we ever need to do some more html editor tweaks it'd be a nice-to-have just imo. This new setup is still infinitely better than the old magic timeout we had though 😄

PR #981, opened 2021-08-20
Setting up Visual Diff Framework

MitchelPaulin:
Adds a V-diff action and sample test

https://rally1.rallydev.com/#/57732444928ud/custom/373260458992?detail=%2Fuserstory%2F606483897497

Excuse the number of commits, both the linter and the v-diff wouldn't run locally. Only way to test was to keep pushing changes


README.md
```diff
@@ -114,6 +114,28 @@ npm test
...
 
+### Visual Diff Testing
+
+This repo uses the [@brightspace-ui/visual-diff utility](https://github.com/BrightspaceUI/visual-diff/) to compare current snapshots against a set of golden snapshots stored in source control.
+
+The golden snapshots in source control must be updated by Github Actions.  If your PR's code changes result in visual differences, a PR with the new goldens will be automatically opened for you against your branch.
+
+To run visual diff tests locally: 
+
+First add `"type": "module"` to your package.json
```
MitchelPaulin, line 125: Workaround to test locally. I was getting errors about not being able to use an import outside a module. I found two solutions, adding a package.json to the test folder with `"type": "module"` or changing the test files from .js to .mjs. Both of these fixed the change locally but unfortunately broke the v-diff bot.  

Open to any suggestions for a permanent fix.

.github/workflows/visual-diff.yml
```diff
@@ -0,0 +1,21 @@
+name: Visual Diff
```
devpow112, line 1: Could we just add this into the `ci.yml`? It only runs on PR's and seems weird to have a separate workflow for just thing.

.gitattributes
```diff
@@ -1 +1,2 @@
-* text eol=lf
+* text=auto eol=lf
```
devpow112, line 1: Yay! 🎉. Thanks for changing this.

README.md
```diff
@@ -114,6 +114,28 @@ npm test
...
+
+```sh
+# Install dependencies locally
+npm i mocha -g
+npm i @brightspace-ui/visual-diff puppeteer --no-save
```
devpow112, line 130: if you try installing the `esm` module as well does this start working without the need for the workaround?

devpow112: This doc refers to the `esm` module as well https://github.com/BrightspaceUI/visual-diff#installation

.github/workflows/visual-diff.yml
MitchelPaulin: I tried that, there was a 403 error when it tried to run the v-diff. My best guess is because of [this](https://github.com/Brightspace/d2l-rubric/blob/master/.github/workflows/ci.yml#L4). Opening a PR maybe counts as a read permission? I couldn't update the permissions if that preferable

devpow112: Yes it does. Feel free to change that to `write` and see if it works or remove it entirely.

README.md
MitchelPaulin: I did notice this when reading through the documentation but wasn't able to get it to work. Trying it again it still seems to not work unfortunately.

devpow112: 😞. K

.github/workflows/visual-diff.yml
devpow112: Thanks 😄

PR #984, opened 2021-08-23
chore: force new major

devpow112:
This is to give a delineation between polymer and lit versions.

BREAKING CHANGE: lit conversion.


devpow112:
This will make `master` v4. We have a `3.x` branch that will continue to be the non-lit version if there are critical bugs that need to be fixed and released.

MitchelPaulin (approved):

semibran:
`feat` just triggers a minor release right? Unless the semver action reads the breaking change label in the description

devpow112:
> `feat` just triggers a minor release right? Unless the semver action reads the breaking change label in the description

Yeah major releases are triggered by including the `BREAKING CHANGE:` part in the body.

semibran (approved):

PR #989, opened 2021-08-24
Us124512 adding grading status to accordion

MitchelPaulin:
https://rally1.rallydev.com/#/57732444928ud/custom/373260458992?detail=%2Fuserstory%2F484015978896

Once we are happy with the changes I will also open a PR against 3.x


devpow112:
Jut an FYI that language terms changed in this won't be localized for 20.21.9 since we are past the deadline for new terms to be added and dealt with by localization. They should be picked up for 20.21.10 though.

d2l-rubric-adapter.js
```diff
@@ -93,9 +104,21 @@ window.customElements.define('d2l-rubric-adapter', class RubricAdapter extends m
 					display: none;
 				}
 
+				.rubric-header-text {
+					color: var(--d2l-color-chromite);
+					font-size: 14px;
+				}
+
+				.rubric-level-achieved{
```
semibran, line 112:
```suggestion
				.rubric-level-achieved {
```

d2l-rubric-adapter.js
```diff
@@ -207,6 +247,27 @@ window.customElements.define('d2l-rubric-adapter', class RubricAdapter extends m
...
 
+	_getCurrentLevelName(assessmentEntity) {
+		if (!assessmentEntity) return null;
+		const levels = assessmentEntity.getSubEntitiesByClass('assessment-overall-level');
+		for (const level of levels) {
+			if (level.class.includes('selected')) {
+				return level.properties.name;
+			}
+		}
+		return null;
```
semibran, line 258: This might be more trouble than it's worth but in other subcomponents we use `AssessmentBehavior` + `CriterionCellAssessmentHelper.isSelected` to determine level selections

d2l-rubric-adapter.js
```diff
@@ -207,6 +247,27 @@ window.customElements.define('d2l-rubric-adapter', class RubricAdapter extends m
...
+	}
+
+	_getLevelText(assessmentEntity) {
+		const levelText = this._getCurrentLevelName(assessmentEntity);
+		if (levelText) {
+			this.overallScoreSet = true;
```
semibran, line 264: Not sure how I feel about including this as a side effect. Would rather use a computed prop

semibran (commented):
I appreciate you picking this up btw, I know I was originally supposed to take care of this but I must have not scrolled down far enough on the design spec to notice how big the scope of the ticket was supposed to be 😞

d2l-rubric-adapter.js
MitchelPaulin: Good point, I forgot computed properties were a thing

semibran (approved):

PR #990, opened 2021-08-24
chore: clean-up ahead of some lint changes coming

devpow112:
Linting changes based on test work causes this to fail. Trying to minimize my PR on the testing swap to karma since it's quite large.


test/perceptual/sanity.visual-diff.js
```diff
@@ -28,6 +25,8 @@ describe('sanity-test', () => {
 
 	it('first-test', async function() {
 		const rect = await visualDiff.getRect(page, '#test');
+
+		/* eslint-disable-next-line no-invalid-this */
```
MitchelPaulin, line 29: In real tests there will be other uses of `this` which look invalid to the linter which are not. I think we should keep this at the top of the file with the intent to use this file as a template for other tests.

Another option would be to add a rule to ignore these file all together

devpow112: I'll see if I can add a rule for it 👍

test/perceptual/.eslintrc.json
```diff
@@ -0,0 +1,5 @@
+{
```
devpow112, line 1: This will just blanket cover everything in `test/perceptual/`.

test/perceptual/.eslintrc.json
```diff
@@ -0,0 +1,5 @@
...
+  "rules": {
+    "no-invalid-this": 0
```
MitchelPaulin, line 3:
```suggestion
    "no-invalid-this": "off"
```

Recommend using `off` for better readability

MitchelPaulin (approved):

PR #998, opened 2021-08-25
Cleaned up legacy adapters for inline grading component

MitchelPaulin:
Part 1 of the cleanup, cleaning up the inline grading component and removing the legacy adapters


devpow112:
😢

PR #1034, opened 2021-08-30
build: add auto language normalization (US130087)

devpow112:
This will open a new PR with changes to merge into the source PR for any language related stuff. If the branch/PR already exists it will update what's needed. This is the process that visual diff uses and it seems to work well. Decided to just do this in our workflows for now. Still looking at making it a GitHub Action but was running into too much issues.


devpow112:
Right now this will just copy the `en` string everywhere. I want to also try and figure out how to auto translate the fr* stuff using something like AWS translate but tats for a future me.

semibran (approved):

semibran:
So this just autocopies langfiles in a new PR right? does that mean we'll have to manually commit fr* translations onto that PR and then merge?

devpow112:
> So this just autocopies langfiles in a new PR right? does that mean we'll have to manually commit fr* translations onto that PR and then merge?

You can do that or just make sure on the original PR to include the French translations. So you would only need to add stuff to `en.js` and `fr*.js` all the rest will be added automatically in the newly opened PR.

PR #1036, opened 2021-08-30
Updated html editor version

MitchelPaulin:
I was getting the following console errors when editing a rubric

Capture

Caused by this call 

Capture2

This call was fixed in the most recent version of HTML editor [here](https://github.com/BrightspaceUI/htmleditor/blob/main/htmleditor.js#L320). This updates the editor to fix the problem


package.json
```diff
@@ -22,7 +22,7 @@
   "dependencies": {
     "@brightspace-ui-labs/accordion": "^2",
     "@brightspace-ui/core": "^1",
-    "@brightspace-ui/htmleditor": "^1",
+    "@brightspace-ui/htmleditor": "^1.24.1",
```
devpow112, line 25: Can you change this back to just `"@brightspace-ui/htmleditor": "^1",` please and thanks.

devpow112 (approved):

PR #1037, opened 2021-08-30
fix: Removed performance-telemetry-flag from rubrics

MitchelPaulin:
3.x version of  #1035


semibran (approved):
are all the feature flag removals going into 3.x?

PR #1044, opened 2021-08-30
[4.x] DE48817 - Fix inline rubric feedback focus recoil

semibran:
Patch #1043 into lit refactor


devpow112 (approved):

devpow112:
Sauce labs vomited for some reason, re-running CI.

PR #1050, opened 2021-09-03
Added a demo page for mobile rubrics

MitchelPaulin:
Some of the json might be a bit outdated as some things don't render quite right. But we will likely replace most of the data referenced here with data we generate from the postman project anyways.


demo/rubric-mobile.html
```diff
@@ -0,0 +1,87 @@
...
+	<script src="/node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
+	<script type="module">
+		import '@brightspace-ui/core/components/demo/demo-page.js';
+		import 'd2l-fetch-siren-entity-behavior/d2l-fetch-siren-entity-behavior.js';
+		import './fetch-siren-entity-whitelist.js';
+		import '../d2l-rubric.js';
+		import '../d2l-rubric-title.js';
+	</script>
+	<style>
+		d2l-demo-snippet {
```
devpow112, line 18: Can you just set this max width to 700px and forget the other styles? This should make the width of the snippets 700px.

devpow112 (approved):

PR #1051, opened 2021-09-07
fix: popout rubric debounce logic and aria-checked handling

semibran:
https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fdefect%2F607911096743

Had to go a bit out of scope for this one. We needed a local copy of selection state in order for `aria-checked` to stay up to date, which required refactoring a bunch of the debouncer stuff.

Screen reader improvements will be visible in-test, but here's the before and after for the debouncing logic:

## Before
rubrics-popout-debouncing-before

## After
rubrics-popout-debouncing-after

Key debouncing changes:
- Checkmark follows user selection
- Immediately makes assessment if no requests queued or in-flight
- Per-criteria debouncing (as opposed to using one debouncer for the entire criteria group)
- Deselection response tweaks

Once we setup some global helpers for this we can add debouncing to the overall score component. There's a bit of minor inconsistencies between popout and inline rubrics right now so I plan on refactoring the debouncer in inline rubrics again eventually to be more up-to-date with some of the names and behavior stuff.


imcdonald-d2l:
Without looking at the code I'm noticing some weirdness in criterion score in the `After` gif state. It doesn't update with the checkmarks and then does a weird `3 --> 4` update at the end. Is this out of scope for this fix or is this something we should be fixing with this ticket?

semibran:
> Without looking at the code I'm noticing some weirdness in criterion score in the `After` gif state. It doesn't update with the checkmarks and then does a weird `3 --> 4` update at the end. Is this out of scope for this fix or is this something we should be fixing with this ticket?

Right, so criterion score is currently calibrated to match backend state and not frontend state. We can handle this in a future ticket but it's a bit lower priority for now.

For the 3->4 update, Matt suggested immediately sending an assessment request if no requests are queued or in-flight. This minimizes the response time for single-click assessments. So what's happening is that it basically requests the first 4->3 movement as soon as it happens, and then all the other movements get bundled up into a single request for level 4.

d2l-rubric-criteria-group.js
```diff
@@ -510,6 +519,28 @@ Polymer({
...
+		if (Object.keys(criterionSelectionMap).length > Object.keys(this._criterionSelectionMap).length) {
+			this._criterionSelectionMap = criterionSelectionMap;
+		}
+	},
+
+	_extractCriterionSelectionMap: function(cellAssessmentMap) {
+		return Object.keys(cellAssessmentMap).reduce((selections, href) => ({
+			...selections,
+			...(this.CriterionCellAssessmentHelper.isSelected(cellAssessmentMap[href])
+				? { [href.slice(0, href.lastIndexOf('/'))]: href }
```
mpharoah-d2l, line 539: Are we parsing the url anywhere else? We generally try to avoid that in hypermedia

semibran: Makes sense. This is just supposed to get initial selection state, I'll have to check what you did for that again.

d2l-rubric-criteria-group.js
```diff
@@ -145,7 +145,7 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-criteria-group">
 			.criterion-cell:focus-within {
 				box-shadow: 0 0 0 4px inset rgba(0, 111, 191, 0.4); /* celestine with 0.4 opacity */
 			}
-			.criterion-cell[aria-checked] {
+			.criterion-cell[aria-checked=true] {
```
mpharoah-d2l, line 148: Is adding `=true` necessary here?

semibran: `aria-checked` has to be either true/false for it to register with screen readers. `[aria-checked]` just checks for property presence not value so it'd match both `aria-checked=true` and `aria-checked=false`

semibran:
partially superseded by #1054, will open a new PR once it gets merged

PR #1054, opened 2021-09-08
[20.21.9] US129402 -- Fixed feedback lagging behind cell selection

mpharoah-d2l:
https://rally1.rallydev.com/#/57732444928d/dashboard?detail=%2Fuserstory%2F602446070537


d2l-rubric-criteria-group.js
```diff
@@ -1051,6 +1069,20 @@ Polymer({
...
+			if (entity.class && entity.class.includes('selected')) {
+				this._dispatcher.selectLevel(criterionHref, null);
+				this._optimisticSelections.set(criterionHref, null);
+				this._optimisticSelectionsVersion++;
+			} else {
+				this._dispatcher.selectLevel(criterionHref, levelHref);
+				this._optimisticSelections.set(criterionHref, levelHref);
+				this._optimisticSelectionsVersion++;
+			}
+			this._dispatcher.selectLevel(criterionHref, levelHref);
```
semibran, line 1085: I'm partial towards keeping this out of the queueing logic but I had to redo all of that stuff anyways for the popout view so it might be a bit tough to merge. Another thing is that we can't base selection/deselection off assessed state alone since assessed state gets outdated by every request. That's another discussion though

d2l-rubric-criteria-group.js
```diff
@@ -507,6 +509,18 @@ Polymer({
...
+			type: OptimisticUpdateDispatcher,
+			value: new OptimisticUpdateDispatcher()
+		},
+		_optimisticSelections: {
+			type: Map,
+			value: new Map()
+		},
+		_optimisticSelectionsVersion: {
+			type: Number,
+			value: 0
```
semibran, line 523: where is versioning used?

d2l-rubric-criteria-group.js
```diff
@@ -368,7 +369,7 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-criteria-group">
 								on-keydown="_handleKey"
 								data-href$="[[_getSelfLink(criterionCell)]]"
 							>
-								<d2l-rubric-criterion-cell href="[[_getSelfLink(criterionCell)]]" token="[[token]]" cell-assessment="[[_lookupMap(criterionCell,cellAssessmentMap)]]">
+								<d2l-rubric-criterion-cell href="[[_getSelfLink(criterionCell)]]" token="[[token]]" cell-assessment="[[_lookupMap(criterionCell,cellAssessmentMap)]]" optimistic-update-dispatcher="[[_dispatcher]]">
```
semibran, line 372: seems like this is unused

d2l-rubric-criterion-mobile.js
```diff
@@ -290,7 +301,22 @@ Polymer({
...
 		if (!this.readOnly) {
-			this.CriterionCellAssessmentHelper.selectAsync(entityGetter);
+			this.CriterionCellAssessmentHelper.selectAsync(entityGetter).catch(err => {
+				this.optimisticUpdateDispatcher.rollback(this.href);
+				throw err;
+			});
+		}
+	},
+
+	_selectCellOptimistic: function(entity) {
```
semibran, line 311: I think the inline view already has a bunch of `selected`/`focused`/`hovered` stuff to avoid having to store optimistic updates separately. Is default feedback supposed to show up for hovered levels or just on selection?

semibran (approved):
Yeah I can see how this'll conflict now. I think we converged on a bunch of similar solutions (keeping a copy of local criteria group selection state) but we went about it in different ways. I'm not sure how necessary it is to use `Map` here in general since the keys are just strings, but it might be a bit cleaner semantically so I'll defer to your judgment on that.

d2l-rubric-criteria-group.js
mpharoah-d2l: The actual value is unused, but it's passed in as an unused parameter when we need a function to be recalculated

mpharoah-d2l: To clarify, changing an array or object (unless you change the reference itself) won't trigger an update in computed methods that depend on it, so you can get around that by passing in an unused version variable so the computation is re-run when the version is incremented

d2l-rubric-criterion-mobile.js
mpharoah-d2l: Just on selection. But it was previously relying on the actual entity being returned by the server and didn't have a local state.

d2l-rubric-criteria-group.js
semibran: got it, just polymer things

d2l-rubric-criteria-group.js
mpharoah-d2l: Yeah, it currently is a little buggy if you deselect a cell before the selection goes through. I don't know of a good way to fix this though since we don't have the unselect action available until the selection response comes back

mpharoah-d2l: Actually, because of the entityGetter being evaluated later, it might still work. Will just require an additional state to be storedd

semibran: Right so the way I figured out how to do it without an explicit deselect action is pretty simple, it just needs to cache the latest requested level href per criterion. If the user selects the latest requested level again then they've deselected it

semibran: I removed the entity getters and just used level hrefs since assessment entities get outdated on every selection. Hard to explain but the scoping stuff just adds an extra layer of complexity. Might have to run through the logic for that one more time to make sure we're not missing anything as a result of that though.

mpharoah-d2l: I got this working by checking the current optimistic state. It still works when sending the actual request to the server, because the entityGetter gets the latest state of the entity before running the action, so it gets the unselect action instead of the selection one, and everything syncs up

semibran: It's difficult to explain but without caching requested state the assessment entity still gets outdated, which ultimately turns up a false positive for cell deselection. The test case I've been using is selecting a cell, moving to another cell, then reselecting the first one.

mpharoah-d2l: It works fine in the popout view since it's only relying on local state for the optimistic logic. There seems to be a bug with the local state in the mobile view though

mpharoah-d2l: Oh, nevermind, I see what you mean

mpharoah-d2l: Though that seems like a bug in the promise queue- it _should_ get the updated assessment before each queued action, so it should work anyways, but someone the entity is in an old state I guess, which shouldn't happen.

semibran: Not necessarily, `queueChange` (now `queueCriterionCellAssessment`) only gets called once as soon as the user selects a cell. So any checks that you do directly inside that function outside of like callbacks and stuff are gonna use the assessment entity at that particular moment.

mpharoah-d2l: That's true, but I don't rely on that anymore. Instead the entity is only used when evaluated in the promise queue, at which point it should be consistent with the state of all actions performed beforehand, so it's select action should now be selection again since the previous action unselected it, but for some reason this isn't happening, and I'm not sure why

semibran: did you end up moving the optimistic selection checks? or do you mean you're not using them anymore

mpharoah-d2l: Setting the new optimistic state now only depends on the current assessment if there isn't an optimistic state yet (and since it never gets cleared, that means that this is the initial state when the rubric was loaded). All other times, it looks at the current optimistic state to determine the next state, so it correctly handled deselection regardless of what the current state validated by the server is.

At the same time, the way the promise queue works is that when it's a queued action's time to send, it uses `entityGetter` to refresh the state to what should be the last state confirmed by the server. Since it does this every step, it should always end up with the exact same state as the optimistic state unless an error occurs. But for some reason, that doesn't seem to be working and it's doing the wrong action in the case you described where you select another level then select the initial one again. It seems to think the initial one is still selected, but it shouldn't.

mpharoah-d2l: Oh, I think it's because `entityGetter` is getting replaced in the debouncer logic for some reason

mpharoah-d2l: Yeah, I think `_changesQueued` is trying to duplicate the work of the promise queue instead of letting the queue handle it, so it ends up with diverging states that cause undefined behaviour based on what the current state happens to be

mpharoah-d2l: Do we actually even still need the debouncer? With optimistic updates, it updates instantly anyways. Removing that would fix everything I think

mpharoah-d2l: Yep, looks like that fixes it. Removing the debouncer and relying on the optimistic update map instead completely fixes all synchronization issues 🎉 ...except the checkmark in the popout view still slightly lags behind for some reason.

semibran: I rewrote the queueing logic in my PR to handle a few more cases but I think we'll need more time + some kind of automated test mocking if we want complete case coverage.

It shouldn't be that difficult to get correct outputs without the debouncer so I'm fine with removing it at least just to get these hotfixes in. I agree the debouncer is less useful now that we have optimistic updates but I think there's still a use for it, since it dedupes requests -> shaves down save times significantly for keyboard users on slow connections. It's just a nice to have though and again it'll take a while to get it to work perfectly.

semibran: There's a fix to the checkmark in my PR btw. It was passing in the assessment entity (old state) to determine whether or not to display the checkmark so I just refactored it into a `selected` boolean that uses the selection state (your optimistic update map) like everything else.

semibran:
as an aside/minor nit I think calling everything optimistic selections etc. is a bit backend-centric for naming whats essentially the local state of a frontend component. But we can just fix this during the lit refactor if you want

d2l-rubric-criteria-group.js
mpharoah-d2l: Yeah, you're right- removing the debouncer does potentially put more load on the server, especially for keyboard users; however removing it significantly simplifies the code and fixes desyncs without the need for extra complexity, plus it completely removes the need to make the "outside world" aware of the debouncer to prevent dialogs from being closed in that time, so I would rather proceed with removing it.

mpharoah-d2l: I have everything working now except that the initial state when you load the page in the mobile view is now partially incorrect, and it doesn't get synced with the new state if you open the popout dialog, make some changes, then close it and open back up the mobile view without reloading the page. I'm working on fixing that now

semibran: I'll figure out how to merge in my aria-checked fixes once your stuff is ready to go

mpharoah-d2l: Cool. Sounds good

.eslintrc.json
```diff
@@ -13,6 +13,7 @@
   "globals": {
     "fastdom": true,
     "Promise": true,
-    "Set": true
+    "Set": true,
+	"Map": true
```
semibran, line 17: Mixing tabs and spaces
```suggestion
    "Set": true,
    "Map": true
```

assessment-helpers/criterion-assessment.js
```diff
@@ -74,6 +107,18 @@ class CriterionAssessmentHelper {
 		return entity.properties.score;
 	}
 
+	isOptimisticlySelected(criterionAssessmentEntity, optimisticUpdateMap) {
```
semibran, line 110:
```suggestion
	isOptimisticallySelected(criterionAssessmentEntity, optimisticUpdateMap) {
```

d2l-rubric-criteria-group.js
```diff
@@ -1044,29 +1045,34 @@ Polymer({
...
-		}
 		const entity = entityGetter();
-		if (this._changesRequested[criterionId] === entity) {
-			delete this._changesQueued[criterionId];
+
+		let criterionHref = entity.getLinkByRel(this.HypermediaRels.Rubrics.criterion);
+		let levelHref = entity.getLinkByRel(this.HypermediaRels.Rubrics.level);
+		criterionHref = criterionHref ? criterionHref.href : null;
+		levelHref = levelHref ? levelHref.href : null;
+		if (this.CriterionAssessmentHelper.isOptimisticlySelected(entity, this._optimisticSelections)) {
```
semibran, line 1054:
```suggestion
		if (this.CriterionAssessmentHelper.isOptimisticallySelected(entity, this._optimisticSelections)) {
```

semibran (changes requested):

semibran (approved):

PR #1060, opened 2021-09-10
fix: Read rubric table headers when tabbing - DE45010

imcdonald-d2l:
Improving table accessibility by notifying screen readers what cell they are in when tabbing through rubrics. Screen readers will now read out: 
`Criterion {criterionName}, level {levelName} cell. {cellDescription}.` or `Criterion {criterionName}, criterion score cell. Criterion score input focused.` for each cell when focused. See Gif below with screen reader output.

rubric-cell-accessibility

https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fdefect%2F607911097295


d2l-rubric-criteria-group.js
```diff
@@ -1379,5 +1381,23 @@ Polymer({
 		}
 
 		return this.localize('addFeedbackFor', 'criterionName', criterion.properties.name);
-	}
+	},
+
+	_getDescriptionCellAriaLabel: function(levelName, criterionName, criterionCell) {
+		// hack to announce the position of the description cell within the rubric for individuals using screen reader
+		// should be removed with the lit conversion and the new d2l-table component
```
imcdonald-d2l, line 1388: When we convert to lit we should use the new [d2l-table-wrapper](https://github.com/BrightspaceUI/core/tree/main/components/table#table-wrapper-d2l-table-wrapper) and use default table tags (`<table>`, `<tr>`, `<th>` etc.) with appropriate `scope` attribute for setting column and row headers. See documentation [here](https://www.w3.org/WAI/tutorials/tables/two-headers/). This should fix the problem for us without any `aria-label` hacks.

d2l-rubric-criteria-group.js
```diff
@@ -387,7 +388,8 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-rubric-criteria-group">
 									read-only="[[readOnly]]"
 									editing-score="{{editingScore}}"
 									criterion-num="[[criterionNum]]"
-									criterion-name="[[criterion.properties.name]]">
+									criterion-name="[[criterion.properties.name]]"
+									rubric-cell-description="[[_getOutOfCellAriaLabel(criterion.properties.name)]]">
```
imcdonald-d2l, line 392: Couldn't seem to set `aria-label` for criterion score cell since the input field gets auto-focused. Instead, I had to pass it to `d2l-rubric-editable-score` component and announce it. Not pretty but it gets the job done for now.

semibran (approved):

PR #1070, opened 2021-09-17
chore: move file into src folder

devpow112:
This moves all but the following into the `src` folder.
* `d2l-rubric.js`
* `d2l-rubric-editor.js`
* `d2l-rubric-title.js`


devpow112:
I'll port this to `3.x` as well, not sure if we want/need it in `3.18.x` since `20.21.9` is rolling out now.

mpharoah-d2l (approved):

PR #1083, opened 2021-09-24
test: add d2l-rubric-editor basic vdiffs

devpow112:
Currently there is some minor issues with the generated data. Once that is resolved these will need to be regenerated.


test/perceptual/utilities.js
```diff
@@ -0,0 +1,21 @@
...
+			const element = document.querySelector(selector);
+			const listener = (e) => {
+				if (e.detail.entity.getLinkByRel('self').href !== element.href) {
+					return;
+				}
+
+				element.removeEventListener('d2l-siren-entity-changed', listener);
+
+				requestAnimationFrame(() => {
+					setTimeout(resolve, 2500);
```
devpow112, line 13: This delay is to make sure that the auto saving has a chance to run and complete. Otherwise we will have sporadic failures since it will sometimes show `Saving...` in the diff and sometimes not

README.md
```diff
@@ -139,11 +139,11 @@ To run visual diff tests locally:
 npm install -g mocha@8
 npm install esm puppeteer @brightspace-ui/visual-diff --no-save
 # run visual-diff tests
-mocha './test/**/*.visual-diff.js' -t 10000 --require esm
+mocha './test/**/*.visual-diff.js' -t 40000 --require esm
```
devpow112, line 142: This is to be more inline with what the auto runner for vdiff does. It's timeout is 40000ms

test/perceptual/d2l-rubric-editor.visual-diff.html
```diff
@@ -0,0 +1,101 @@
...
+		</d2l-rubric-editor>
+	</div>
+
+	<div class="container-inner" id="holistic-percentage-published">
+		<d2l-rubric-editor href="data/rubrics/organizations/123659/holistic-percentage-published.json">
+		</d2l-rubric-editor>
+	</div>
+</body>
+
+</html>
```
semibran, line 101: out of curiosity do we have a way to reuse these for demos?

semibran (approved):

test/perceptual/d2l-rubric-editor.visual-diff.html
devpow112: Not yet. I'[m working on that. I need to make all the paths root paths (start with `/`) for that to work. If they are that instead of relative then we could have a single `data` folder with everything in it for demons vdiff and tests if needed.

PR #1085, opened 2021-09-24
US129749 - Remove render path for select-outcomes component

semibran:
Part of Brightspace/lms#14424 - avoids extra conversion of `d2l-select-outcomes` and `d2l-alignment-update` in activity alignments


src/editor/d2l-rubric-criterion-editor.js
```diff
@@ -563,10 +552,6 @@ Polymer({
 			type: Boolean,
 			computed: '_computeIsFlagOn(entity)'
 		},
```
semibran, line 554: Can't really tell what "flag" refers to in this context so just leaving it in for now

imcdonald-d2l (approved):

src/editor/d2l-rubric-criterion-editor.js
imcdonald-d2l: It doesn't even look like this is checking for a flag. It is just checking for this [rel](https://github.com/Brightspace/d2l-hypermedia-constants/blob/master/index.js#L62). 😕 ❓

semibran: Yeah it's weird. I'm guessing that link only gets added hypermedia-side if a certain flag is on, which is one less prop to pass down through the component I guess. Not very self-documenting at all though lol.

PR #1101, opened 2021-09-28
fix: Added callback in order to correctly update pending saves

MitchelPaulin:
We keep track of how many pending saves we have. For specifically the overalls score, the pending saves variable was not being decremented properly due to a missing callback. This caused a guard which checks for pending saves to be zero before allowing level resets to prevent the overall level from updating when moving to another student. 

https://rally1.rallydev.com/#/57732444928ud/custom/373260458992?detail=%2Fdefect%2F609980473945


tdevos-d2l (approved):
lgtm

devpow112:
@MitchelPaulin make sure to get this into the 3.x branch so BSI picks it up.

PR #1102, opened 2021-09-28
test: add criterion group to data

devpow112:
Not to be confused with criterion groups. Requires https://github.com/Brightspace/hm-api-tests/pull/119 to be merged. Diffs are much better. Still working on the id sanitization within the json so that the diffs get even smaller.


devpow112:
> Visual diff tests failed - pull request #1103 has been opened with the updated goldens.

This reverted back to the previous vdiffs now that the data is fixed.

MitchelPaulin (approved):

PR #1109, opened 2021-09-29
Setup v-diff demo pages

imcdonald-d2l:
Adding a hack utility to pass a bogus token to html elements on v-diff pages so they can be used in a demo environment locally.  Token isn't added by default because of work @devpow112 added to `waitForEntityReady` utility.

Note that actions won't work on demo pages but you can interact with elements as you would normally. 

To setup future demo pages make sure you add:

1. `?demo=true` url param on new pages added to `test\perceptual\index.html`. Utility checks for the param so that it doesn't get added when running vdiff tests. 
2. Add the following script to the end of your `<body>` element on your v-diff html file.

```
<script type="module">
		import { addTokenToDemoPageElements } from './utilities.js';
		addTokenToDemoPageElements([array of string html tags you want to add token to]);
</script>
```
rubrics-vdiff-demo-pages


djwester:
Can we add the note in your comment above into the readme for d2l-rubrics so that future us can find that info?

imcdonald-d2l:
> Can we add the note in your comment above into the readme for d2l-rubrics so that future us can find that info?

Just updated the README with the suggested information 🙂

devpow112:
Looks like some dependency added anew version of something with a license we don't like. Looking into it.

devpow112:
The https://github.com/BrightspaceUILabs/media-player repo did a release that added the new version. They don't have license checking in that repo since it's a labs component. Adding exception for it now.

MitchelPaulin (approved):

PR #1111, opened 2021-09-29
De45931 feedback sticking when navigating students

MitchelPaulin:
Information wasn't getting reset between students because the criteria entity doesn't change, only the assessment does. This now also forces an update of the level info when we change assessments. 

https://rally1.rallydev.com/#/57732444928ud/custom/373260458992?detail=%2Fdefect%2F610209725039


devpow112:
For the unit test stuff, it's always the ` d2l-rubric-criterion-editor` that fails. Might be worth disabling those tests for now since they are flaky.

mpharoah-d2l (approved):

PR #1118, opened 2021-09-29
test: add default rubric vdiffs

devpow112:
This involved adding some new data set-up to generate a "default rubric". The commit **test: add default rubric vdiffs (5eceb4c)** is the main stuff that matters. All other stuff was auto generated data.


devpow112:
I have some more ideas about limiting the diff size. For now this will happen any time we add new configurations to the rubric test data collection.

devpow112:
Abandoning due to stupid issues with stupid scroll wrapper.

PR #1122, opened 2021-09-30
[v3] DE44388 -- Update htmleditor

mpharoah-d2l:


devpow112 (approved):

devpow112:
We are still using semantic release (haven't switched it yet) so when you merge make sure to include `fix:` in the commit title for the merge commit

PR #1123, opened 2021-09-30
test: add default rubric vdiffs

devpow112:
This unfortunately caused a change to a bunch of the ids. There will need to be more work to make this a bit better. Right now any time we add stuff to the collection the ids will shift and git can't figure out the JSON spaghetti mess that is hypermedia.


djwester (approved):
2,155 files. I looked at every one, I swear.

devpow112:
> 2,155 files. I looked at every one, I swear.

Lol, yeah. I have some ideas how we came make it better. Hopefully once we get all our cases in place this won't be an issue.

PR #1132, opened 2021-10-01
test: update data

devpow112:
Adjusted so id sanitization is per collection run, not for the entire run of all collections. Means that if an upstream collection changes the downstream ones should be unaffected. Effect of https://github.com/Brightspace/hm-api-tests/pull/130.


devpow112:
Most are just renamed with minor changes. All data is automatically generated.

MitchelPaulin (approved):

PR #1134, opened 2021-10-01
test: update to @web/test-runner

devpow112:
This is more stable, faster, easier to configure and is the recommended way based on open-wc stuff.


test/d2l-rubric-criterion-editor.axe.js
```diff
@@ -1,7 +1,7 @@
 import '../src/editor/d2l-rubric-criterion-editor.js';
 import { expect, fixture, html } from '@open-wc/testing';
 
-describe('d2l-rubric-criterion-editor', () => {
+describe.skip('d2l-rubric-criterion-editor', () => {
```
devpow112, line 4: A slightly newer version of axe doesn't like this anymore.

PR #1150, opened 2021-10-04
Convert `d2l-rubric-title` and test global switch

devpow112:


src/is-lit-flag-enabled.js
```diff
@@ -0,0 +1,3 @@
+export const isLitFlagEnable = () => {
+	return window.D2L?.LP?.Web?.UI?.Flags?.Flag('rubrics-lit', false);
```
devpow112, line 2: This will probably need to be expanded to account for FRA stuff better.

package.json
```diff
@@ -38,12 +38,15 @@
     "d2l-organizations": "github:BrightspaceHypermediaComponents/organizations#semver:^5",
     "d2l-polymer-siren-behaviors": "github:Brightspace/polymer-siren-behaviors#semver:^1",
     "d2l-resize-aware": "github:BrightspaceUI/resize-aware#semver:^1",
+    "d2l-rubric-polymer": "github:Brightspace/d2l-rubric#semver:^3",
```
devpow112, line 41: This is kinda a hack but we can alias an older version of rubrics into itself so we don't need to keep around the polymer code in `master`. We can import it as needed from the older version.

src/entities/rubric-entity.js
```diff
@@ -0,0 +1,9 @@
+import { Entity } from 'siren-sdk/src/es6/Entity';
+
+export class RubricEntity extends Entity {
```
devpow112, line 3: We can add to this entity as we need for pieces of the rubric hypermedia entity.

devpow112:
Requires an additonal step for BSI to work in the short term.
```
# this will eventually have to be added to BSI once we start trying to ship v4
npm i d2l-rubric-polymer@github:Brightspace/d2l-rubric#semver:^3 --no-save
# can be to a v3 or v4 rubrics but needs the above line if v4
npm link d2l-rubric
```

src/is-lit-flag-enabled.js
```diff
@@ -0,0 +1,3 @@
+export const isLitFlagEnable = () => {
+	return window.D2L?.LP?.Web?.UI?.Flags?.Flag('rubrics-lit', false) ?? false;
```
devpow112, line 2: This will probably need to be expanded to account for FRA stuff better. This will work for all BSI stuff in the LMS once the flag is created an exposed to JavaScript.

devpow112:
Once we try to integrate this into BSI this will require updating `d2l-activities`, `d2l-consistent-evaluation` and BSI to `4.x` and also adding `"d2l-rubric-polymer": "Brightspace/d2l-rubric#semver:^3"` in order for it to pass the build and work.

PR #1256, opened 2021-10-28
Try re-enabling sauce labs

devpow112:


devpow112:
Done as part of #1260.

PR #1259, opened 2021-10-29
Wrap component creation (Polymer/`3.x`)

devpow112:
This is to allow us to support the new lit version and be able to not define the polymer components when the lit flag is set. It is pretty hacky but it's kinda the only option left. Each file no longer automatically defines its component. It checks for a specific value first. Only the lit layer defines `window.D2L.Rubric.Lit` so if the polymer based code is not being loaded through the Lit version of the code it will always define all it's elements keeping current compatibility. Once we port over BSI and other FRA's to `4.x` then this will actually become a toggle. The `klass` variable naming is because `class` is a reserved word. I've seen this naming used before to went with it.


mpharoah-d2l:
`prototype` already has a meaning in Javascript.

devpow112:
> `prototype` already has a meaning in Javascript.

So not `prototype`?

mpharoah-d2l:
I'm fine with anything really. Keeping it as klass is fine with me

devpow112:
Fair enough. @semibran if you are fine with `component` I'm gonna change it to that.

semibran:
:+1: Yup that works

semibran (approved):

PR #1260, opened 2021-10-29
Wrap component creation (Lit)

devpow112:


package.json
```diff
@@ -40,7 +40,7 @@
     "d2l-organizations": "github:BrightspaceHypermediaComponents/organizations#semver:^5",
     "d2l-polymer-siren-behaviors": "github:Brightspace/polymer-siren-behaviors#semver:^1",
     "d2l-resize-aware": "github:BrightspaceUI/resize-aware#semver:^1",
-    "d2l-rubric-polymer": "github:Brightspace/d2l-rubric#semver:^3",
+    "d2l-rubric-polymer": "github:Brightspace/d2l-rubric#depowell/wrap-component-creation",
```
devpow112, line 43: Once #1259 merges this line needs to be reverted.

src/helpers/is-lit-flag-enabled.js
```diff
@@ -1,24 +1,15 @@
...
-
-	forcedFraFlagSync = true;
-};
-
-export const isLitFlagEnable = async() => {
+const isLitFlagEnabledInternal = () => {
 	try {
-		await forceFraFlagSync();
-
 		return window?.D2L?.LP?.Web?.UI?.Flags?.Flag('rubrics-lit', false) ?? false;
```
devpow112, line 3: This will need to be expanded to allow for FRA support but we can't use the `ifrau` stuff since it's all async. Looking at appending some option to the FRA `iframe` URL to help with this.

semibran (approved):

PR #1262, opened 2021-10-29
Us132827 description editor conversion

MitchelPaulin:


MitchelPaulin:
The v-diff changed here actually fixed a 2px issue that I introduced in my previous PR.

Most of these problems are getting caused by the not so clear interaction between the rich text flag and the use new editor flag

src/editor/d2l-rubric-description-editor.js
```diff
@@ -1,385 +0,0 @@
...
-				action = description.getActionByName('update');
-			}
-		}
-		return action;
-	},
-
-	_getDescription: function(entity) {
-		const description = entity.getSubEntityByClass(this.HypermediaClasses.rubrics.description);
-		if (description) {
-			return this._richTextAndEditEnabled(this.richTextEnabled, this._canEditDescription) ? description.properties.html : description.properties.text;
```
MitchelPaulin, line 277: This seemed wrong since I would assume we would want to show the html description even if editing was not enabled, changed this to no longer depend on `_canEditDescription` but I can change it back if there was a good reason why we had it this way

src/mixins/rubric-siren-autosave-action-mixin.js
```diff
@@ -0,0 +1,20 @@
+import { dedupeMixin } from '@open-wc/dedupe-mixin';
+import { performSirenAction } from 'siren-sdk/src/es6/SirenAction';
+
+const RubricSirenAutosaveActionMixinInternal = (superclass) => class extends superclass {
+
+	performAutosaveAction(token, action, fields, pendingPropPath) {
```
imcdonald-d2l, line 6: Why was a new mixin file created when it could be added to the existing [auto-saving mixin](https://github.com/Brightspace/d2l-rubric/blob/master/src/mixins/rubric-siren-auto-save-mixin.js)?

src/components/editor/rubric-description-editor.js
```diff
@@ -0,0 +1,363 @@
...
+			<div class="points d2l-body-compact" ?hidden="${!this._showPoints}">
+				<d2l-rubric-autosaving-input
+					id="cell-points"
+					value="${this._points}"
+					@save="${this._savePoints}"
+					?invalid="${this._pointsInvalid}"
+					label="${this.localize('cellPoints')}"
+					?enabled="${this._canEditPoints}"
+					pending-saves="${this._pendingPointsSaves}"
+					?editing="${this._pointsChanging}"
```
imcdonald-d2l, line 151: `value` and `editing` were previously using two-way data binding, how are these property values being updated when changes are made now that we're using lit?

src/mixins/rubric-siren-autosave-action-mixin.js
MitchelPaulin: Must have missed this file, I'll switch over to using this one

src/components/editor/rubric-description-editor.js
MitchelPaulin: `value` as far as I can tell was two way bound to help with optimistic updates. I can get the same effect by immediately setting points in the save event then rolling back to the old value if anything goes wrong.

For the `editing` property, I am not sure why that used two way binding, its set to readOnly so I am not sure how it could ever change and need to notify the parent.

imcdonald-d2l: The corresponding editing value is used in the parent [here](https://github.com/Brightspace/d2l-rubric/pull/1262/files#diff-1755e604cd26b52bfaad7ea4f42137b0dbbb3327ac57151b36941f54976aee55L377-L382). It uses the `_pointsChanging` property to make sure they are not updating the points value while the user is still editing. Looks like it might be necessary.

src/components/editor/rubric-description-editor.js
```diff
@@ -0,0 +1,363 @@
...
+		this._pendingDescriptionSaves = 0;
+		this._pointsInvalid = false;
+		this._pointsInvalidError = null;
+		this._pendingPointsSaves = 0;
+		this.updatingLevels = false;
+	}
+
+	render() {
+		return html`
+			<div class="points d2l-body-compact" ?hidden="${!this._showPoints}">
```
imcdonald-d2l, line 140: We should stay consistent with the newer components and just not render this div if `!this._showPoints` and remove the `hidden` styles.

src/mixins/rubric-siren-auto-save-mixin.js
```diff
@@ -32,6 +32,18 @@ const RubricSirenAutoSaveMixinInternal = (superclass) => class extends RubricSir
 		}
 	}
 
+	performAutoSaveActionWithAttribute(action, fields, attr) {
```
devpow112, line 35: For this can you just use the `_pendingSaves` property that this mixin provides. The idea was to centralize on that variable for pending saves since before we used 2 way bindings in polymer which don't exist in lit.

src/components/editor/rubric-description-editor.js
```diff
@@ -0,0 +1,363 @@
...
+			this._updateDescription(this.entity);
+		}
+	}
+
+	get _canEditPoints() {
+		return this.entity?.hasActionByName('update-points') && !this.updatingLevels && !this.partialEditMode;
+	}
+
+	get _pointsRequired() {
+		const action = this?.entity.getActionByName('update-points');
```
imcdonald-d2l, line 191: Why is the `?` after `this` in this particular case? Also, you could move `this.entity.getActionByName('update-points')` into the entity definition file.

src/components/editor/rubric-description-editor.js
```diff
@@ -0,0 +1,363 @@
...
+		}
+		return field.hasClass('required');
+	}
+
+	get _showPoints() {
+		return this.entity?.hasClass(Classes.rubrics.overridden);
+	}
+
+	_arePointsRequired(entity) {
+		const action = entity?.getActionByName('update-points');
```
imcdonald-d2l, line 207: Should be `this.entity` here.

src/components/editor/rubric-description-editor.js
```diff
@@ -0,0 +1,363 @@
...
+				if (entity.hasLinkByRel(rel)) {
+					constructed += `${entity.getLinkByRel(rel).href}|`;
+				}
+			});
+		}
+		return constructed;
+	}
+
+	_getAriaLabel(criterionName) {
+		const levelName = this.entity?.properties?.levelName || this.entity?.properties?.name;
```
imcdonald-d2l, line 231: This logic to get the level name should be moved into an entity definition file.

src/mixins/rubric-siren-auto-save-mixin.js
MitchelPaulin: The specific component has two pending saves variable, one for the description and one for the points, I think if I only used the one variable I would be crossing wires

devpow112: Ah fair enough. Missed that. Damn that's so weird. I wonder if we really need 2 variables. Maybe an investigation for later though

src/components/editor/rubric-description-editor.js
MitchelPaulin: Typo, meant to have it after the `entity`

src/mixins/rubric-siren-auto-save-mixin.js
imcdonald-d2l: Could we provide a pending saves map instead? So `this._pendingSaves.get('propertyName')`?

src/components/editor/rubric-description-editor.js
MitchelPaulin: You're right, changed it to a `get` type function and forgot to delete the original

src/mixins/rubric-siren-auto-save-mixin.js
MitchelPaulin: Maybe, but adding a map to save on one property definition isn't worth the added complexity imo.

imcdonald-d2l: I'll have two pending save variables when I convert `d2l-rubric-level-editor` so having a pending save map could be useful for other components.

src/components/editor/rubric-description-editor.js
```diff
@@ -0,0 +1,361 @@
...
+				if (entity.hasLinkByRel(rel)) {
+					constructed += `${entity.getLinkByRel(rel).href}|`;
+				}
+			});
+		}
+		return constructed;
+	}
+
+	_getAriaLabel(criterionName) {
+		return this.localize(this.ariaLabelLangTerm, 'criterionName', criterionName, 'levelName', this.levelName);
```
imcdonald-d2l, line 207: Does calling `this.levelName` work? I thought you had to call `this.entity.levelName` to access the entity [levelName function](https://github.com/Brightspace/d2l-rubric/blob/master/src/entities/rubric-criterion-cell-entity.js#L29)

devpow112 (approved):

PR #1264, opened 2021-10-29
US132990 - Conversion primer for criterion-editor

semibran:
[Whitespace diff off](https://github.com/Brightspace/d2l-rubric/pull/1264/files?diff=split&w=1)

Partially resolves two-way data binding by adding event handling for the alignment tags component. Mainly for activity-alignments testing - integration with the AA outcomes selector just doesn't work in Lit without receiving events. I'll be working on the rest of the conversion for this component shortly anyways but the plan is just to get the integration point working as soon as possible so we can test AA once FRA flagging gets sorted out.

https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F612451191399


src/editor/d2l-rubric-criterion-editor.js
```diff
@@ -527,17 +527,18 @@ const component = {
...
+		_hideOutcomes: {
 			type: Boolean,
 			value: true,
 			computed: '_canHideOutcomes(isHolistic, _isAlignmentTagListEmpty, partialEditMode)',
 		},
-		_isAlignmentTagListEmpty:{
-			type:Boolean,
+		_isAlignmentTagListEmpty: {
+			type: Boolean,
+			value: true
```
semibran, line 541: Needs to default to true from here since the component isn't getting the default value from the criterion editor (displays an impossible state otherwise)

src/editor/d2l-rubric-criterion-editor.js
```diff
@@ -239,7 +239,7 @@ const template = `<dom-module id="d2l-rubric-criterion-editor">
 			}
 
 			.select-outcomes-hierarchical {
-				overflow-y: auto;
+				overflow-y: hidden;
```
semibran, line 242: This is basically the only change that could potentially cause problems.... I'm not sure why it was there to begin with. In the lit version it makes double scrollbars show up in the dialog. The way the height for the outcomes selector is set is really janky anyhow so I'll have to see experiment and see there's a better workaround.

MitchelPaulin (approved):

PR #1265, opened 2021-11-01
(3.x) Fix print button screen reader issue - DE45450

imcdonald-d2l:
Description attribute was incorrectly set.
 
https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fdefect%2F610438225455


lang/en.js
```diff
@@ -155,7 +155,6 @@ export default {
 	"pointsSaveFailed": "Saving points failed",
 	"preview": "Preview",
 	"printRubric": "Print Rubric",
-	"printRubricDescription": "Print Rubric. A print preview will be opened.",
```
imcdonald-d2l, line 158: This wasn't used anywhere and doesn't accurately describe the button so removing it. Preview doesn't open in all browsers depending on your settings.

semibran (approved):

PR #1268, opened 2021-11-01
Switch to external tool and fix image issue (`3.x`)

devpow112:
Mainly the same as #1267 just for `3.x`. Also removed `package-lock.json` and caching around it. This was already done in `master`.


semibran:
was the caching causing errors on the polymer branch or something? Or are we removing it just to be consistent

devpow112:
> was the caching causing errors on the polymer branch or something? Or are we removing it just to be consistent

The issue in `src/d2l-rubric-competencies-icon.js` was in both branches.

semibran (approved):

PR #1272, opened 2021-11-01
Convert d2l-rubric-feedback to Lit

mpharoah-d2l:
~~Current has a bug where feedback doesn't update change changing levels. This bug is also present in the current Polymer version in master. Looking into that.~~ Fixed


src/components/rubric-feedback.js
```diff
@@ -0,0 +1,598 @@
...
+
+	connectedCallback() {
+		super.connectedCallback();
+		this.addEventListener('focus', this._focus);
+
+		/* Once d2l-rubric-criteria-group and d2l-rubric-criteria-mobile have been
+		 * converted to Lit, this can be removed and we can pass it through directly
+		 * in the HTML template
+		 */
+		let node = this.ahadowRoot;
```
mpharoah-d2l, line 227: Huh, just realized the type here, but it seems to be working anyways??? Maybe you actually _can_ pass object references between Polymer and Lit. Need to investigate if it's actually working or if only non-function properties are getting passed through, which might actually fix the feedback bug?

mpharoah-d2l: Looks like it does work because it's not passed through the attribute but is set directly as a property

src/components/rubric-feedback.js
```diff
@@ -0,0 +1,576 @@
...
+				value: false,
+				attribute: 'focus-feedback-html-editor',
+				reflect: true
+			},
+
+			criterionEntity: { attribute: false },
+			criterionAssessment: { attribute: false },
+			optimisticUpdateDispatcher: { type: Object },
+			optimisticLevelHref: { attribute: false },
+			addingFeedback: { attribute: false },
```
semibran, line 46: are these private or public? I was under the impression that we'd make sure `_`-prefixing and `attribute: false` were 1:1 and vice versa

src/components/rubric-feedback.js
```diff
@@ -0,0 +1,576 @@
...
+import { rubricStyles } from '../styles/rubric-styles.js';
+import { tryGetSirenErrorMessage } from '../helpers/error-normalizer.js';
+
+class RubricFeedback extends AssessmentMixin(EntityWatcherMixin(LocalizeMixin(LitElement))) {
+
+	static get properties() {
+		return {
+			criterionHref: { type: String, attribute: 'criterion-href' },
+			criterionAssessmentHref: { type: String, attribute: 'criterion-assessment-href' },
+			token: { type: String, attribute: 'token' },
```
semibran, line 17: the feedback component doesn't have an HM entity?

src/components/rubric-feedback.js
```diff
@@ -0,0 +1,576 @@
...
+				label="${this.localize('criterionFeedback')}${this._feedbackHeadingExtra || ''}"
+				class="rubric-cell-editor"
+				type="inline"
+				html=${this._feedback}
+				?no-border="${!this.compact}"
+				escape-scroll-areas="true"
+				height="100px"
+				@focusin="${this._focusEditor}"
+				@d2l-htmleditor-blur="${this._blurHandler}"
+				@feeback-changed="${this._feedbackChanged}">
```
semibran, line 255:
```suggestion
				@feedback-changed="${this._feedbackChanged}">
```

src/components/rubric-feedback.js
```diff
@@ -0,0 +1,576 @@
...
+				?data-mobile="${this.compact}"
+			>
+				<div class="feedback-heading d2l-label-text">
+					${this.localize('criterionFeedback')}
+				</div>
+				<div class="feedback-text d2l-body-compact">
+					<s-html
+						style="white-space: pre-line;"
+						html="${feedbackHtml}"
+					></s-html>
```
semibran, line 312: Not sure if we wanted to switch over to `d2l-html-block`. I forget if there were pros and cons, I was sure that something stopped working in lit. Might've been MathML/latex related but would have to test.

src/components/rubric-feedback.js
```diff
@@ -0,0 +1,576 @@
...
+		}
+
+		this._disabled = true;
+		this._saveFeedback('').finally(() => {
+			if (!this._feedbackInvalid) {
+				this._disabled = false;
+				this.dispatchEvent(new CustomEvent('close-feedback', {
+					bubbles: true,
+					composed: true,
+				}));
```
semibran, line 444: This'll bubble outside of the shadow root right? We might to establish some guidelines around prefixing lint events. I feel like it'd be better to increase the name specificity a bit to something like `d2l-rubric-feedback-closed` - correct me if I'm missing something though

mpharoah-d2l: That would probably be better, but we were already firing the event this way before, and I'm not sure if other things outside of d2l-rubric depend on it or not.
https://github.com/Brightspace/d2l-rubric/blob/3.x/src/d2l-rubric-feedback.js#L496

semibran: Yeah a quick [code search](https://search.d2l.dev/search?project=Brightspace&project=BrightspaceHypermediaComponents&project=BrightspaceUI&project=BrightspaceUILabs&project=lms&full=%22close-feedback%22&defs=&refs=&path=&hist=&type=&xrd=&nn=5&si=full&si=full&n=25&start=25) says it's only used internally. Up to you whether you want to keep this and the rest of the events as is or not. Alternatively we might be able to get away with not flipping the bubbles/composed flags since it looks like they're only being used by direct parents anyway.

src/components/rubric-feedback.js
mpharoah-d2l: In some cases, we have public non-attribute properties (like `optimisticUpdateDispatcher`). The rest though are actually private I think. I just copied the named from the Polymer version, but I can update them

src/components/rubric-feedback.js
mpharoah-d2l: I'll update it to have a more specific name then.

src/components/rubric-feedback.js
mpharoah-d2l: Yeah, I just kept it as s-html to be safe. Too much potential for regressions with HTML stuff like this

src/components/rubric-feedback.js
mpharoah-d2l: I recently discovered that Javascript now natively supports private member variables with the `#` prefix. (I doubt that would work here though since it's probably used by the Lit framework in a technically non-private way. But that's cool)

src/components/rubric-feedback.js
```diff
@@ -0,0 +1,574 @@
...
+	_clearFeedbackHandler() {
+		if (this._disabled) {
+			return;
+		}
+
+		this._disabled = true;
+		this._saveFeedback('').finally(() => {
+			if (!this.feedbackInvalid) {
+				this._disabled = false;
+				this.dispatchEvent(new CustomEvent('close-feedback', { bubbles: false }));
```
semibran, line 447: All event init params default to `false`
```suggestion
				this.dispatchEvent(new CustomEvent('close-feedback'));
```

src/components/rubric-feedback.js
```diff
@@ -0,0 +1,574 @@
...
+	async _saveFeedback(value) {
+		this._feedbackModified = false;
+		this.feedbackInvalid = false;
+		this._pendingFeedbackSaves++;
+		this.dispatchEvent(new CustomEvent('save-feedback-start', {
+			detail: {
+				hasPendingSaves: true,
+			},
+			bubbles: true,
+			composed: true,
```
semibran, line 539: `save-feedback-start` is [internal-only](https://search.d2l.dev/search?project=Brightspace&project=BrightspaceHypermediaComponents&project=BrightspaceUI&project=BrightspaceUILabs&project=lms&full=%22save-feedback-start%22&defs=&refs=&path=&hist=&type=&xrd=&nn=5&si=full&si=full) - can remove bubbling

src/components/rubric-feedback.js
```diff
@@ -0,0 +1,574 @@
...
+		}
+
+		this._pendingFeedbackSaves--;
+		this.dispatchEvent(new CustomEvent('save-feedback-finished', {
+			detail: {
+				success: !this.feedbackInvalid,
+				hasPendingSaves: this._pendingFeedbackSaves > 0
+			},
+			bubbles: true,
+			composed: true,
```
semibran, line 561: `save-feedback-finished` is [internal-only](https://search.d2l.dev/search?project=Brightspace&project=BrightspaceHypermediaComponents&project=BrightspaceUI&project=BrightspaceUILabs&project=lms&full=%22save-feedback-finished%22&defs=&refs=&path=&hist=&type=&xrd=&nn=5&si=full&si=full) - can remove bubbling

semibran (approved):

PR #1276, opened 2021-11-02
Add top level structure editor file

devpow112:
This component is directly referenced in the lms [here](https://github.com/Brightspace/lms/blob/master/platformTools/rubrics/D2L.PlatformTools.Rubrics/Views/RubricEdit.Renderer.cs#L117). Want to make it a top level "exposed" component so it's consistent and deals with flagging the same as the other top level components. This will require lms and BSI changes as well.


devpow112:
I'll do this to `master` as well once this is merged.

imcdonald-d2l (approved):

PR #1281, opened 2021-11-02
Add FRA related rubric check

devpow112:
Pairs with https://github.com/Brightspace/lms/pull/15689. This looks at query parameters for rurbic lit flag instead of trying to use the async stuff provided by the ifrau. It falls back to checking the feature flags that get set-up by the ifrau workflow if the previous checks fail.


src/helpers/is-lit-flag-enabled.js
```diff
@@ -1,6 +1,33 @@
+const flagName = 'rubrics-lit';
+
+const isFramed = () => {
```
devpow112, line 3: This is mostly from https://github.com/BrightspaceUI/core/blob/main/helpers/framed.js but can't take the async parts. Should still work for iframe FRAs in terms of checking.

semibran (approved):

PR #1282, opened 2021-11-03
`d2l-rubric-level-editor` conversion - US132992

imcdonald-d2l:
https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F612451706715


src/components/editor/rubric-description-editor.js
```diff
@@ -310,9 +311,9 @@ class RubricDescriptionEditor extends RubricSirenAutoSaveMixin(LocalizeMixin(Ent
 					composed: true,
 				}));
 				this._updateDescription(this.entity);
-			}).catch(() => {
+			}).catch((err) => {
 				this._descriptionInvalid = true;
-				this._descriptionInvalidError = this.localize('descriptionSaveFailed');
+				this._descriptionInvalidError = tryGetSirenErrorMessage(err) || this.localize('descriptionSaveFailed');
```
imcdonald-d2l, line 316: @MitchelPaulin just added this to my PR since you missed it in the description editor conversion.

devpow112:
> > Visual diff tests failed - pull request #1283 has been opened with the updated goldens.
> 
> These visual diff files show a difference in the last level points field being editable in the converted component. Looks to be disabled in the old component but not sure how looking at the logic. It is disabled based off of this logic `this._entity?.hasActionByName('update-points')` so should be editable based off my understanding 🤔

It seems reasonable to me that you can't update the `0%` field. I wonder if that was just a bug before?

imcdonald-d2l:
Any ideas on why the following error is happening in unit tests when I didn't touch the `criteria-group-editor`?
```
d2l-rubric-criteria-group-editor > smoke test > edit group name > saves group name
      Error: Uncaught TypeError: this.performAutosaveAction is not a function
```

imcdonald-d2l:
> > > Visual diff tests failed - pull request #1283 has been opened with the updated goldens.
> > 
> > 
> > These visual diff files show a difference in the last level points field being editable in the converted component. Looks to be disabled in the old component but not sure how looking at the logic. It is disabled based off of this logic `this._entity?.hasActionByName('update-points')` so should be editable based off my understanding 🤔
> 
> It seems reasonable to me that you can't update the `0%` field. I wonder if that was just a bug before?

@devpow112 it was actually a bug with the auto-save input field. The default `enabled` value was set to `true` so it was never updating to `false` when `enabled` wasn't set. I fixed it in this [commit](https://github.com/Brightspace/d2l-rubric/pull/1282/commits/e7e59ebdd1035a69624e58b1c9d72b1813a3e4d8). Hopefully it doesn't break anything else 🤞🏻

src/components/editor/rubric-description-editor.js
MitchelPaulin: Caught this during my current conversion as well, thanks

src/components/editor/rubric-level-editor.js
```diff
@@ -0,0 +1,435 @@
...
+import { tryGetSirenErrorMessage } from '../../helpers/error-normalizer.js';
+import { updateEntity } from 'siren-sdk/src/es6/EntityFactory.js';
+
+class RubricLevelEditor extends RubricSirenAutoSaveMixin(LocalizeMixin(EntityMixin(LitElement))) {
+
+	static get properties() {
+		return {
+			hasOutOf: { type: Boolean },
+			percentageFormatAlternate: { type: Boolean },
+			partialEditMode: { type: Boolean },
```
MitchelPaulin, line 23: Do you need to specify `attribute: 'partial-edit-mode'` here?

src/components/editor/rubric-level-editor.js
```diff
@@ -0,0 +1,435 @@
...
+		this._pointsRequired = entity.arePointsRequired;
+		this._canDelete = entity.canDelete;
+		this._showPoints = this._computeShowPoints(this.hasOutOf, entity);
+		this._updateName(entity);
+		this._updatePoints(entity);
+
+		if (!this._canEditPoints) {
+			this._pointsInvalid = false;
+			this._pointsInvalidError = null;
+		} else if (!this._canEditName) {
```
MitchelPaulin, line 236: Should this be an `else if`? Seems like these should be two separate `if` statements.

src/components/editor/rubric-level-editor.js
imcdonald-d2l: It seemed to work without it just fine but I'll add it to be safe 👍🏻

src/components/editor/rubric-level-editor.js
imcdonald-d2l: Yup, should be separate `if` statements. Good catch 👍🏻

src/components/editor/rubric-level-editor.js
```diff
@@ -0,0 +1,435 @@
...
+				.points div {
+					margin-left: 0.5rem;
+				}
+
+				.points[alt-percent-format] div {
+					margin-left: 0;
+					margin-right: 0.5rem;
+				}
+
+				:dir(rtl) .points div {
```
MitchelPaulin, line 86: Does this selector work for you? For me to get rtl capability to work I had to use `:host-context([dir='rtl'])`

imcdonald-d2l: It probably doesn't. I didn't test `rtl` behavior so I'll update that

MitchelPaulin (approved):

PR #1285, opened 2021-11-03
Get level name as part of entity update

devpow112:
Recommend viewing with [white-space diff off](https://github.com/Brightspace/d2l-rubric/pull/1285/files?w=1).


src/components/editor/rubric-description-editor.js
```diff
@@ -233,18 +237,21 @@ class RubricDescriptionEditor extends RubricSirenAutoSaveMixin(LocalizeMixin(Ent
 	}
 
 	_onEntityChanged(entity, oldEntity) {
-		if (entity) {
```
devpow112, line 236: This didn't diff well but this callback isn't called if the entity is `null` so we don't need to check entity nullity here.
https://github.com/Brightspace/d2l-rubric/blob/master/src/mixins/entity-mixin.js#L13

src/components/editor/rubric-description-editor.js
```diff
@@ -233,18 +237,21 @@ class RubricDescriptionEditor extends RubricSirenAutoSaveMixin(LocalizeMixin(Ent
...
+
+		const pointsChanged = oldEntity ? entity.properties.points !== oldEntity.properties.points : true;
+
+		if (pointsChanged) {
+			this._updatePoints(entity);
+		}
+
+		// key needs to be updated after description so that the html-editor uses the updated value when its observer runs
+		this._key = this._constructKey(this.keyLinkRels, entity);
+		this._levelName = entity.levelName;
```
devpow112, line 254: We should really try and grab all the entity values we need in `_onEntityChanged` and store them as properties. This includes actions as well. That way if the entity changes it will trigger a re-render.

imcdonald-d2l (approved):

PR #1270, opened 2021-11-01
Fix rubrics not assessible - DE45869

imcdonald-d2l:
Also fixing the following error: 
image

Rally defect: https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fdefect%2F613372307005

PR #1287, opened 2021-11-03
Fix styles and pop-out selection in chrome/ff/safari - DE45869

imcdonald-d2l:
Change that broke safari/firefox [here](https://github.com/Brightspace/d2l-rubric/pull/1270/files#diff-42691203dbbf2e3d9604d29e6246fba62d792e7a3d6e3befdbba72415f4c31b0R220). Used a hack that was suggested [here](https://stackoverflow.com/questions/3215553/make-a-div-fill-an-entire-table-cell#comment119339531_34781198) but Firefox and Safari don't like it. To fix the issue I'm making the hack worse by applying it to chrome only so all browsers work as expected. Hopefully this isn't an issue when we convert to the new `d2l-table`. 🤞🏻 

css-irl


src/d2l-rubric-criteria-group.js
```diff
@@ -217,7 +217,14 @@ const template = `<dom-module id="d2l-rubric-criteria-group">
 			}
 
 			d2l-table[type="default"] d2l-td {
-				height: 1px; /* HACK: enables "height: 100%" to work on children */
+				height: 100%;
+			}
+
+			/* Chromium Only - HACK */
+			@supports (contain: paint) and (not (-moz-appearance: none)) {
```
imcdonald-d2l, line 224: Using selector from article I found [here](https://www.bram.us/2021/06/23/css-at-supports-rules-to-target-only-firefox-safari-chromium/#chromium). If anyone has any better suggestions I'm all ears. 👂🏻

semibran (approved):

PR #1289, opened 2021-11-04
Us132994 ovearll level editor conversion

MitchelPaulin:


src/components/editor/rubric-description-editor.js
```diff
@@ -319,9 +320,9 @@ class RubricDescriptionEditor extends RubricSirenAutoSaveMixin(LocalizeMixin(Ent
 					composed: true,
 				}));
 				this._updateDescription(this.entity);
-			}).catch(() => {
+			}).catch((err) => {
 				this._descriptionInvalid = true;
-				this._descriptionInvalidError = this.localize('descriptionSaveFailed');
+				this._descriptionInvalidError = tryGetSirenErrorMessage(err) || this.localize('descriptionSaveFailed');
```
MitchelPaulin, line 325: Missed this when translating the description editor but @imcdonald-d2l beat me to it :)

src/components/editor/rubric-overall-level-editor.js
```diff
@@ -0,0 +1,347 @@
...
+	_saveName(saveEvent) {
+		const action = this.entity?.getActionByName('update-name');
+		this._nameChanging = false;
+		if (action) {
+			const old_name = this._overallLevelName;
+			this._overallLevelName = saveEvent.value;
+			if (this._nameRequired && !saveEvent.value.trim()) {
+				this._nameInvalid = true;
+				this._nameInvalidError = this.localize('nameIsRequired');
+				this._overallLevelName = old_name;
```
imcdonald-d2l, line 270: I don't think we should be reverting to the `old_name` on failures (same with saving range). The user is presented with visual queues that the value is invalid when saving fails and that should prompt them to fix the value themselves. If we revert the value, the field then becomes valid but the error is still present in the UI. (correct me if this is wrong) It's not a good user-experience.

src/entities/rubric-overall-level-entity.js
```diff
@@ -1,6 +1,46 @@
...
+			return false;
+		}
+		const field = action.getFieldByName('rangeStart');
+		if (!field) {
+			return false;
+		}
+		return field.hasClass('required');
+	}
+	get levelName() {
+		return this._entity?.properties?.name;
```
imcdonald-d2l, line 39: `this._entity?.properties` can be swapped for just `this.properties` (it's accessible from `EntityBase`)

src/components/editor/rubric-overall-level-editor.js
MitchelPaulin: Hmm, I think you're right, I'll remove this.

src/components/editor/rubric-overall-level-editor.js
```diff
@@ -0,0 +1,347 @@
...
+			value="${this._overallLevelName}"
+			@save="${this._saveName}"
+			@input="${this.nameIsChanging}"
+			?invalid="${this._nameInvalid}"
+			label="${this.localize('overallLevelName')}"
+			?enabled="${this.entity?.canEditName}"
+			pending-saves="${this._pendingNameSaves}"
+		></d2l-rubric-autosaving-input>
+		<div class="operations" text-only="${!this.entity?.hasRangeStart}">
+			<div class="points" ?hidden="${!this.entity?.hasRangeStart}">
```
imcdonald-d2l, line 108: Should remove `hidden` property to stay consistent by conditionally rendering components.

src/components/editor/rubric-overall-level-editor.js
```diff
@@ -0,0 +1,347 @@
...
+			if (selfLinkChanged || rangeStartChanged) {
+				this._updateRangeStart(entity, selfLinkChanged);
+			}
+		}
+	}
+	_rangeStartIsChanging() {
+		this._rangeStartChanging = true;
+	}
+	_renderDeleteButton() {
+		if (this.partialEditMode || !this.entity?.canDeleteLevel) {
```
imcdonald-d2l, line 201: `this.entity?.canDeleteLevel` should be its own property set in `_onEntityChanged` so re-renders happen when this changes. `this.entity?.canDeleteLevel` should also be moved into the entity definition file a `get` function. See @devpow112  comment on this topic [here](https://github.com/Brightspace/d2l-rubric/pull/1285#discussion_r742221137).

src/components/editor/rubric-overall-level-editor.js
```diff
@@ -0,0 +1,347 @@
...
+		<d2l-dialog-confirm id="deleteOverallLevelDialog"
+			title-text="${this.localize('deleteLevelConfirmationTitle')}"
+			text="${this._getDeleteLevelConfirmText()}">
+			<d2l-button slot="footer" primary data-dialog-action="delete">
+				${this.localize('deleteConfirmationYes')}
+			</d2l-button>
+			<d2l-button slot="footer" data-dialog-action="cancel">
+				${this.localize('deleteConfirmationNo')}
+			</d2l-button>
+  		</d2l-dialog-confirm>
```
imcdonald-d2l, line 95: nit: tab indent of `</d2l-dialog-confirm>`. This whole dialog should be conditionally rendered on if they have delete permissions as well so we don't render it unnecessarily. Did something similar in my `level-editor` conversion.

src/components/editor/rubric-overall-level-editor.js
imcdonald-d2l: @devpow112 do you mind clarifying on if `entity` changes won't always trigger a re-render? I assume if we set properties for these type of items we'll get better `lit` re-rendering as it will know what specific parts of the `dom` need to be updated.

MitchelPaulin: I was wondering this as well, would it be best just to pass all the fields as function arguments even if its a bit redundant just to ensure render gets called when it should

src/components/editor/rubric-overall-level-editor.js
```diff
@@ -0,0 +1,363 @@
...
+			const selfLinkChanged = entity?.getSelfLink() !== oldEntity?.getSelfLink();
+			const nameChanged = oldEntity ? entity.levelName !== oldEntity.levelName : true;
+			if (selfLinkChanged || nameChanged) {
+				this._updateName(entity, selfLinkChanged);
+			}
+			const rangeStartChanged = oldEntity ? entity.rangeStart !== oldEntity.rangeStart : true;
+			if (selfLinkChanged || rangeStartChanged) {
+				this._updateRangeStart(entity, selfLinkChanged);
+			}
+		}
```
imcdonald-d2l, line 183: Don't need the `if (entity)` here since it is checked in the `EntityMixin` before making it to this function. Also, don't think you need to check `oldEntity` in here. It would be cleaner to just check in the  `_updateName` function if `this._overallLevelName !== entity.levelName` and update accordingly. (same thing for the `rangeStartChanged`. Shouldn't be checking the `oldEntity` but the actual component properties to see if they've changed.

imcdonald-d2l (approved):

PR #1295, opened 2021-11-05
US132990 - Convert d2l-rubric-criterion-editor to LitElement

semibran:
https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F612451191399


src/entities/rubric-criterion-cell-entity.js
```diff
@@ -26,6 +26,10 @@ export class RubricCriterionCellEntity extends EntityBase {
 			(!!this.properties.points || this.properties.points === 0);
 	}
 
+	get href() {
```
devpow112, line 29: I think the entity base already has this
https://github.com/Brightspace/d2l-rubric/blob/master/src/entities/entity-base.js#L129

src/helpers/compose-mixins.js
```diff
@@ -0,0 +1,2 @@
+export const composeMixins = (base, ...mixins) =>
+	mixins.reduce((subclass, superclass) => superclass(subclass), base);
```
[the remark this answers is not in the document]
devpow112: This might be a good thing to see if we can get it in BrightspaceUI/core. Could be useful elsewhere

src/components/editor/rubric-description-editor.js
```diff
@@ -218,7 +221,9 @@ class RubricDescriptionEditor extends RubricSirenAutoSaveMixin(LocalizeMixin(Ent
 	_getDescription(entity) {
 		const description = entity?.getSubEntityByClass(Classes.rubrics.description);
 		if (description) {
-			return this.useNewBrightspaceEditor ? description.properties.html : description.properties.text;
+			return this.richTextEnabled && this._canEditDescription
+				? description.properties.html
+				: description.properties.text;
```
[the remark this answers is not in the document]
MitchelPaulin: I originally swapped this out because I thought it was odd that whether or not you can edit the description determined whether we displayed plaintext and html. Guess there was a reason for it.

At any rate I think we need to also check `useNewBrightspaceEditor` somewhere here. If rich-text is off and brightspace editor is on I'm not sure if this would behave how we would like it to.

src/components/editor/rubric-description-editor.js
```diff
@@ -173,14 +174,16 @@ class RubricDescriptionEditor extends RubricSirenAutoSaveMixin(LocalizeMixin(Ent
...
-		return description && (description.hasActionByName('update-description') || description.hasActionByName('update')) && !this.updatingLevels;
-	}
-
-	set _canEditDescription(value) {
-		this._canEditDescription = value;
-		if (this.entity && this.richTextEnabled) {
-			this._updateDescription(this.entity);
+		const canEditDescription = (description?.hasActionByName('update-description') || description?.hasActionByName('update'))
+			&& !this.updatingLevels
+			&& !this.disabled;
```
MitchelPaulin, line 179: Not sure about this, I think we should try to make getters and setters pure functions if we can and check the disabled and updatingLevels fields elsewhere

src/components/editor/rubric-criterion-editor.js
```diff
@@ -0,0 +1,802 @@
...
+				bubbles: true,
+				composed: true
+			}));
+			this.requestUpdate('criterionDetailWidth', oldWidth);
+		}
+	}
+
+	render() {
+		return html`
+			<div class="gutter -left${this.isHolistic ? ' -holistic' : ''}">
```
MitchelPaulin, line 329:
```suggestion
			<div class="gutter-left${this.isHolistic ? ' -holistic' : ''}">
```

src/components/editor/rubric-criterion-editor.js
```diff
@@ -0,0 +1,802 @@
...
+				${!this.disabled && !this.partialEditMode ? html`
+					<slot name="gutter-left"></slot>
+				` : html``}
```
MitchelPaulin, line 332: nit: I think the preferred approach we are going with is just returning null over empty html (might be faster as well?)

src/components/editor/rubric-criterion-editor.js
```diff
@@ -0,0 +1,802 @@
...
+				d2l-rubric-feedback-editor {
+					width: 100%;
+				}
+
+				.criterion-tags {
+					display: flex;
+					align-items: center;
+					box-sizing: border-box;
+					border-right: thin solid var(--d2l-color-galena);
+					margin-right: var(--d2l-rubric-editor-start-gutter-width);
```
imcdonald-d2l, line 163: I believe you need to import [these styles](https://github.com/Brightspace/d2l-rubric/blob/master/src/styles/rubric-editor-cell-styles.js#L6) to get this defined variable. Noticed it isn't defined in this file or the other current style imports.

src/components/editor/rubric-criterion-editor.js
```diff
@@ -0,0 +1,802 @@
...
+								id="input-name"
+								max-rows="-1"
+								?disabled="${!this._canEdit}"
+								value="${this._name}"
+								label="${this.localize('criterionNameAriaLabel', 'positionNumber', this.positionNumber)}"
+								label-hidden
+								aria-invalid="${this._nameError ? 'true' : 'false'}"
+								@input="${this._handleNameInput}"
+							></d2l-input-textarea>
+							${!this.partialEditMode && this.rubricsCriterionAction && (this.entity?.canCopy || this.entity?.canDelete) ? html`
```
imcdonald-d2l, line 349: Nit: I'm personally not a big fan of a bunch of ternary conditions in the main `render()` function and think it would look cleaner and easier to read with separate render functions but that isn't a hill I will die on.

src/components/editor/rubric-criterion-editor.js
```diff
@@ -0,0 +1,802 @@
...
+											aria-label-langterm="criterionFeedbackAriaLabel"
+										></d2l-rubric-feedback-editor>
+									</div>
+								`)}
+							</div>
+						` : null}
+					</div>
+
+					<!-- TODO: use classcat or something better - eslint indentation is still a bit janky -->
+					<div class="cell criterion-out-of${this._isTextOnly ? ' -text-only' : ''}${this.isHolistic ? ' -holistic' : ''}">
```
imcdonald-d2l, line 443: If you were to `reflect: true` these properties you could conditionally style them in css instead of having to add all these custom style classes. See example [here](https://github.com/Brightspace/d2l-rubric/blob/master/src/components/rubric-feedback.js#L74).

src/components/editor/rubric-criterion-editor.js
```diff
@@ -0,0 +1,802 @@
...
+										label="${this.localize('criterionOutOf', 'name', this.entity?.name, 'value', this.entity?.outOf)}"
+										@input="${this._handleOutOfInput}"
+										@blur="${this._handleOutOfBlur}"
+										@save="${this._handleOutOfSave}"
+									></d2l-rubric-autosaving-input>
+								</div>
+								${this._outOfError ? html`
+									<d2l-tooltip
+										for="input-out-of"
+										class="is-error"
```
imcdonald-d2l, line 461: This should be `state="error"` to align with current core component [properties](https://github.com/BrightspaceUI/core/tree/main/components/tooltip#properties). See [example](https://github.com/Brightspace/d2l-rubric/blob/master/src/components/editor/rubric-alignment-editor.js#L148).

src/editor/d2l-rubric-criteria-editor.js
```diff
@@ -94,6 +94,7 @@ const template = `<dom-module id="d2l-rubric-criteria-editor">
 
 			.dnd-drag-handle {
 				cursor: move;
+				height: 42px;
```
imcdonald-d2l, line 97: Should this value be a css variable?

src/entities/rubric-criterion-entity.js
```diff
@@ -0,0 +1,42 @@
+import { Classes, Rels } from 'd2l-hypermedia-constants';
+import { EntityBase } from './entity-base.js';
+import { RubricCriterionCellEntity } from './rubric-criterion-cell-entity.js';
+
+export class RubricCriterionEntity extends EntityBase {
+	get activitiesHref() {
+		return this._entity?.getLinkByRel(Rels.Activities.activityUsage)?.href;
+	}
+
+	get activityAlignmentsHref() { // TODO: Add href to HM constants
```
imcdonald-d2l, line 10: You planning on doing this `TODO` in this PR?

semibran: Probably not. I'll have to get something merged into d2l-hypermedia-constants first

src/components/editor/rubric-criterion-editor.js
semibran: Huh, guess another one of our files must be importing it somehow. Weird how that's making the vars register globally.

src/components/editor/rubric-criterion-editor.js
imcdonald-d2l: Yeah, both approaches aren't ideal. But not sure if it's worth being consistent across components? Right now every lit component is using render functions and this one breaks that consistency.

src/components/editor/rubric-criterion-editor.js
imcdonald-d2l: Yeah, that is odd. Wouldn't hurt to explicitly import it just in case something changes down the road with the component structure.

src/components/editor/rubric-criterion-editor.js
```diff
@@ -0,0 +1,835 @@
...
+			criterionDetailWidth: { type: Number, attribute: 'criterion-detail-width' },
+			disabled: { type: Boolean, attribute: 'disabled' },
+			isHolistic: { type: Boolean, attribute: 'is-holistic', reflect: true },
+			isLast: { type: Boolean, attribute: 'is-last' },
+			useNewBrightspaceEditor: { type: Boolean, attribute: 'use-new-brightspace-editor' }, // proposed rename: enabledNewHtmlEditor
+			richTextEnabled: { type: Boolean, attribute: 'rich-text-enabled' },                  // proposed rename: enabledRichText
+			rubricsCriterionAction: { type: Boolean, attribute: 'rubrics-criterion-action' },    // proposed rename: enabledCriterionActions
+			partialEditMode: { type: Boolean, attribute: 'partial-edit-mode' },                  // proposed rename: isPartiallyEditing
+			_isOutcomesListEmpty: { attribute: false },
+			_isTextOnly: { type: Boolean, attribute: 'text-only', reflect: true },
```
semibran, line 60: wasn't sure how to handle this particular attribute - it's not public facing but it's also somewhat helpful to have it reflected for styling stuff. If it's simpler I could always just revert it back to the way it was (computed prop that flips a CSS class on the target element).

imcdonald-d2l:
> Visual diff tests failed - pull request #1304 has been opened with the updated goldens.

@semibran I'm seeing some issues with these diffs that got merged. 
image

src/components/editor/rubric-criterion-editor.js
imcdonald-d2l: As long as this functions as expected and doesn't break anything I'm fine with keeping it around. Note that it should probably be `:host([text-only])` **not** `:host([is-text-only])` to match the defined attribute name.

imcdonald-d2l (approved):

PR #1298, opened 2021-11-08
Move event handler back to d2l-td and fix click issues in different browsers - DE45869

imcdonald-d2l:
Moving click handler to `d2l-td` instead of the `d2l-rubric-criterion-cell` as a lot of styling issues were presented when trying to  make child the full height of the parent. This solution is cleaner and less hacky than previous attempts. Seems to work nicely in chrome and firefox so I expect it will work in safari as well. 

Defect: https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fdefect%2F613372307005


devpow112:
This probably won't  pass until #1299 is merged. Something strange is going on with node 16/npm 8 right now.

semibran (approved):

PR #1314, opened 2021-11-15
';./hacky script.sh;#

awikkerink:


devpow112:
Because the source ref is always qouted you'll have to change the action script to not quite the ref
```
SOURCE_REF='refs/heads/';./hacky-script.sh;#'
```
I'm assuming your trying to get that to be
```
SOURCE_REF='refs/heads/';./hacky-script.sh;#
```

devpow112:
Ah NM I see what you did got it to execute

PR #1316, opened 2021-11-15
';./hacky script.sh;exit;#

awikkerink:


awikkerink:
This doesn't appear to break anything. According to https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token, forks are only ever given readonly access. So, this shouldn't be an issue.

Just checking things out.

PR #1332, opened 2021-11-19
Group flags and langterms into rubric context namespace

semibran:
The main draw here is that it allows us to avoid excess property propagation - just register flags/langterms at the top level, pass thru the context, and use wherever. No more `richTextEnabled: { type: Boolean, ... }` in subcomponent properties.


src/editor/d2l-rubric-editor.js
```diff
@@ -695,6 +686,21 @@ const component = {
...
+			enabledNewHtmlEditor: this.useNewBrightspaceEditor,
+			enabledCriterionActions: this.rubricsCriterionAction,
+			isRubricSinglePage: this.isSinglePageRubric,
+			isPartiallyEditing: this._partialEditMode,
+			langterms: {
+				alignOutcomes: this.alignOutcomesText,
+				browseOutcomes: this.browseOutcomesText,
+				outcomes: this.outcomesTitle,
+			},
+		};
```
semibran, line 702: This is the top level where all the context stuff gets initiated. It might not be worth it to try and rename all these fields in the lms.

The feature flag-like attributes are prefixed with `enabled` - I tried making a `flags` namespace at first but I wasn't happy with how verbose things were getting.

To me the ideal seems to be having all the rubric properties prefixed with `isRubric` e.g. `isRubricHolistic` or `isRubricTextOnly` so we don't have to propagate/inherit that data through HM entities and rubric subcomponents since all that stuff is getting duplicated in a really inefficient way right now. Wasn't quite sure how to deal with the name collision between `isRubricSinglePage` and the rest of those rubric properties but I figured it's best to just not lose sleep over it. Either way the key drawback to all this (and why it's not included in this PR) is that we rely on reflected attributes for styling a lot, so I'll have to see if there's a workaround for that that doesn't involve a nasty amount of boilerplate

semibran:
Just a note - all feature flags are and have been default off, including `rubrics-criterion-action`. We can change this in the future if we want the copy/delete dropdown to show up in the criterion editor instead of the trash can

src/components/editor/rubric-criterion-editor.js
```diff
@@ -404,42 +398,51 @@ class RubricCriterionEditor extends composeMixins(
...
 
 	get _canShowOutcomesButton() {
-		return this.outcomesTitle
+		return this.context?.langterms?.['outcomes']
 			&& this._isAlignmentListEmpty
 			&& !this._isIntentListEmpty
-			&& !this.partialEditMode;
+			&& this.context?.enabledOutcomesToolIntegration
+			&& !this.context?.isPartiallyEditing;
 	}
```
semibran, line 430: example usage

src/editor/d2l-rubric-editor.js
imcdonald-d2l: Do you mind explaining the benefits to pre-fixing context properties with `enabled` or `is`? Personally, I don't think prefixing flags with `enabled` or `is` increases code read-ability in most cases. For example, `isPartiallyEditing` isn't as clear as `partialEditModeEnabled` and `enabledNewHtmlEditor` isn't as clear `newHtmlEditorEnabled`. Want to make sure I understand the benefits before making code potentially harder to read/understand .

semibran: The goal with the prefixes is to group variables with the same general use case semantically. There's a fuzzy "convention" (not formally defined to my knowledge) floating around the JS community and in some other spaces where booleans are prefixed with auxiliaries like `is`/`was`/`can`/`will` etc. to distinguish them from verb prefixes like `open` or `render` which are typically reserved for methods. Using `enabled` in this scenario is meant to be within that same spirit but specifically for LMS-side feature flags. This might not be a valuable distinction to make, in which case we could prefix all of that stuff with `is` and suffix with `Enabled`. Personally I'm against just going freestyle on the first letters for everything though.

The reason why grouping with a suffix alone doesn't work as well in this scenario is twofold: 1. variable names of different lengths break visual continuity when scanning compared to prefixes, and 2. large variable groups get broken up by auto-alphabetization.

Obviously there's a lot of overlap between different categories so none of the names here are really set in stone in my mind. Open to changing all of this stuff to something we can agree on - one of the (marginal) benefits about the context setup is that in order to change the names here we only have to find and replace e.g. `enabledNewHtmlEditor` in a few places without having to change 10 different files with the kebab-case stuff mixed in

imcdonald-d2l: Thanks for taking the time to explain that, that makes more sense now. As far as finding a pattern for LMS-side feature flags, I understand how it could be nice to clearly identify them. But since the `enabled` prefix to me really obfuscates the intention of the variable I would prefer if we could prefix it with `is` and suffix it with `Enabled` like you suggested. That would make the understanding of the variable more clear I think.

src/components/editor/rubric-criterion-editor.js
```diff
@@ -383,42 +377,51 @@ class RubricCriterionEditor extends composeMixins(
...
+		return this.entity?.canDelete
+			&& !this.context?.isCriterionActionsEnabled
+			&& !this.context?.isPartialEditModeEnabled;
 	}
 
 	get _canShowMoreMenu() {
-		return !this.partialEditMode
-			&& this.rubricsCriterionAction
+		return !this.context?.isPartialEditModeEnabled
+			&& this.context?.enabledCriterionActions
```
imcdonald-d2l, line 391: Is this supposed to be `isCriterionActionsEnabled`?

src/editor/d2l-rubric-editor.js
```diff
@@ -369,19 +360,18 @@ const template = `
 								id="rubric-description"
 								key="[[_getSelfLink(entity)]]"
 								token="[[token]]"
+								context="[[context]]"
```
imcdonald-d2l, line 363: Is this supposed to be `_context`?

imcdonald-d2l (approved):

src/components/editor/rubric-criteria-editor.js
```diff
@@ -26,19 +26,12 @@ class RubricCriteriaEditor extends composeMixins(
 
 	static get properties() {
 		return {
+			context: { type: Object },
+			rubricLevelLoaMapping: { type: Map, attribute: 'rubric-level-loa-mapping' },
```
devpow112, line 30: Does `map` actually work? Doesn't look like it's one of the default converters
https://lit-element.polymer-project.org/guide/properties#conversion-type

semibran: It doesn't, I'm pretty sure it just defaults to `Object`. In this case the mapping is "supposed" to be an object since there are non-href keys (it stores the reversed state for some reason). I'm looking into a solution for that in the overlay conversion

devpow112 (approved):

MitchelPaulin (approved):

PR #1334, opened 2021-11-19
Converted exisiting translated components to using composeMixin

MitchelPaulin:


src/components/rubric-alignments-indicator.js
```diff
@@ -13,8 +14,14 @@ import { LocalizeMixin } from '../mixins/localize-mixin.js';
 import { OutcomeEntity } from '../entities/outcome-entity.js';
 import { RubricCriterionActivityUsageEntity } from '../entities/rubric-criterion-activity-usage-entity.js';
 import { rubricStyles } from '../styles/rubric-styles.js';
+EntityWatcherMixin(LocalizeMixin(EntityMixin(LitElement)))
```
devpow112, line 17: I'm guessing this was a copy/paste error?

MitchelPaulin: Correct

devpow112:
You can actually use the `composeMixins` from brightspace ui core now. Brandon was good enough to port it over.
https://github.com/BrightspaceUI/core/blob/main/helpers/composeMixins.js

src/components/rubric-alignments-indicator.js
MitchelPaulin: Funnily enough the linter actually caught it because I was missing a semi colon lol

devpow112 (approved):

PR #1335, opened 2021-11-19
Removed intermediary handler for width changed event

MitchelPaulin:
A small proposed change here, thought it was strange that we were changing the property directly on a web component. This just moves the event handler into the web component that cares about the property change and makes the flow a bit easier to follow.


src/components/editor/rubric-levels-editor.js
```diff
@@ -189,7 +189,7 @@ class RubricLevelsEditor extends composeMixins(
 		requestAnimationFrame(() => {
 			const section = this.shadowRoot.querySelector('#levels-section');
 			if (section && section.offsetWidth) {
-				this.dispatchEvent(new CustomEvent('d2l-rubric-editor-levels-width-changed', {
+				window.dispatchEvent(new CustomEvent('d2l-rubric-editor-levels-width-changed', {
```
devpow112, line 192: Does using `this.` still not work here? Usually you want to keep using `this` where possible.

MitchelPaulin: Yes, when publishing to `this.` the event wasn't being received by the criteria editor

src/components/editor/rubric-criteria-editor.js
```diff
@@ -187,6 +187,20 @@ class RubricCriteriaEditor extends composeMixins(
 		this.partialEditMode = false;
 		this.useNewBrightspaceEditor = false;
 	}
+	connectedCallback() {
+		super.connectedCallback();
+		window.addEventListener('d2l-rubric-editor-levels-width-changed', (e) => {
```
devpow112, line 192: If you switch this to use `this` and does the dispatch work with `this` then?

MitchelPaulin: No, but I checked and dispatch the event can use `this.`, but the listeners need to be listening to `window`

MitchelPaulin:
Vdiff changes are from resizing issues, though most work, oddly enough its only the analytic points rubrics which are failing, will look into this further if I have some extra time

PR #1340, opened 2021-11-22
`rubrics-brightspace-editor` flag set to false was breaking rubrics -  DE46131

imcdonald-d2l:
With `rubrics-brightspace-editor` set to `false` it was causing rubrics editor functionality to break. Old editor wasn't functioning as expected, couldn't click in it or edit the text for all description editor fields

### Issue
Reason for issue was that the [this.disabled](https://github.com/Brightspace/d2l-rubric/blob/master/src/components/editor/rubric-text-editor.js#L99) field on legacy polymer editor was updating from `true` to `false` through the component update cycle. `tinymce` didn't catch this update in the legacy editor and didn't update the appropriate classes to make the editor editable. 

old-editor-defect

### Fix
To fix this issue I had to prevent the editor `this.disabled` property from flip-flopping values. To do that I just don't render anything until `this.entity` is loaded. If anyone has any better ideas to force the legacy editor to re-render `tinymce` classes with changes I'm all ears. 

editor-flag-fixed

Rally defect: https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fdefect%2F616557383953


src/components/editor/rubric-criteria-group-editor.js
```diff
@@ -458,20 +458,23 @@ class RubricCriteriaGroupEditor extends composeMixins(
 	}
 
 	_renderFooter() {
-		if (this.partialEditMode || this.isHolistic) {
```
imcdonald-d2l, line 461: This file contains unrelated changes to the main defect. Noticed that the bottom border was missing on partial edit mode. This fixes that issue.

src/components/editor/rubric-criterion-editor.js
```diff
@@ -80,17 +80,6 @@ class RubricCriterionEditor extends composeMixins(
 					display: flex;
 					--d2l-rubric-editor-browse-outcomes-button-margin: 5px;
 					--d2l-rubric-editor-outcomes-list-label-margin: calc(0.6rem + var(--d2l-rubric-editor-browse-outcomes-button-margin));
-					--d2l-input-textarea: {
```
imcdonald-d2l, line 83: Unrelated change here as well. Don't know how this css variable assignment was working in polymer but wasn't doing anything in lit. Removing it as it wasn't making any difference in the UI.

semibran (commented):
Haven't been seeing this on master, are there extra repro steps?

imcdonald-d2l:
> Haven't been seeing this on master, are there extra repro steps?

Set `"rubrics-brightspace-editor":false` in launch darkly overrides file should cause the error described.

semibran (approved):

PR #1341, opened 2021-11-22
Convert `d2l-rubric-autosaving-input` US132824

imcdonald-d2l:
https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F611712072471


src/components/editor/rubric-autosaving-input.js
```diff
@@ -0,0 +1,79 @@
...
+
+	render() {
+		return html`
+			<d2l-input-text
+				id="input"
+				value="${this._getDisplayedValue(this._isFocused, this.invalid, this.pendingSaves, this._enteredValue, this.value)}"
+				@focus="${this._onFocus}"
+				@blur="${this._onBlur}"
+				@input="${this._onInput}"
+				?aria-invalid="${Boolean(this.invalid).toString()}"
```
MitchelPaulin, line 34: This doesn't look right, if you are using a boolean binding the result should be a boolean, this may also be why the test was failing

imcdonald-d2l: Yeah, I copied and pasted without thinking much about it. I'll take a look at this and fix it.

imcdonald-d2l: So `aria-invalid` is actually supposed to be a `string` property. Fixed it by removing the `?` in front of the property.

imcdonald-d2l: See core component docs: https://github.com/BrightspaceUI/core/blob/main/components/inputs/docs/input-text.md#properties

MitchelPaulin (approved):

PR #1349, opened 2021-11-24
`d2l-rubric-structure-editor` conversion - US132996

imcdonald-d2l:
https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F612451718597


src/components/editor/rubric-criteria-group-editor.js
```diff
@@ -49,6 +49,7 @@ class RubricCriteriaGroupEditor extends composeMixins(
 			showGroupName: { type: Boolean, attribute: 'show-group-name' },
 			isHolistic: { type: Boolean, attribute: 'is-holistic' },
 			isReordered: { type: Boolean, attribute: 'is-reordered' },
+			updatingLevels: { type: Boolean, attribute: 'updating-levels' },
```
semibran, line 52: correct me if I'm wrong but the `updatingLevels` stuff was originally a two-way data binding chain from the level-editor component all the way up to the structure editor right? So we don't necessarily need to define the prop/send up the event manually from each step in the chain, we could just try bubbling a single composed event from the base. I think Mitch was doing some of that work in the criteria-groups conversion

imcdonald-d2l: We actually need to send  the event **down** in this case since `structure-editor` can modify the updating levels state when reversing the order. So we need to pass the variable down all the way to the `description-editor` and `feedback-editor`. That's why I added they properties back on the components. Without them the description and feedback editor won't get the proper state.

semibran: Gotcha. Thanks for clarifying

semibran:
FYI -- noticed the score shows up on text only rubrics as `- / undefined` in the new goldens

image

imcdonald-d2l:
> FYI -- noticed the score shows up on text only rubrics as `- / undefined` in the new goldens
> 
> image

Yup, saw that too. Fixed in this [commit](https://github.com/Brightspace/d2l-rubric/pull/1349/commits/de07bdc3f9589952add8b112167d245811eb5e60).  Lit converts it to string automatically so can't use `typeof`

src/components/editor/rubric-structure-editor.js
```diff
@@ -0,0 +1,518 @@
...
+					display: flex;
+					justify-content: space-between;
+					flex-wrap: wrap;
+					margin-left: var(--d2l-rubric-editor-start-gutter-width); /* ie11 fix */
+					margin-left: calc(var(--d2l-rubric-editor-start-gutter-width) - 0.1rem);
+					margin-right: var(--d2l-rubric-editor-end-gutter-width); /* ie11 fix */
+					margin-right: calc(var(--d2l-rubric-editor-end-gutter-width));
+				}
+
+				:dir(rtl) #rubric-structure-editor-header-content {
```
semibran, line 95:
```suggestion
				:host-context([dir='rtl']) #rubric-structure-editor-header-content {
```

semibran (approved):

semibran (approved):

PR #1354, opened 2021-11-25
US132999 - Convert competencies icon to LitElement

semibran:
[`3ff3846`](https://github.com/Brightspace/d2l-rubric/commit/3ff3846fab80b1e2ac055a6dd189a002d0564370) for conversion diff


src/components/rubric-competencies-icon.js
```diff
@@ -0,0 +1,55 @@
...
+
+	constructor() {
+		super();
+		this.tooltipPosition = 'top';
+	}
+
+	connectedCallback() {
+		super.connectedCallback();
+		this.addEventListener('blur', () => this._hasFocus = false);
+		this.addEventListener('focus', () => this._hasFocus = true);
```
semibran, line 33: Doesn't appear to change anything with the focus behavior. Leaving it in case I'm missing something.

src/components/rubric-competencies-icon.js
```diff
@@ -0,0 +1,55 @@
...
+		return html`
+			<img
+				id="icon"
+				src="${competencyIconPng}"
+				tabindex="0"
+			/>
+
+			<d2l-tooltip
+				for="icon"
+				position="${this.tooltipPosition}"
```
semibran, line 46: local lint starts coughing up blood here since I guess there's no runtime restrictions around `tooltipPosition`? CI seems to be fine though. Not sure what it's looking for

devpow112: That's weird since CI runs the exact same command for linting. Whats the warning/error it spits out?

semibran: just gives `Type 'string' is not assignable to '"top" | "bottom" | "left" | "right"'` - it's reading from the annotations in d2l-tooltip

devpow112: Oh weird. Ok.

devpow112 (approved):

PR #1359, opened 2021-11-26
DE46006 - Fix default feedback no longer overrides custom feedback fo…

jesseag:
…r inline rubrics.


src/d2l-rubric-levels-mobile.js
```diff
@@ -954,7 +959,10 @@ const component = {
 		if (this._preventClick) {
 			return;
 		}
-		this._toggleLevelAt(event.currentTarget.dataIndex);
+		// Delay using afterNextRender so the _blurHandler in
+		// d2l-rubric-feedback will fire before this._toggleLevelAt.
+		let dataIndex = event.currentTarget.dataIndex
```
jesseag, line 964: event.currentTarget.dataIndex is null within the anonymous function. As a result, had to declare it outside

src/d2l-rubric-levels-mobile.js
```diff
@@ -954,7 +959,10 @@ const component = {
 		if (this._preventClick) {
 			return;
 		}
-		this._toggleLevelAt(event.currentTarget.dataIndex);
+		// Delay using afterNextRender so the _blurHandler in
+		// d2l-rubric-feedback will fire before this._toggleLevelAt.
+		const dataIndex = event.currentTarget.dataIndex;
```
jesseag, line 964: event.currentTarget.dataIndex is null within the anonymous function. As a result, had to declare it outside.

jesseag:
The Pop-out rubric view has the correct order of events that occurs:

1) _blurHandler in d2l-rubric-feedback.js, Line 411
https://github.com/Brightspace/d2l-rubric/blob/3.x/src/d2l-rubric-feedback.js
2) set html function gets called in brightspaceUI/html-editor, Line 293
https://github.com/BrightspaceUI/htmleditor/blob/ff1f90292beb250fdf82cb431ad4bcd7d0e83f7a/htmleditor-mixin.js

The bug in the inline rubric view had the reversed order of events that occurred:

1)  set html function gets called in brightspaceUI/html-editor, Line 293
2) _blurHandler in d2l-rubric-feedback.js

To fix this race condition, we use afterNextRender() to allow the _blurHandler to be the first event that happens in the inline view.

src/d2l-rubric-levels-mobile.js
```diff
@@ -920,11 +921,15 @@ const component = {
...
+		// Delay using afterNextRender so the _blurHandler in
+		// d2l-rubric-feedback will fire before this._handleMove.
+		afterNextRender(this, () => {this._handleMove(event, -1);});
 	},
 
 	_handleRight: function(event) {
-		this._handleMove(event, 1);
+		// Delay using afterNextRender so the _blurHandler in
+		// d2l-rubric-feedback will fire before this._handleMove.
+		afterNextRender(this, () => {this._handleMove(event, 1);});
```
MitchelPaulin, line 932: Looking at this again, since we have a delay already in `_handleMove` do we need these additional delays in the `_handleRight` and `_handleLeft` methods?

src/d2l-rubric-levels-mobile.js
MitchelPaulin: This might have only been the case with `setTimeout`, worth trying again just to see

jesseag: I tried the following but unfortunately, the same error was still observed.

afterNextRender(this, () => {this._toggleLevelAt(event.currentTarget.dataIndex);});

Error:
render-status.js:56 Uncaught TypeError: Cannot read properties of null (reading 'dataIndex')
    at HTMLElement.<anonymous> (d2l-rubric-levels-mobile.js:965)

src/d2l-rubric-levels-mobile.js
jesseag: Yes, we do need them. I removed them and tried testing without them and the left/right chevron arrows give the same incorrect behaviour without this afterNextRender here. Originally, the left/right chevron arrows were not part of the defect but since it is a simple fix then I figure we might as well fix it.

image

MitchelPaulin: Yes we should leave them in. I misread and I thought we added a delay to `_handleMove` (github kind of tricked me and made it look like we changed `_handleMove` lol)
image

At any rate we could probably get away with moving the afterNextRender into `_handleMove` but thats not a big deal, I'm fine with how it is currently

src/d2l-rubric-levels-mobile.js
MitchelPaulin: Looks like its a quirk with `currentTarget` https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget

src/d2l-rubric-levels-mobile.js
jesseag: I see what you mean regarding GitHub being confusing. I can definitely move afterNextRender into _handleMove if others feel that way so we have less code lines of code. I'll wait to see what others say.

semibran: It does feel like a bit of a nit but I think it makes more logical sense to go ahead and add it into the move handler. Could probably go without the braces/semi as well.

jesseag: Done. Had to move all 3 statements into afterNextRender() to get it to work. Otherwise, if I tried to use afterNextRender on only 1 statement in _handleMove I was still able to reproduce the defect.

src/d2l-rubric-levels-mobile.js
```diff
@@ -928,9 +929,13 @@ const component = {
...
 
 	_handleMove: function(event, delta) {
-		event.preventDefault();
-		event.stopPropagation(delta);
-		this._moveSelection(delta);
+		// Delay using afterNextRender so the _blurHandler in
+		// d2l-rubric-feedback will fire before the below statements.
+		afterNextRender(this, () => {
+			event.preventDefault();
+			event.stopPropagation(delta);
```
semibran, line 936:
```suggestion
			event.stopPropagation();
```
this was a typo on my end, stopPropagation doesn't take args

jesseag: Ahh ok. Well thats good to know.
https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation

jesseag: Is it necessary to re-test this? I'm in the middle of setting up Bsi for 4.x so I can't really test it. If we don't need to test then I am fine to just commit it. As you said event.stopPropagation doesn't take any parameters so removing this variable shouldn't change any behaviour. 
Since javascript is loosely typed then I don't think we should really need to re test this.
https://stackoverflow.com/questions/11107823/what-happens-if-i-dont-pass-a-parameter-in-a-javascript-function#:~:text=JavaScript%20functions%20can%20be%20invoked,any%20type%20to%20any%20function.

semibran: Shouldn't be necessary to re-test. It's interesting that stopPropagation doesn't panic at all with the extra arg since some DOM methods are really picky about the props that you pass in. It won't change anything about the behavior though since it's not supposed to be receiving anything in the first place.

On top of that I'd doubt stopPropagation is even necessary here since the radio buttons shouldn't have any children that receive key events. I must've just added it as a failsafe on top of preventDefault. That's something I'd have to re-test though since native html inputs can be weird at times

jesseag: Sure that works for me. I'll commit this.

semibran (approved):

PR #1360, opened 2021-11-26
Silence demo 500s for POST

devpow112:
They can be annoying in demo pages. Makes sure to not silence them in tests.


wds.config.mjs
```diff
@@ -17,3 +18,17 @@ export default {
...
 	]
 };
+
+export default merge(
+	devServerConfig,
+	{
+		plugins: [{
+			name: 'stub-post',
+			serve(context) {
+				if (context.request.method === 'POST') {
```
devpow112, line 28: Intercepts all `POST` calls and returns an empty JSON body

devpow112: Also we may be able to somehow use this to emulate posts for tests if need be

devpow112: I'll also do this setup to other repos as needed/desired

semibran (approved):

PR #1361, opened 2021-11-29
`d2l-rubric-editor` conversion - US132991

imcdonald-d2l:
https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F612451196021


src/components/editor/rubric-structure-editor.js
```diff
@@ -412,15 +398,6 @@ class RubricStructureEditor extends composeMixins(
 		`;
 	}
 
-	_renderEditorHeader() {
-		if (this.context?.isRubricSinglePage) {
```
imcdonald-d2l, line 416: Removing all references to `isRubricSinglePage` as it is an obsolete config that we don't need to handle anymore

src/components/editor/rubric-text-editor.js
```diff
@@ -36,7 +35,8 @@ class RubricTextEditor extends composeMixins(
 						}
 					}
 				}
-			}
+			},
+			_inputChanging: { attribute: false }
```
imcdonald-d2l, line 39: Changed `inputChanging` to a private property as it shouldn't be set externally. All consumers of this component now listen to `input-changing` event to observe changes to this state.

test/d2l-rubric-editor.test.js
```diff
@@ -258,31 +258,6 @@ describe('d2l-rubric-editor', () => {
...
-			it('sets aria-invalid if saving score visibility change fails', done => {
-				const fetch = sandbox.stub(window.d2lfetch, 'fetch');
-				const promise = Promise.resolve({
-					ok: false,
-					json: () => Promise.resolve(JSON.stringify({}))
-				});
-
-				fetch.returns(promise);
-
-				const subElement = element.$$('#hide-score-checkbox');
```
imcdonald-d2l, line 271: This element is a [d2l-input-checkbox](https://github.com/BrightspaceUI/core/blob/main/components/inputs/docs/input-checkbox.md#properties) which doesn't have a `ariaInvalid` property. Removing this test since it assumes that property exists.

src/entities/rubric-entity.js
```diff
@@ -39,11 +115,35 @@ export class RubricEntity extends EntityBase {
...
+		return this._entity?.getActionByName('update-status')?.getFieldByName('rubric-status')?.value || [];
+	}
+
+	getDescription(richTextEnabled) {
+		let descriptionValue = this._entity?.getActionByName('update-description')?.getFieldByName('description')?.field;
+
+		if (!descriptionValue) {
+			const descriptionEntity = this._entity?.getSubEntityByClass(Classes.rubrics.description);
+			descriptionValue = descriptionEntity &&
+			(richTextEnabled ? descriptionEntity?.properties?.html : descriptionEntity?.properties?.text);
```
MitchelPaulin, line 144:
```suggestion
			descriptionValue = richTextEnabled ? descriptionEntity?.properties?.html : descriptionEntity?.properties?.text;
```

src/components/editor/rubric-editor.js
```diff
@@ -0,0 +1,1071 @@
...
+	disconnectedCallback() {
+		window.removeEventListener('d2l-siren-entity-save-start', (e) => this._onEntitySave(e));
+		window.removeEventListener('d2l-siren-entity-save-error', (e) => this._onEntitySave(e));
+		window.removeEventListener('d2l-siren-entity-save-end', (e) => this._onEntitySave(e));
+		window.removeEventListener('d2l-siren-entity-error', (e) => this._handleError(e));
+		window.removeEventListener('d2l-rubric-editor-save-error', (e) => this._handleSaveError(e));
+		super.disconnectedCallback();
+	}
+
+	// TODO - verify this first updated function works
```
MitchelPaulin, line 355: Does this still need to be verified?

src/components/editor/rubric-editor.js
```diff
@@ -0,0 +1,1071 @@
...
+	}
+
+	_getRelativeUrl(absoluteUrl) {
+		const a = document.createElement('a');
+		a.href = absoluteUrl;
+		return (a.pathname[0] === '/' ? '' : '/') + a.pathname + a.search;
+	}
+
+	_getTelemetryEndpoint() {
+		return (this.dataset && this.dataset.telemetryEndpoint) || null;
```
MitchelPaulin, line 483:
```suggestion
		return this.dataset?.telemetryEndpoint;
```

src/components/editor/rubric-editor.js
```diff
@@ -0,0 +1,1071 @@
...
+				this._rubricStatusText = this.localize('rubricStatusArchived');
+			} else if (entity.isDraft) {
+				this._rubricStatusText = this.localize('rubricStatusDraft');
+			}
+		}
+
+		this._isHolistic = entity.isHolistic;
+		this._updateName(entity);
+		this._updateDescription(entity);
+		// TODO - does this break stuff?
```
MitchelPaulin, line 630: Same here

src/components/editor/rubric-structure-editor.js
```diff
@@ -412,15 +398,6 @@ class RubricStructureEditor extends composeMixins(
...
-			return;
-		}
-		return html`
-			<d2l-rubric-editor-header id="rubric-header"></d2l-rubric-editor-header>
```
MitchelPaulin, line 420: Double check, we dont want to render this header anymore?

src/components/editor/rubric-editor.js
imcdonald-d2l: This has been verified - will remove comment.

src/components/editor/rubric-editor.js
```diff
@@ -0,0 +1,1071 @@
...
+			makeRubricAvailableElem.addEventListener('d2l-siren-entity-save-error', this._onShareRubricSaveError.bind(this));
+		}
+
+		const telemetryData = {
+			endpoint: this._getTelemetryEndpoint(),
+			errorEndpoint: this.errorLoggingEndpoint,
+		};
+		this.setTelemetryData(telemetryData);
+
+		// TODO - make sure name is being focused
```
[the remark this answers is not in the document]
imcdonald-d2l: This still needs to be double checked. Left myself a note and forgot to circle back to it.

src/components/editor/rubric-editor.js
imcdonald-d2l: This still needs to be double checked. Left myself a note and forgot to circle back to it.

src/components/editor/rubric-structure-editor.js
imcdonald-d2l: `isRubricSinglePage` will always be true so that is why it is being removed. This will never be needed as a result. The header just shows the `Saving...` component twice if this is added by default.

MitchelPaulin: Got it, wasn't sure if `isRubricSinglePage` was always true or false now

src/components/editor/rubric-editor.js
imcdonald-d2l: Verified this is not an issue.

src/components/editor/rubric-editor.js
```diff
@@ -0,0 +1,1071 @@
...
+					bubbles: true,
+					composed: true,
+				}));
+
+				const scoreVisibilityLangTerm = checkbox.checked ? 'scoresVisibilityHidden' : 'scoresVisibilityVisible';
+
+				announce(this.localize(scoreVisibilityLangTerm));
+			}).catch((err) => {
+				checkbox.checked = !checkbox.checked;
+				checkbox.style.borderColor = '#cd2026';
```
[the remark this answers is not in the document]
imcdonald-d2l: I don't think you can use css variables in javascript though 🤔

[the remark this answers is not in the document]
imcdonald-d2l: I could look at reflecting the `_setScoreVisibilityFailed` property to conditionally apply styles to clean it up though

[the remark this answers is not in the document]
MitchelPaulin: Looks like it can be accessed but not super nicely https://stackoverflow.com/questions/41725725/access-css-variable-from-javascript

image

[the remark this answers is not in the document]
devpow112: Can you not just Set the border color to `var(--d2l-color-cinnabar)`. That is how you usually resolve CSS variables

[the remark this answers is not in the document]
imcdonald-d2l: Yeah, I'm going to try and conditionally apply styles through reflecting the `_setScoreVisibilityFailed` attribute. Will try a hackier approach if conditional styles/classes don't work.

src/components/editor/rubric-editor.js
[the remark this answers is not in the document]
imcdonald-d2l: Fixed name being focused in [this commit](https://github.com/Brightspace/d2l-rubric/pull/1361/commits/69f8382bf0499038ecf0edeb08cf74c8cf4e2ea1). Hooking into `autofocus` property on `d2l-input-text`.

name-is-focused

src/components/editor/rubric-editor.js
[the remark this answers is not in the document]
imcdonald-d2l: Fixed styles in [this commit](https://github.com/Brightspace/d2l-rubric/pull/1361/commits/69f8382bf0499038ecf0edeb08cf74c8cf4e2ea1). Looks like this now: 

score-visibility-error-state

devpow112 (approved):

PR #1364, opened 2021-11-29
Switch to using `d2l-resize-aware`

devpow112:
This allows us to remove the use of `@polymer/iron-media-query`. Recommend viewing with [white-space diff off](https://github.com/Brightspace/d2l-rubric/pull/1364/files?w=1) as this indents all of the rubric template inside the new `d2l-resize-aware` component.


src/d2l-rubric.js
```diff
@@ -204,7 +208,7 @@ const template = `<dom-module id="d2l-rubric">
 			}
 
 		</style>
-		<iron-media-query query="(max-width: 614px)" query-matches="{{_isMobile}}"></iron-media-query>
+		<d2l-resize-aware on-d2l-resize-aware-resized="_handleResized"></d2l-resize-aware>
```
devpow112, line 211: Ideally we'd have all the content to be resized inside this component. This is closer to the use if `iron-media-query` so leaving it like this for now. We can revisit when we convert this element.

semibran (approved):

PR #1368, opened 2021-11-30
More `d2l-resize-aware` clean-up

devpow112:
Added a missing import and removed an unneeded one.


imcdonald-d2l (approved):
Removed it from my converted `rubric-editor` component just now 👍🏻

PR #1374, opened 2021-11-30
Added a show-published-warning attribute to show a warning that the rubric has been published

mpharoah-d2l:
https://xd.adobe.com/view/1c49f605-e43c-4811-953a-248980d661e8-491a/

PR #1377, opened 2021-12-01
Added a show-published-warning attribute to show a warning that the rubric has been published

mpharoah-d2l:
Same as #1374 but for the `master` branch


MitchelPaulin (approved):

lang/fr-on.js
```diff
@@ -155,6 +155,7 @@ export default {
 	"pointsSaveFailed": "Échec de l’enregistrement des points",
 	"preview": "Prévisualiser",
 	"printRubric": "Imprimer la grille d’évaluation",
+	"publishedWarning": "This assessment has been published. Any changes made to the rubric will be immediately visible to learners. To avoid this, retract feedback before making changes.",
```
MitchelPaulin, line 158: Make sure to google translate the french terms

MitchelPaulin: You are probably right but I think its better to be safe since there are financial consequences if these terms slip through the cracks somehow

MitchelPaulin (approved):

devpow112 (approved):

PR #1380, opened 2021-12-01
`d2l-rubric-overall-score` conversion - US132985

imcdonald-d2l:
https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F612450672825


src/components/rubric-overall-score.js
```diff
@@ -0,0 +1,339 @@
...
+					</div>
+					<div class="info-container d2l-body-small" id="${index}">
+						${this._localizePoints(overallLevel)}
+						<d2l-html-block id="${index}">
+							<template>${unsafeHTML(overallLevel.descriptionHtml)}</template>
+						</d2l-html-block>
+					</div>
+					${this._renderOverrideBtn(overallLevel, index)}
+				</div>
+			</div>
```
imcdonald-d2l, line 306: I added `id="${index}"` to all elements in this `div` so we can get the appropriate `overallLevel` in the `this._levelClicked` and `this._handleKeypress` event handlers but it is certainly not an ideal situation. Anyone have any other ideas on a more graceful way to get the appropriate overall level in the event handlers? 🤔

src/d2l-rubric.js
```diff
@@ -308,7 +308,7 @@ const template = `<dom-module id="d2l-rubric">
 										has-out-of="[[_hasOutOf(entity)]]"
 										compact="[[compact]]"
 										compact-overall-score="[[compactOverallScore]]">
-									</d2l-rubric-overall-score>
+									</d2l-rubric-overall-score-adapter>
```
imcdonald-d2l, line 311: Don't know how this closing tag mess-up made it through the linter but fixing it here.

src/components/rubric-overall-score.js
semibran: Not sure about graceful but I've seen stuff to the effect of `data-href` and `data-index` in other levels-related components. I think the ideal would be defining the handler in the same context where it's assigned whether that's by binding the index or what have you but I know the linter doesn't usually play nice with those kinds of idioms.

semibran (approved):
I think it'd be good to make a note to move more of the entity-related logic into the actual entity classes themselves but that looks like it'd be a lot of work. this is probably fine for now

imcdonald-d2l:
> I think it'd be good to make a note to move more of the entity-related logic into the actual entity classes themselves but that looks like it'd be a lot of work. this is probably fine for now

What specific entity-related logic did you see that you felt should be moved?

src/components/rubric-overall-score.js
imcdonald-d2l: Not sure I understand why you mean by,
> defining the handler in the same context where it's assigned

could you elaborate? Do you mean like: `@click=${(e} => { define some function logic }}` instead of assigning a pre-defined function? I was going to try and do that, passing in `index` with the event but lit does not like that. It throws [this no template arrow error](https://github.com/43081j/eslint-plugin-lit/blob/master/docs/rules/no-template-arrow.md).

semibran: Yeah basically. I haven't tried partially applying the index but that'd be another option too provided eslint doesn't vomit.

semibran:
> What specific entity-related logic did you see that you felt should be moved?

- [`_ariaChecked`](https://github.com/Brightspace/d2l-rubric/pull/1380/files#diff-9cc6ce59bba8dd4e534c09a718c7289d854d678287a8b0a1344c43473ef622c9R199-R204)
- [`_getAssessmentLevel`](https://github.com/Brightspace/d2l-rubric/pull/1380/files#diff-9cc6ce59bba8dd4e534c09a718c7289d854d678287a8b0a1344c43473ef622c9R206-R208)
- [`_isAchieved`](https://github.com/Brightspace/d2l-rubric/pull/1380/files#diff-9cc6ce59bba8dd4e534c09a718c7289d854d678287a8b0a1344c43473ef622c9R225-R228)
- [`_isClickable`](https://github.com/Brightspace/d2l-rubric/pull/1380/files#diff-9cc6ce59bba8dd4e534c09a718c7289d854d678287a8b0a1344c43473ef622c9R230-R233)

Basically anything that receives an entity that doesn't require the dynamic localization stuff. We were using Matt's assessment helpers before for that kind of thing but the entity class framework supersedes that to an extent (to my knowledge -- at least for logic that only applies to a single entity like score getters etc). The `_assessmentMap` stuff for instance is really abstract and it'd be better to move that outside of the component into a more self-contained interface. Again if you don't feel like that's something worth tackling at this stage I think that's okay as well.

imcdonald-d2l:
> The `_assessmentMap` stuff for instance is really abstract and it'd be better to move that outside of the component into a more self-contained interface. Again if you don't feel like that's something worth tackling at this stage I think that's okay as well.

Agreed, this logic could be abstracted to a class to clean it up. I will circle back to it when converting another component that uses the same logic. 👍🏻

PR #1382, opened 2021-12-02
Add stylelint and correct all issue it reports

devpow112:
Add stylelint and correct issues it reports. Recommended viewing with [white-space diff off](https://github.com/Brightspace/d2l-rubric/pull/1382/files?w=1) as it changed indenting.


.stylelintrc.json
```diff
@@ -0,0 +1,14 @@
+{
+  "extends": "@brightspace-ui/stylelint-config",
+  "rules": {
+    "order/properties-alphabetical-order": null,
+    "function-name-case": [
```
devpow112, line 5: Porting this rule over into the stylelint config. Will remove once that merges.

.stylelintrc.json
```diff
@@ -0,0 +1,14 @@
+{
+  "extends": "@brightspace-ui/stylelint-config",
+  "rules": {
+    "order/properties-alphabetical-order": null,
```
devpow112, line 4: Still having conversations about this rule. For now just disabling it so no sorting occurs.

imcdonald-d2l (approved):

PR #1387, opened 2021-12-02
Remove unneeded serge config

devpow112:
This is now part of serge's default config. See https://d2l.slack.com/archives/C0TF4KSGZ/p1638465040052200.


devpow112:
NM it's not exactly the same.

PR #1388, opened 2021-12-02
Add lit-analyzer

devpow112:
Found a few minor errors.


tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+  "compilerOptions": {
+    "plugins": [
+      {
+        "name": "ts-lit-plugin",
+        "strict": true,
+        "format": "list",
+        "rules": {
+          "no-unknown-tag-name": "off",
+          "no-complex-attribute-binding": "warn"
```
devpow112, line 10: Might just switch it to a property over an attribute eventually. For now just make the `error` a `warn`.
https://github.com/Brightspace/d2l-rubric/blob/master/src/components/icons/rubric-competencies-icon.js#L10
https://github.com/Brightspace/d2l-rubric/blob/master/src/components/rubric-overall-score.js#L279

src/components/editor/rubric-autosaving-input.js
```diff
@@ -32,7 +32,7 @@ class RubricAutoSavingInput extends composeMixins(
 				@focus="${this._onFocus}"
 				@blur="${this._onBlur}"
 				@input="${this._onInput}"
-				aria-invalid="${Boolean(this.invalid).toString()}"
+				aria-invalid="${this.invalid}"
```
MitchelPaulin, line 35: Doesn't aria-invalid expect a string? A common pattern I have seen is `aria-invalid="${this.invalid ? 'true : 'false'}`

devpow112: This should serialize to `aria-invalid="true"` or `aria-invalid="false"` but I can make it explicit with the ternary.

devpow112: Done in 68c4c9c.

MitchelPaulin (approved):

PR #1390, opened 2021-12-02
Re-add missed logic around `key` attribute and HTML editors

devpow112:
Somehow this was missed during the conversion of the text editor and HTML editors (oops my bad). Re-adding logic around the `key` attribute. This also includes some auto clean-up of spacing (apologies for the diff noise).


semibran (approved):
Lol part of this is my bad I hadn't run into the lit property binding weirdness yet. Thanks for picking up on it

devpow112:
> Lol part of this is my bad I hadn't run into the lit property binding weirdness yet

No problem 😄

PR #1392, opened 2021-12-02
DE46032 Fix consistent eval inline rubric statistic button so it is l…

jesseag:
…abeled and can be read from the screen reader.

I'll create a fix for 3.x in d2l-rubric-adapater.js once I get some feedback. 

After updating the label, the following is seen from the NVDA screen reader Elements List. This name matches the Rubric Statistics dialog text box title, Line 333 (same file). I also got NVDA to read my change out loud to ensure that I was editing the correct field. 
Let me know if the localization is incorrect, etc.

Updated_button_name
rubric_statistics_dialog_title

https://github.com/Brightspace/d2l-rubric/blob/master/src/mixins/localize-mixin.js
https://github.com/Brightspace/d2l-rubric/blob/master/lang/en.js


semibran (approved):

devpow112 (approved):

jesseag:
Is the string "Rubric Statistics" acceptable to label this button as? The defect doesn't really say what the button should be called. It just says that the button should not be unlabeled. I more or less just followed the pattern of naming and this made sense to me. 
https://rally1.rallydev.com/#/57732444928d/dashboard?detail=%2Fdefect%2F614612924875&fdp=true

semibran:
> Is the string "Rubric Statistics" acceptable to label this button as?

counting aria attributes it'll read out as `clickable Rubric Statistics button` which is probably fine -- were you thinking of adding a langterm like `Open rubric statistics popout`? Not sure. Could be nice

jesseag:
I didn't have plans to add that in a langterm like that. I think the name "Rubric Statistics" should be okay for that button. I did a bit more digging and there is a duplicate defect to this one (filed by Mark) and this is what he calls the button. If necessary, I can look into the langterm but I feel like it would be out of the scope of this defect.

PR #1395, opened 2021-12-03
Add null check to criterion delete action

semibran:
should fix this stuff -- I was missing a check on the delete action before performing. The other actions in this file seem to be fine.

image
(sc: Devon)


semibran:
I guess we're actually expecting it to error. We don't really have a good way to get the delete handler to throw unlike the save handling where we can just provide invalid input values. Could provide a more descriptive error message? But this isn't supposed to show up in prod anyway. There's no precedent for non-default/non-localized error messages in rubrics either.

PR #1398, opened 2021-12-03
Switch to `styleMap` for CSS generation

devpow112:
Helps with validity. A bit nicer then manually generating strings of CSS.


semibran (approved):
pretty neat stuff

devpow112:
> pretty neat stuff

Yeah saw it when looking at something in core. They also have a `classMap` for easy toggling of classes
https://lit.dev/docs/components/styles/#dynamic-classes-and-styles

semibran:
> @semibran yeah saw it when looking at something in core. They also have a `classMap` for easy toggling of classes https://lit.dev/docs/components/styles/#dynamic-classes-and-styles Might be useful somewhere to help clean something up.

Oh true. `classMap` looks nice, I was actually trying to get `classcat` to work with our setup a while back but I guess our eslint doesn't play nice with multilining objects in attribute defs. We'd probably run into similar problems here :cry:

PR #1401, opened 2021-12-03
Clean-up `d2l-rubric-visibility-editor`

devpow112:
Clean-up a few things in `d2l-rubric-visibility-editor`. Was fixing some issues with styling of the options components and got a little carried away. I'm definitely gonna clean-up the `async`/`await` stuff elsewhere. I split this into 4 commits since doing it all together is unreadable.

* Split out render functions (45c38ec9790186de285c037273ce3f6b27c50162)
* Convert to `async`/`await` (25e70b380b3caa5040cc8196cf78a93d6c8456e3)
* Clean-up entity action getting (e62c48cf8a1309367c5aa23b84d76e2b05728314)
* Rename function for better code grouping (036e0b82668acbc954d6569e0d51315cf3e8ca81)


src/components/editor/rubric-visibility-editor.js
```diff
@@ -229,24 +209,73 @@ class RubricVisibilityEditor extends composeMixins(
 		}
 	}
```
semibran, line 210: This is more of an architectural nit but I feel it'd be better to relocate this kind of HM entity unpacking logic into the actual entity classes at some point. Probably depends on the original intention with those are...... we haven't really discussed this stuff at length though. I've been envisioning the entity classes as interfaces around the entity data to abstract that out of the actual component state and rendering logic. I guess doing things this way does let us tap into lit observers. It kind of seems like a lot though? Interested in getting some more perspective here

Or maybe we're on the same page and I'm just getting confused by the diffs. Lol

devpow112: I think I'm on the same page as you. A bunch of the logic in here can probably go into the entity classes. I would still want to pull some stuff into lit properties that influence rendering but some of the stuff around actually getting fields and properties could be moved out of here eventually into the entity classes. Maybe we can talk about this specific example a bit more on Monday so we can make sure we are indeed on the same page

semibran (approved):

PR #1402, opened 2021-12-03
Clean-up `d2l-rubric-alignment-editor`

devpow112:
Similar clean-up to #1401. Also pulled some common CSS properties out for use in the other parts of the options.
- Clean-up rendering (2fa9492679881dc101ece055c9da9f69fd09cdcd)
- Clean-up entity action getting (e4456758294cbd7c2ffb8d17940ba24a23a31178)
- Convert to `async`/`await` (64e4cfe1c331db59e081fa16a1488cbb33b49827)


src/components/editor/rubric-editor.js
```diff
@@ -317,6 +317,10 @@ class RubricEditor extends composeMixins(
 				d2l-organization-availability-set {
 					padding-top: 0.5rem;
 				}
+
+				#alignment-container {
```
devpow112, line 321: This makes it so when the alignment editor is not rendering anything the section it's in is empty with no height. Before this would be a block and cause unneeded spacing.

PR #1404, opened 2021-12-06
Remove need for eslint global config for `fastdom`

devpow112:


devpow112:
NM some browsers hate this.

PR #1413, opened 2021-12-07
`d2l-rubric-levels-mobile` conversion US132984

imcdonald-d2l:
There are still some lingering issues with `d2l-rubric-overall-score-mobile` that I'll clean up with that conversion.. haven't gotten to the bottom of those yet. It's likely that there are some other defects hiding in this conversion but putting up a PR as I have cleaned up all the ones I can find. 

This file was quite large so I probably missed some obvious code clean-up items as well. Just point them out if you see anything that can be refactored.

https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F612450179531


devpow112:
I think the lit error is this opne
```
    Type 'string' is not assignable to 'boolean'
    887:  ?aria-checked="${this._getCriter
    no-incompatible-type-binding
```
Unfortunately it uses color to distinguish error from warning. I'll see if I can fix that.

src/components/rubric-levels-mobile.js
```diff
@@ -0,0 +1,1037 @@
...
+			hovered: this._isHovered(index, hovered),
+			focused: this._isFocused(index, focused) && (this.readOnly || selected === -1),
+			dragging: this._currentDragContext?.elementHasBeenDragged,
+			moving: this._currentDragContext?.elementHasBeenDragged || moving
+		});
+	}
+
+	_getLevelDescription(cell) {
+		const cellEntity = cell && cell.getSubEntityByClass && cell.getSubEntityByClass(Classes.rubrics.description);
+		return cellEntity && cellEntity.properties.text || this.localize('noDescription');
```
semibran, line 474:
```suggestion
		const cellEntity = cell?.getSubEntityByClass?.(Classes.rubrics.description);
		return cellEntity?.properties?.text || this.localize('noDescription');
```
or use `RubricCriterionCellEntity.descriptionText`

src/components/rubric-levels-mobile.js
```diff
@@ -0,0 +1,1037 @@
...
+	}
+
+	_getLevelIndex(index, levelCount) {
+		return getComputedStyle(this).dir === 'rtl'
+			? levelCount - index
+			: index + 1;
+	}
+
+	_getLevelName(level) {
+		return level && level.properties && level.properties.name;
```
semibran, line 488:
```suggestion
		return level?.properties?.name;
```
or use `RubricLevelEntity.name`

src/components/rubric-levels-mobile.js
```diff
@@ -0,0 +1,1038 @@
...
+	}
+
+	disconnectedCallback() {
+		this.removeEventListener('mouseup', (e) => this._handleDragEnd(e));
+		super.disconnectedCallback();
+	}
+
+	firstUpdated() {
+		this._handleResize();
+		new ResizeObserver(this._handleResize.bind(this)).observe(this.shadowRoot.host);
```
semibran, line 295: pros/cons of `ResizeObserver` vs. `d2l-resize-aware`?

src/components/rubric-levels-mobile.js
```diff
@@ -0,0 +1,1038 @@
...
+	getSelfLink(entity) {
+		return entity?.getLinkByRel('self')?.href || '';
+	}
+
+	_canScrollLeft(scroll) {
+		return scroll > 0;
+	}
+
+	_canScrollRight(scroll, visibleWidth) {
+		const scrollContainer = this.shadowRoot && this.shadowRoot.querySelector('.levels-container');
```
semibran, line 388:
```suggestion
		const scrollContainer = this.shadowRoot?.querySelector('.levels-container');
```
I'm not sure what scenario could have the shadow root resolve as nullish. Doesn't hurt to be sure I guess

src/components/rubric-levels-mobile.js
```diff
@@ -0,0 +1,1038 @@
...
+			? this.localize('selectPreviousLevel')
+			: this.localize('selectNextLevel');
+	}
+
+	_getScore(level, cell) {
+		const levelScore = level && level.properties && level.properties.points;
+		if (!isNaN(levelScore)) {
+			return levelScore;
+		}
+		const cellScore = cell && cell.properties && cell.properties.points;
```
semibran, line 557:
```suggestion
		const levelScore = level?.properties?.points;
		if (!isNaN(levelScore)) {
			return levelScore;
		}
		const cellScore = cell?.properties?.points;
```
These should be available from the entities as well. Not sure if we can give this function a better name or (in the future) write a more comprehensive abstraction around HM levels and cells that isn't so tightly linked to this particular webcomponent.

src/components/rubric-levels-mobile.js
```diff
@@ -0,0 +1,1038 @@
...
+		}
+	}
+
+	_getScoreLabel(score, outOf) {
+		return !isNaN(score) && !isNaN(outOf) && this.localize('scoreOutOfAria', 'score', score, 'outOf', outOf);
+	}
+
+	_getSelfLink(entity) {
+		return entity?.getLinkByRel('self')?.href || '';
+	}
```
semibran, line 569: duplicate of L379-381?

src/components/rubric-levels-mobile.js
```diff
@@ -0,0 +1,1038 @@
...
+		const index = this._selected === -1 ? 0 : this._selected;
+		const radio = this._getRadioAt(index);
+		if (radio) {
+			radio.focus();
+		}
+	}
+
+	getSelfLink(entity) {
+		return entity?.getLinkByRel('self')?.href || '';
+	}
```
semibran, line 381: Should be handled by `EntityBase` already

src/components/rubric-levels-mobile.js
```diff
@@ -884,7 +880,7 @@ class RubricLevelsMobile extends composeMixins(
 				class="${this._getLevelClassName(index, this._selected, this._hovered, this._focused, this._animating)}"
 				dataIndex="${index}"
 				cellHref="${this._getCriterionCellHref(this.criterionCells, index)}"
-				?aria-checked="${this._getCriterionCellAriaChecked(index, this._selected)}"
+				?aria-checked="${index === this._selected}"
```
devpow112, line 883: This should probably be. I've seen this in other spots done this way.
```
aria-checked="${this._getCriterionCellAriaChecked(index, this._selected)} ? 'true' : 'false'"
```

devpow112: Yeah same..ish error
```
    You are using a boolean binding on a non boolean type '"true" | "false" | "mixed" | "undefined"'
    883:  ?aria-checked="${index === this._selected}"
    no-incompatible-type-binding
```

src/components/rubric-levels-mobile.js
```diff
@@ -0,0 +1,1038 @@
...
+				@transitionend="${this._handleThumbTransitionEnd}"
+			>
+				<d2l-icon class="check-icon" icon="d2l-tier1:check"></d2l-icon>
+			</span>
+		`;
+	}
+
+	_say(text) {
+		const statusWrap = this.shadowRoot.querySelector('.criterion-status');
+		const status = statusWrap && statusWrap.querySelector('p');
```
semibran, line 934:
```suggestion
		const status = this.shadowRoot.querySelector('.criterion-status p');
```
I actually added this before I learned about `@brightspace-ui/core/helpers/announce`. So we could probably scrap this whole function and replace it with the core helper. Only possible downside to discuss is whether inline rubrics can be opened in `d2l-dialog` which would still break aria-live afaik.

src/components/rubric-levels-mobile.js
```diff
@@ -0,0 +1,1034 @@
...
+			|| (getComputedStyle(this).direction === 'rtl'
+				? index > 0
+				: index < levels.length - 1);
+	}
+
+	_isSelected(index, selected) {
+		const isSelected = selected >= 0 && index === selected;
+		if (isSelected) {
+			const selectedLevel = this.assessmentMap[this._getSelfLink(this.criterionCells[index])];
+			if (selectedLevel && selectedLevel.properties && selectedLevel.properties.score) {
```
semibran, line 802:
```suggestion
			if (selectedLevel?.properties?.score) {
```
or use `RubricLevelEntity.score`

src/components/rubric-levels-mobile.js
```diff
@@ -0,0 +1,1034 @@
...
+	}
+
+	_handleBlur() {
+		if (!this._animating) {
+			this._updateFocused(-1);
+		}
+	}
+
+	_handleDrag(e) {
+		if (!this._hasDragStarted || this.readOnly) {
```
semibran, line 599: Still a bit unsatisfied with this var name. Something like `_hasDragStarted` -> `_isMousePressed`? Might not be input-agnostic enough for touch devices. Not sure what else would work here - I feel like the current name has too much overlap with `elementHasBeenDragged` though.

src/components/rubric-levels-mobile.js
imcdonald-d2l: Honestly, not familiar with either of them so I couldn't tell you 🤷🏻‍♂️

imcdonald-d2l: I know `ResizeObserver` is a default web API and supported on all browsers we support.

src/components/rubric-levels-mobile.js
imcdonald-d2l: Just accessed these properties from the HM entity definitions instead. I kept `level?.properties?.points` though since default level entity behavior is to return `0` if points value doesn't exist... can't remember why I did that though when I wrote the entity `get points()` accessor. 🤔   Doesn't look like we want that in this scenario.

src/components/rubric-levels-mobile.js
imcdonald-d2l: Yup, good catch! 👍🏻

src/components/rubric-levels-mobile.js
imcdonald-d2l: I'm pretty sure it can't be opened in a dialog as far as I'm aware... 🤔 
I'll make your suggested change and we can circle back to this if other people are confident this can't be in a dialog as well.

src/components/rubric-levels-mobile.js
imcdonald-d2l: Agreed, `_hasDragStarted` is not an accurate property name for the state it is tracking and certainly overlaps with `elmentHasBeenDragged`. I swapped it for `_isMousePressed` as it represents the state I am tracking. As for touch devices... haven't tested if drag even works with those types of devices. If that's something we want to fix we can add that logic and re-look at variable names at that point IMO.

src/components/rubric-levels-mobile.js
devpow112: Can't this just use https://github.com/Brightspace/d2l-rubric/blob/master/src/entities/entity-base.js#L129

src/components/rubric-levels-mobile.js
devpow112: We should probably start trying to convert to use the native resize observer of we can over the custom component resize stuff. Just less dependencies

mpharoah-d2l: Just make sure to test thoroughly on Safari

mpharoah-d2l: `d2l-resize-aware` is a component I made a while ago because the existing resize observer polyfills failed to account for the shadow dom boundaries and would fail with nested web components. Now that all our supported browsers should natively support ResizeObserver, we can probably switch over to the native version, but we'll have to be careful to test all supported browsers.

mpharoah-d2l: More information: https://github.com/BrightspaceUI/resize-aware#browser-support

src/components/rubric-levels-mobile.js
devpow112: You can also just do `entity?.self()` I believe
https://github.com/BrightspaceHypermediaComponents/siren-sdk/blob/master/src/es6/Entity.js#L28

imcdonald-d2l: So this is used in contexts where an entity definition isn't yet defined. For example:

- `this._getSelfLink(this.criterionCells[index])` `criterionCells` isn't attached to any entity definition. It is currently being passed from a polymer component.
- this._getSelfLink(assessmentEntity)` `assessmentEntity` is also not attached to an entity definition.

I could just do `new Entity(assessmentEntity).getSelfLink()` and `new Entity(this.criterionCells[index]).getSelfLink()` if we think that is a better approach 👍🏻

devpow112: Ah no that's fine. Missed that. Assumed there was an attached entity.

imcdonald-d2l:
So this doesn't seem right... is this a timing thing with vdiff @devpow112 @MitchelPaulin? 🤔 

image

test/perceptual/d2l-rubric-criterion-mobile-student.visual-diff.js
```diff
@@ -1,60 +0,0 @@
...
-			kinds: ['custom-points', 'no-score', 'points']
-		}, {
-			type: 'holistic',
-			kinds: ['no-score', 'percentage']
-		}];
-
-		for (const configuration of configurations) {
-			for (const kind of configuration.kinds) {
-				const id = `#criterion-mobile-${configuration.type}-${kind}`;
-				it(`select ${configuration.type} ${kind}`, async function() {
```
imcdonald-d2l, line 39: The whole purpose of the vdiffs in this file was to get select and focus states on the levels-mobile component. This is no longer possible because `selectLevelandFocus` utility relies on public `select` and `hover` properties. These are private now so we can't do that anymore. In order to get select and focus state set we need an assessment map and assessment entity which we can't do right now 😢

semibran: Lit attributes aren't "hard-private" are they? you should still be able to access those attributes via `_selected` etc. as long as they're still on the element. Not sure if we got around setting up the assessment data in full yet. At the very least I think deleting the tests altogether might be a little extreme when we can just describe.skip

mpharoah-d2l: Yeah, you should still be able to access the "private" properties. They are still on the HTML element as object properties.

devpow112: The `_` should just a notational thing.

imcdonald-d2l: I'll try setting the private properties then to see if we can get the vdiffs working then 👍🏻

devpow112:
Unit test failure was some weird intermittent script error on safari. Seen it a few times over the last month or so.

test/perceptual/d2l-rubric-criterion-mobile-student.visual-diff.js
imcdonald-d2l: So trying to set the private properties does not seem to fix the vdiff tests. It either isn't setting them properly or isn't triggering a re-render. Hard to say for sure since my local vdiffs aren't wanting to work for me right now so I can't debug 😞

test/perceptual/utilities.js
```diff
@@ -1,4 +1,4 @@
-/* global querySelectorShadowDom */
+const querySelectorShadowDomPath = '/node_modules/query-selector-shadow-dom/dist/querySelectorShadowDom.js';
```
imcdonald-d2l, line 1: This diff is getting stuff PR: https://github.com/Brightspace/d2l-rubric/pull/1419
Not sure why, maybe some rebasing issue 🤷🏻‍♂️

devpow112 (approved):

devpow112:
I think I mentioned this but make sure to squash this one when merging. It's got a lot of commits so squashing is probably best.

PR #1417, opened 2021-12-08
Note that we must use version `6` of vdiff locally to work.

devpow112:
Version `7` is a mess and won't work for us right now.


devpow112:
We also can't update to the latest version of vdiff in CI as it will exhibit the same issues.

MitchelPaulin (approved):

PR #1421, opened 2021-12-09
Us132978 rubric criteria group

MitchelPaulin:


src/components/rubric-criteria-group.js
```diff
@@ -0,0 +1,1096 @@
...
+				styles.push(`border-${side}: 1px solid var(--d2l-color-mica)`);
+				if (loaLevelEntity.levelHref === rubricLevelHref) {
+					styles.push(`border-${side}-width: 2px`);
+					styles.push(`border-${side}-color: ${loaLevelEntity.color}`);
+				}
+			}
+		}
+
+		return styles.join(';');
+	}
```
semibran, line 476: Wonder if we could coalesce this with [criterion-editor](https://github.com/Brightspace/d2l-rubric/blob/master/src/components/editor/rubric-criterion-editor.js#L729-L761) into some shared styling construct as part of one of our post-conversion items

src/components/rubric-criteria-group.js
```diff
@@ -0,0 +1,1096 @@
...
+
+		return null;
+	}
+	_getNextRubricLevel(allRubricLevels, rubricLevelEntity) {
+		const nextHref = rubricLevelEntity.nextHref;
+		return this._resolveRubricLevel(allRubricLevels, nextHref);
+	}
+	_getOutcomesTitleText() {
+		return D2L?.Custom?.Outcomes?.TermTitleText;
+	}
```
semibran, line 541: Don't need any action around this right now per se but I guess this was one potential (hacky) workaround for the excess propagation stuff - `outcomesTitleText` gets loaded in via the rubric renderer and [set as a global component-side](https://github.com/Brightspace/d2l-rubric/blob/master/src/d2l-rubric.js#L753-L764). Targeting this atm as part of the context work

src/components/rubric-criteria-group.js
```diff
@@ -0,0 +1,1096 @@
...
+		if (!this._loaded) {
+			const entityId = this._getSelfLink(entity);
+			this.logViewRubricEvent({ id: entityId });
+			this._loaded = true;
+		}
+	}
+	_queueChange(entityGetter) {
+		// TODO, when updating the parent component change it to use a more specialized entity class
+		const entity = entityGetter();
+		const criterionHref = entity.getLinkByRel('https://rubrics.api.brightspace.com/rels/criterion')?.href;
```
semibran, line 781: nit: in HM constants as `Rels.Rubrics.criterion`

src/components/rubric-criteria-group.js
```diff
@@ -0,0 +1,1096 @@
...
+		const levelHref = entity.getLinkByRel('https://rubrics.api.brightspace.com/rels/level')?.href;
```
semibran, line 782: nit: in HM constants as `Rels.Rubrics.level`

src/components/rubric-criteria-group.js
```diff
@@ -0,0 +1,1096 @@
...
+			this.optimisticUpdateDispatcher.rollback(criterionHref);
+			this._optimisticSelections.delete(criterionHref);
+			this._optimisticSelectionsVersion++;
+			throw err;
+		});
+	}
+	_queueRadio(radio) {
+		const href = radio.getAttribute('data-href');
+		const entity = this.cellAssessmentMap[href];
+		const criterionHref = entity?.getLinkByRel('https://rubrics.api.brightspace.com/rels/criterion')?.href;
```
semibran, line 809: nit: in HM constants as `Rels.Rubrics.criterion`

src/components/rubric-criteria-group.js
```diff
@@ -0,0 +1,1096 @@
...
+	_renderCriterion(criterion, criterionNum) {
+
+		let rowHeader = null;
+		if (this._showRowHeaders(this.rubricType)) {
+
+			let competenciesIcon = null;
+			if (this._showCompetencies(this.criterionResultMap, criterion, this.readOnly)) {
+				competenciesIcon = html`
+					<d2l-rubric-competencies-icon
+						competency-names="${this._getCriterionCompetencies(this.criterionResultMap, criterion)}"
```
semibran, line 823:
```suggestion
						.competencyNames="${this._getCriterionCompetencies(this.criterionResultMap, criterion)}"
```
This'll propagate as `[object Object]` otherwise

src/components/rubric-criteria-group.js
```diff
@@ -0,0 +1,1096 @@
...
+						.token="${this.token}"
+						?read-only="${this.readOnly}"
+						?enable-feedback-copy="${this.enableFeedbackCopy}"
+						criterion-num="${criterionNum}"
+						@save-feedback-start="${this._handleSaveStart}"
+						@save-feedback-finished="${this._handleSaveFinished}"
+						@close-feedback="${this._closeFeedback}"
+						@focus="${this._handleFeedbackTextFocus}"
+						?use-new-brightspace-editor="${this.useNewBrightspaceEditor}"
+						focus-feedback-html-editor="${this._isCriterionFeedbackFocused(criterionNum, this._addingFeedback, this._focusFeedbackHtmlEditor)}"
```
semibran, line 918: I'm not completely clear on what the two-way data binding in the legacy component was originally used for (traces back to #976) but if we're just trying to 1:1 it for now we may have to hook into [`focus-feedback-html-editor-changed`](https://github.com/Brightspace/d2l-rubric/blob/master/src/components/rubric-feedback.js#L411-L413)

src/entities/rubric-criterion-cell-entity.js
```diff
@@ -63,4 +63,8 @@ export class RubricCriterionCellEntity extends EntityBase {
 		return this.properties?.points;
 	}
 
+	get upHref() {
+		return this._entity?.getLinkByRel('up')?.href;
+	}
+
```
semibran, line 69: thoughts on adding this to the entity base?

src/components/rubric-criteria-group.js
semibran: probably isn't super important to address all these right away since I think we're still gonna do some work around getting the entity mappings to resolve to actual entity instances as we move further up the tree

MitchelPaulin: Yeah, thats the plan currently, I am going to change all these to use the entity getters when I tackle the groups component next

src/components/rubric-criteria-group.js
devpow112: This is also probably a good place to use the `styleMap` stuff
https://lit.dev/docs/templates/directives/#stylemap
Did something similar in https://github.com/Brightspace/d2l-rubric/blob/master/src/components/editor/rubric-criterion-editor.js#L729-L761

src/entities/rubric-criterion-cell-entity.js
MitchelPaulin: Could be worth it, do all entities have an up link?

src/components/rubric-criteria-group.js
```diff
@@ -0,0 +1,1096 @@
...
+		return this.localize('rubricCellAriaLabel', 'criterionName', criterionName, 'levelName', levelName, 'cellDescription', cellDescription);
+	}
+	_getFeedbackID(criterion, criterionResultMap, index) {
+		if (this._hasFeedback(criterion, criterionResultMap)) {
+			return `feedback${index}`;
+		}
+	}
+	_getHeaderStyle(loaLevel, levelsReversed) {
+		const side = levelsReversed ? 'left' : 'right';
+		return [
```
devpow112, line 512: See comment about `stylemap`
https://github.com/Brightspace/d2l-rubric/pull/1421/files/165f02d3865853fdd1e8bfc7298b4a549348b746#r766028513

src/components/rubric-criteria-group.js
devpow112: Ah yes @semibran would be good if we just use the same chunk of code.

src/components/rubric-criteria-group.js
MitchelPaulin: So this ones a bit weird, if you look at the original component the two way binding is done to a function call similar to the call you see here, from what I understand that means the two way binding had no effect in the other component

src/components/rubric-criteria-group.js
```diff
@@ -0,0 +1,1096 @@
...
+		this._criteriaCollectionHref = entity.criteriaLink;
+		// The first time that the entity loads, we send out an Open event
+		if (!this._loaded) {
+			const entityId = this._getSelfLink(entity);
+			this.logViewRubricEvent({ id: entityId });
+			this._loaded = true;
+		}
+	}
+	_queueChange(entityGetter) {
+		// TODO, when updating the parent component change it to use a more specialized entity class
```
MitchelPaulin, line 779: Going to address most of the naked `getLinkByRel` calls when I convert the groups component

src/entities/rubric-criterion-cell-entity.js
semibran: Only certain entities have one. It's not uncommon though which makes me think if we end up implementing this function multiple times for certain entity types specifically we might as well implement it generically.

MitchelPaulin: Doesn't look super common 
image

I may just remove this function and use `getLinkByRel` directly since its seems to be what we are doing in other converted components

PR #1422, opened 2021-12-09
De46258/text overlap copy delete row button

jesseag:
Original defect with pictures of error where text overlaps more button: 
https://rally1.rallydev.com/#/57732444928d/dashboard?detail=%2Fdefect%2F618604236235&fdp=true

Fixed defect:
image

To adjust the width, I needed to change the position: relative to absolute. Then just tweaked the percentages until I got the text box to closely line up  with minimal spacing between the More (...)/Standards button(s).


MitchelPaulin:
I don't think this solution will work as is, you'll need to make the styles conditionally apply based on [the can _canShowMoreMenu variable](https://github.com/Brightspace/d2l-rubric/pull/1422/files#diff-43346811d3ee56b51898c1e5cbf25f345f5783143318752728930f5a409f8d93L802) otherwise the input box will be made smaller even if the dropdown menu isn't there.

jesseag:
I took Line 802 and swapped the true/false conditions to see what would happen if this._canShowMoreMenu was not there. 
${this._canShowMoreMenu ? null : this._renderCriterionMoreMenu()}

https://github.com/Brightspace/d2l-rubric/blob/master/src/components/editor/rubric-criterion-editor.js#L802

image

It looks like the text box still gets rendered and the size is not changed even if the More menu is not there. However, this breaks the Visual Diff tests so I should probably look for another solution.

MitchelPaulin:
This is wrong though, the input box should take up the width of the cell when the dropdown menu isn't rendered

jesseag:
Fair enough, this makes sense that this is the desired behaviour.

New change: 

image

Swapping true/false conditions on Line 802 to remove the More(...) button.

image

MitchelPaulin (approved):

jesseag:
I'm not going to merge this quite yet. I discovered another defect where writing a bunch of text into the text box goes over top of the Standards button. I talked to Mark and I'm going to fix this as part of the defect as well. 

image

jesseag:
Talked to Andrew and we are actually going to split this up into 2 defects for clarity/since it is sprint end today. 
Once CI tests run for latest master changes I will merge it.

PR #1425, opened 2021-12-09
`d2l-rubric-overall-score-mobile` conversion - US133001

imcdonald-d2l:
https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F612452551675


src/components/rubric-overall-score-mobile.js
```diff
@@ -0,0 +1,231 @@
...
+				<span>${this.localize('overallScore')}</span>
+			</div>
+			<d2l-rubric-levels-mobile
+				.href="${this.href}"
+				.token="${this.token}"
+				assessment-href="${this.assessmentHref}"
+				.assessmentMap="${this._assessmentMap}"
+				name="${this.localize('overallScore')}"
+				?read-only="${this.readOnly}"
+				.onSelect="${this._selectCell}"
```
semibran, line 102: does `@select` work here?

src/components/rubric-overall-score-mobile.js
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	_getClearOverrideButtonDescription() {
+		const criterionName = this.localize('overallScore');
+
+		return this.localize('clearOverrideFor', 'criterionName', criterionName);
+	}
+
+	_getVisibleLevel() {
+		return this._levels && this._levels[this._levelVisible];
```
semibran, line 136:
```suggestion
		return this._levels?.[this._levelVisible];
```

src/components/rubric-overall-score-mobile.js
```diff
@@ -0,0 +1,231 @@
...
+		}, {});
+	}
+
+	_renderCompetencies() {
+		if (!this._areCompetenciesVisible()) {
+			return;
+		}
+		return html`
+			<d2l-rubric-competencies-icon
+				competency-names="${this._competencies}"
```
semibran, line 183:
```suggestion
				.competencyNames="${this._competencies}"
```

src/components/rubric-overall-score-mobile.js
devpow112, line 102: Should this be am event handler?
```
@select="${this._selectCell}"
```

src/components/rubric-overall-score-mobile.js
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	_selectCell(entityGetter, callback) {
+		if (!this.readOnly) {
+			this.OverallLevelAssessmentHelper.selectOrClearAsync(entityGetter).then(
+				() => {
+					if (callback) {
+						callback();
+					}
+				}
```
semibran, line 225:
```suggestion
				() => callback?.()
```

src/components/rubric-overall-score-mobile.js
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	_getVisibleLevelDescription() {
+		const level = this._getVisibleLevel();
+		return typeof level?.properties.rangeStart === 'number' && this.localize('pointsMinimum', 'number', level?.properties.rangeStart.toString()) || '';
+	}
+
+	_getVisibleLevelName() {
+		const level = this._getVisibleLevel();
+		const levelName = level && level.properties.name;
```
semibran, line 146:
```suggestion
		const levelName = level?.properties.name;
```

src/components/rubric-overall-score-mobile.js
devpow112: Oh wait I see what this is trying to do. You are trying to pass a function down into a component. Not sure if that will work. You may just have to switch to event driven stuff.

semibran: I remember the initial intent when I was writing this was to use the same levels component for both criteria and the overall score but use different click handlers for each so they could trigger different actions. Passing down the callback worked fine in polymer but not sure what the analog would be here.... I wanted to avoid disambiguating parent types from inside levels-mobile but I'm not sure we have a choice anymore

devpow112: I think the analog would be event firing. So instead of passing the callback down you would fire and event from the lower component and have an event handler in the higher component. I'm surprised that worked in polymer. Must have been how they resolve properties and event handlers allowed this to happen.

semibran: Okay yeah of course. Was locked into the idea that the levels component itself had to perform the selection logic directly for some reason. That should work here as well

imcdonald-d2l: Passing the callback in lit element seems to be working just fine locally. I've debugged locally and it is hitting the callback function as expected. So don't think we need to switch to even firing in this case.

imcdonald-d2l: I can re-name the property in `levels-mobile` to see if it takes care of linting.

devpow112: We might not need to if it works I guess. Still feels like event firing is the better way to have this set-up but that might just be me. Maybe it's a later thing. If we aren't going to change it we'll have to figure out if there is a change we can make to make the lit linter happy or just disable that rule for this instance.

devpow112: You can laos probably remove the type and mark it as `{attribute: false}` since it isn't really an attribute.

semibran: interesting -- though it'd still probably be better to change it to something more orthodox in the future just as a matter of consistency with how we're already doing things elsewhere

imcdonald-d2l: If this is something we want to change I would recommend waiting until `d2l-rubric-criterion-mobile` is converted (if we can get around linting). It is a parent of `d2l-rubric-levels-mobile` so better to convert the two parents before trying to adjust the interface around the function properties.

imcdonald-d2l: @devpow112 I don't know if we can add `{attribute: false}` just yet since `d2l-rubric-criterion-mobile` polymer component is passing a function through this attribute.

imcdonald-d2l: Linter seems happy 🎉

PR #1426, opened 2021-12-09
Ran npm format

MitchelPaulin:
This is just the result of running `npm run format`


devpow112:
This looks like it juust changes the end lines for styles to remove white-space. That doesn't trigger the linter. Not sure if we want to merge this since it changes the indenting. Let me see if I can make stylelint not care about that last line indent.

MitchelPaulin:
> This looks like it just changes the end lines for styles to remove white-space. That doesn't trigger the linter. Not sure if we want to merge this since it changes the indenting. Let me see if I can make stylelint not care about that last line indent.

I agree that changing the rule would be preferable. I do want to merge this if its not though since I use `npm run format` pretty often to clean up my conversions and it would be nice if I didn't have these files change when I did

devpow112:
> > This looks like it just changes the end lines for styles to remove white-space. That doesn't trigger the linter. Not sure if we want to merge this since it changes the indenting. Let me see if I can make stylelint not care about that last line indent.
> 
> I agree that changing the rule would be preferable. I do want to merge this if its not though since I use `npm run format` pretty often to clean up my conversions and it would be nice if I didn't have these files change when I did

Yeah agreed. I can't get it to leave that last line alone. Let's merge this for now and If I figure out how to fix it I'll just put up a PR.

devpow112 (approved):

PR #1427, opened 2021-12-09
Update stylelint config to not remove indent on final css lines

devpow112:
For some reason disabling `no-eol-whitespace` still removed trailing whitespace but it leave the last line alone. Feels a bit weird to disable this but it works the way we'd like and it still throws errors when linting if thee is trailing white-space.


devpow112:
@MitchelPaulin can you give this a try and see if it works as expected for you? Just want to make sure there isn't something weird in my set-up that's causing it to behave well.

MitchelPaulin (approved):
Works for me, nothing is changed when running `npm run format`

PR #1428, opened 2021-12-10
Add assessment context logic for flags and langterms

semibran:
Would've like to add more than just flags and langterms (rubric name, type, scoring, text only, compact) so we don't have to extract that data from the rubric multiple times but I didn't want to conflict too much with in progress conversion tickets. There's already some overlap with the criteria-groups conversion going on so wanted to keep that to a minimum.


src/d2l-rubric.js
```diff
@@ -540,7 +542,14 @@ const component = {
...
 	ready: function() {
-		this._updateOutcomesTitleText();
+		this._context = {
+			...this._context,
+			isFeedbackCopyEnabled: this.enableFeedbackCopy,
+			isNewHtmlEditorEnabled: this.useNewBrightspaceEditor,
+			langterms: {
+				outcomes: this.outcomesTitleText,
+			},
+		};
```
semibran, line 552: context definition - pretty barebones but we can extend it later

devpow112 (approved):
Looks good to me :tada:

PR #1439, opened 2021-12-13
`d2l-rubric-criterion-mobile` conversion - US132982

imcdonald-d2l:
https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F612450173975


src/components/rubric-editable-score.js
```diff
@@ -370,7 +370,7 @@ class RubricEditableScore extends composeMixins(
 
 		this._overriddenStyling = this._shouldShowOverrideState;
 
-		this._score = score;
+		this._score = score ? score : null;
```
imcdonald-d2l, line 373: Was getting `NaN` here so added this check.

src/components/rubric-levels-mobile.js
```diff
@@ -828,11 +828,13 @@ class RubricLevelsMobile extends composeMixins(
 		if (this.readOnly) return;
 		const entityGetter = () => this.assessmentMap[entityHref];
 		this._pendingChanges++;
+		this.dispatchEvent(new CustomEvent('d2l-siren-entity-save-start', { bubbles: true, composed: true }));
```
imcdonald-d2l, line 831: This is necessary so we get the `Saving...` in the top of the criterion mobile view.

devpow112: We fire that event in here as well but I'm guessing this code isn't using that
https://github.com/Brightspace/d2l-rubric/blob/master/src/mixins/rubric-siren-action-mixin.js

imcdonald-d2l: It's using this guy [here](https://github.com/Brightspace/d2l-rubric/blob/master/src/mixins/assessment-mixin.js#L19). Now that I think about it again, I should just add the dispatchers before and after `performLazySirenAction`

devpow112:
Looks like something is up with the test utils maybe?
https://github.com/Brightspace/d2l-rubric/runs/4511155832?check_suite_focus=true#step:5:166

src/components/rubric-editable-score.js
MitchelPaulin: `0 ? 0 : null` will be null, I think this may cause problems if setting the score to 0

src/components/rubric-levels-mobile.js
imcdonald-d2l: Moved the emitters on [this commit](https://github.com/Brightspace/d2l-rubric/pull/1439/commits/a62e43f5e85833ab19f475894e7587a5ae1d8974).

test/perceptual/d2l-rubric-criterion-mobile-student.visual-diff.html
```diff
@@ -23,59 +23,54 @@ <h2>d2l-rubric-criterion-mobile student</h2>
 		<h3>Analytic custom points criterion</h3>
 		<div class="visual-diff" id="criterion-mobile-analytic-custom-points">
 			<d2l-rubric-criterion-mobile
-				compact
```
MitchelPaulin, line 26: Was there a reason why we had to remove `compact` here?

imcdonald-d2l: It's no longer a property on `d2l-rubric-criterion-mobile`, it was always set to true so I just removed the property entirely.

src/components/rubric-criterion-mobile.js
```diff
@@ -0,0 +1,429 @@
...
+
+	_renderDescription() {
+		if (!this._currentLevel || !this._criterionCells?.[this._currentLevel]?.descriptionHtml) {
+			return;
+		}
+		return html`
+			<div class="criterion-description">
+				<p class="criterion-description-html">
+					<d2l-html-block>
+						<template id="criterion-description-html">${unsafeHTML(this._criterionCells?.[this._currentLevel]?.descriptionHtml)}</template>
```
MitchelPaulin, line 337: Whats the reason for unsafeHtml here?

src/components/rubric-criterion-mobile.js
```diff
@@ -0,0 +1,429 @@
...
+		this._slide(1);
+	}
+
+	_updateCurrentScore() {
+		// prevent updates to the editable score component if the user is in the process of editing it
+		if (this._isEditingCriterionScore) {
+			return;
+		}
+
+		this._editableScore = this._hoveredScore ? this._hoveredScore : (this._score || this._selectedScore);
```
MitchelPaulin, line 418: This line feels like it will cause issues when the scores are zero

src/components/rubric-criterion-mobile.js
imcdonald-d2l: We use this everywhere we render html. See more on the directive [here](https://lit.dev/docs/templates/directives/#unsafehtml).

src/components/rubric-criterion-mobile.js
imcdonald-d2l: That's a good point, I'll adjust the condition to account for when `_hoveredScore === 0`

src/components/rubric-criterion-mobile.js
```diff
@@ -0,0 +1,429 @@
...
+		this._slide(1);
+	}
+
+	_updateCurrentScore() {
+		// prevent updates to the editable score component if the user is in the process of editing it
+		if (this._isEditingCriterionScore) {
+			return;
+		}
+
+		this._editableScore = this._hoveredScore || this._hoveredScore === 0 ? this._hoveredScore : (this._score || this._selectedScore);
```
MitchelPaulin, line 418: I think we can get the behavior we want actually with 

```
this._editableScore = this._hoveredScore ?? this._score ?? this._selectedScore;
```

MitchelPaulin (approved):

devpow112:
CI issues caused by me cleaning up some upstream dependencies. Will have fix up in a bit.

PR #1442, opened 2021-12-14
Fixed issue where the score input would disapear if the sum of the criterion score was 0

MitchelPaulin:
https://rally1.rallydev.com/#/57732444928d/dashboard?detail=%2Fdefect%2F618611464791


src/components/editor/rubric-criterion-editor.js
```diff
@@ -439,7 +440,7 @@ class RubricCriterionEditor extends composeMixins(
 	}
 
 	get _isTextOnly() {
-		return !this.entity?.outOf;
+		return this.entity?.outOf === null || this.entity?.outOf === undefined;
```
devpow112, line 443: Change if you want. This call also be written as `this.entity?.outOf ?? false`. Not sure if that's any more readable but it is shorter. I believe this will only return `false` if the `this.entity?.outOf` is `null` or `undefined`.

devpow112 (approved):

src/components/editor/rubric-criterion-editor.js
MitchelPaulin: It looks like `0 ?? false` is `0`, which could get treated as false when used in an if statement, not sure if that would work

devpow112: Oh damn. NM I can't read. My suggestion won't work here 😛. Ignore me.

PR #1455, opened 2021-12-15
Fixed issue where score would not hide when rubric type would change

MitchelPaulin:
https://rally1.rallydev.com/#/57732444928ud/custom/373260458992?detail=%2Fdefect%2F618596385087

Issue ended up being a race condition between `_onEntityChanged` and when `hasOutOf` got updated. `_onEntityChanged` ran before `hasOutOf` was updated resulting in `_showPoints` not getting set to the right value and not hiding the score input. By moving the check into `render` this ensures we get the right values for `hasOutOf`.


src/components/editor/rubric-level-editor.js
```diff
@@ -126,7 +125,7 @@ class RubricLevelEditor extends composeMixins(
 				?enabled="${this._canEditName}"
 				pending-saves="${this._pendingNameSaves}"
 			></d2l-rubric-autosaving-input>
-			<div class="operations" ?nopoints="${!this._showPoints}">
+			<div class="operations" ?nopoints="${!this._canShowPoints(this.hasOutOf, this.entity)}">
```
devpow112, line 128: You should still be able to move the `this.hasOutOf` and `this.entity` into the `_canShowPoints` function itself. No need to pass them.

devpow112 (approved):

PR #1456, opened 2021-12-15
Fixed issue where score would not be saved if you clicked out of the input too quickly

MitchelPaulin:
https://rally1.rallydev.com/#/57732444928ud/custom/373260458992?detail=%2Fdefect%2F618611107029

Turns out the issue actually wasn't related to multi digit inputs but rather had to do with a race condition that would occur when you preformed too many actions too quickly. Removed the blur handler, made sure only one save event was sent and switched to an optimistic update strategy and I can no longer reproduce the problem


imcdonald-d2l (approved):
As long as there are no issues with removing the blur handler and the value gets saved as expected this looks fine to me 👍🏻

test/d2l-rubric-criterion-editor.test.js
```diff
@@ -201,9 +201,10 @@ describe('d2l-rubric-criterion-editor', () => {
 				});
 
 				requestAnimationFrame(() => {
-					subElement.dispatchEvent(new CustomEvent('focus', eventOptions));
 					subElement.dispatchEvent(new CustomEvent('input', eventOptions));
-					subElement.dispatchEvent(new CustomEvent('blur', eventOptions));
```
MitchelPaulin, line 206: The blur event no longer triggers a save, sending a save event manually.

imcdonald-d2l (approved):

devpow112 (approved):

PR #1462, opened 2021-12-16
Fix override button style issue - DE46272

imcdonald-d2l:
Using French to accentuate the issue. It has one of the longest translations for `Clear Override`.

### Before
override-btn-before

### After
override-btn-after


src/components/rubric-criteria-group.js
```diff
@@ -85,7 +85,6 @@ class RubricCriteriaGroup extends composeMixins(
 
 				d2l-table-wrapper[type="default"] td.out-of.assessable {
 					pointer-events: auto;
-					max-width: 75px;
```
imcdonald-d2l, line 88: Previously hard-coded width did not account for `Clear Override` button. Adjusted the styling of editable score component so we don't have to hard code this. Can't hardcode a width here since different translations have different width's of text.

devpow112 (dismissed):
Looks good to me :tada:

MitchelPaulin (approved):

imcdonald-d2l:
Just chatted with Fiona about design specifications and she is good with always stacking the elements in the pop-out/full-page view. Some changes I need to make though:

- make cancel button appear below input
- align `X` icon with input. Just need a margin-left of `10px`  on the input to get this aligned.

Will get these tomorrow morning as I wrap up other things.

devpow112:
> Align X icon with input. Just need a margin-left of 10px on the input to get this aligned.

Make sure to take RTL in account 😄

imcdonald-d2l:
> Make sure to take RTL in account 😄

Good catch, I would have definitely forgot that. 😅

devpow112:
I think eventually we should ad a couple of RTL vdiffs just to make sure we don't miss it

src/components/rubric-editable-score.js
```diff
@@ -87,7 +88,7 @@ class RubricEditableScore extends composeMixins(
 
 				:host([disable-inline]) {
 					display: flex;
-					flex-direction: column;
+					flex-direction: column-reverse;
```
devpow112, line 91: Will this also need an RTL set-up? I can't remember if flex direction is auto adjusted based on RTL

imcdonald-d2l: Yeah, I assume we wouldn't need a `RTL` set-up for the  flex stuff. I'll make sure I test it though 👍🏻

imcdonald-d2l: I think it would only really be a potential concern if it was `row-reverse` since column direction shouldn't matter in RTL

devpow112: Oh fair. Yup 👍

src/components/rubric-editable-score.js
```diff
@@ -83,6 +85,21 @@ class RubricEditableScore extends composeMixins(
...
+				}
+
+				:host([disable-inline][overridden-styling]) #editable-container {
+					margin-left: 12px;
+				}
+
+				:host([disable-inline][overridden-styling][dir="rtl"]) #editable-container {
+					margin-right: 12px;
+					margin-left: 0;
+				}
```
imcdonald-d2l, line 101: Adjusts the margin so the `x` icon aligns with the edge of input container.
override-btn-alignment
override-btn-alignment-rtl

MitchelPaulin (approved):

PR #1464, opened 2021-12-17
Remove unnecessary event and enable tab focus on feedback button - US134498

imcdonald-d2l:
Event isn't consumed anywhere so deleting it. Also removing `tabindex="-1"` preventing tab focus in pop-out view of rubric. Not sure why this was there in the first place 🤷🏻‍♂️


src/components/rubric-criteria-group.js
```diff
@@ -870,7 +870,6 @@ class RubricCriteriaGroup extends composeMixins(
 				addFeedbackButton = html`
 					<d2l-button-subtle
 						aria-hidden="true"
-						tabindex="-1"
```
imcdonald-d2l, line 873: @MitchelPaulin looks like you converted this component. Do you know why this might have been added in the first place?

MitchelPaulin: Looks like it was in the [original component](https://github.com/Brightspace/d2l-rubric/blob/3.x/src/d2l-rubric-criteria-group.js#L346). Probably just copy pasted in, I'm fine to remove it

MitchelPaulin (approved):

PR #1465, opened 2021-12-17
Us132979 rubric criteria groups conversion

MitchelPaulin:


src/components/rubric-criteria-groups.js
```diff
@@ -0,0 +1,159 @@
...
+	}
+	get assessmentEntity() {
+		return this._assessmentEntity;
+	}
+	set assessmentEntity(assessmentEntity) {
+		this._assessmentEntity = assessmentEntity;
+		const criterionLinks = assessmentEntity?.getSubEntitiesByClass('criterion-assessment-links') || [];
+		this._criterionHrefs = [];
+		criterionLinks.forEach(criterionLinkEntity => {
+			const criterionAssessmentLink = criterionLinkEntity.getLinkByRel('https://assessments.api.brightspace.com/rels/assessment-criterion');
```
MitchelPaulin, line 62: These classes and links are not available in hypermedia-constants, will probably add them and then update these accordingly

src/components/rubric-criteria-group.js
```diff
@@ -844,10 +844,10 @@ class RubricCriteriaGroup extends composeMixins(
 	}
 	_queueRadio(radio) {
 		const href = radio.getAttribute('data-href');
-		const entity = this.cellAssessmentMap[href];
-		const criterionHref = entity?.getLinkByRel('https://rubrics.api.brightspace.com/rels/criterion')?.href;
+		const entity = this.cellAssessmentMap.get(href);
+		const criterionHref = entity?.getLinkByRel(Rels.Rubrics.criterion)?.href;
 		if (criterionHref) {
-			this._queueChange((() => this.cellAssessmentMap[href]).bind(this));
+			this._queueChange((() => this.cellAssessmentMap.get(href)).bind(this));
```
imcdonald-d2l, line 850: nit: do we need to bind `this` if it's using an arrow function here?

src/components/rubric-criteria-groups.js
```diff
@@ -0,0 +1,159 @@
...
+	updated(changedProperties) {
+		super.updated(changedProperties);
+		if (changedProperties.has('criterionAssessmentMap') && this.criterionAssessmentMap.entries()) {
+			for (const entity of this.criterionAssessmentMap.values()) {
+				const rubricCriterionHref = entity.criterionHref;
+				if (rubricCriterionHref) {
+					this.criterionAssessmentMap.set(rubricCriterionHref, entity);
+				}
+
+				entity.getSubEntitiesByClass('assessment-criterion-cell').forEach(cell => {
```
imcdonald-d2l, line 85: Do we need a `?` after `getSubEntitiesByClass(...)` in case it doesn't return anything? Or does `forEach` able to handle that by default?

src/components/rubric-criteria-groups.js
```diff
@@ -0,0 +1,159 @@
...
+					}
+				});
+			}
+			this.dispatchEvent(new CustomEvent('cell-assessment-map-update', { detail: { value: this.criterionAssessmentMap }, bubbles: true, composed: true }));
+		}
+	}
+
+	_isHolistic(rubricType) {
+		return rubricType === 'holistic';
+	}
```
imcdonald-d2l, line 98: I'm not seeing this function referenced anywhere, is it still needed?

src/components/rubric-criteria-groups.js
```diff
@@ -0,0 +1,159 @@
...
+	_onEntityChanged(entity) {
+		this._groups = entity.criteriaGroups;
+		this._showContent = true;
```
imcdonald-d2l, line 101: Is this property necessary? Or can we get away with just referencing `this.entity` and see if it's set?

src/components/rubric-criteria-groups.js
```diff
@@ -0,0 +1,159 @@
...
+	}
+	_renderCriterionGroup(item) {
+		return html`
+				<d2l-rubric-criteria-group
+					href="${item.getLinkByRel('self').href}"
```
imcdonald-d2l, line 106: We should either bind to a property here with `.href` or return `''` if `item.getLinkByRel('self').href` returns undefined. I've run into this issue a couple times and it messes with entity fetching and vdiff when it returns undefined.

src/components/rubric-criteria-groups.js
```diff
@@ -0,0 +1,159 @@
...
+					.criterionResultMap="${this.criterionAssessmentMap}"
+					.cellAssessmentMap="${this.cellAssessmentMap}"
+					.context="${this.context}"
+				></d2l-rubric-criteria-group>
+			`;
+	}
+	_renderCriterionGroupMobile(item) {
+		return html`
+				<d2l-rubric-criteria-group-mobile
+					href="${item.getLinkByRel('self').href}"
```
imcdonald-d2l, line 120: Same goes with this href attribute. Should be bound to a property or set to `''` by default

src/components/rubric-levels-mobile.js
```diff
@@ -622,7 +622,7 @@ class RubricLevelsMobile extends composeMixins(
 		}
 
 		/**
-		 * If a mousemove event has occured after mousedown, we set elementHasMoved to true
+		 * If a mousemove event has occurred after mousedown, we set elementHasMoved to true
```
imcdonald-d2l, line 625: Thanks for fixing all my spelling mistakes 😅

src/d2l-rubric.js
```diff
@@ -535,10 +539,16 @@ const component = {
 
 	attached: function() {
 		window.addEventListener('beforeunload', (e) => this._showPopoutDialogWarning(e));
+		window.addEventListener('cell-assessment-map-update', (e) => this.updateAssessmentMap(e));
```
imcdonald-d2l, line 542: nit: you could do this in one line:
```suggestion
		window.addEventListener('cell-assessment-map-update', (e) => this._criterionAssessmentMap = e.detail.value);
```

src/components/rubric-criteria-groups.js
MitchelPaulin: In this case I think its better to have an error get spit out to the console if this is not defined, would make tracking down errors easier rather than silently not rendering it IMO

src/components/rubric-criteria-groups.js
MitchelPaulin: Correct me if I'm wrong but because entity is not in properties updates to it are not observable so we need an extra variables for it

src/components/rubric-criteria-groups.js
MitchelPaulin: Referenced on line 126, but there is a typo there, good catch

imcdonald-d2l: If it's only used once we probably don't need a separate function for this simple check too.

src/components/rubric-criteria-group.js
MitchelPaulin: Seems to work without it

src/components/rubric-criteria-groups.js
imcdonald-d2l: Hmmm you may be right. It seems to be a private property on `EntityMixinLit` which is then accessed from `EntityMixin`. So theoretically I think it _should_ trigger updates. Also fine to have a separate property to be extra safe 👍🏻

src/components/rubric-criteria-group.js
imcdonald-d2l: Cool 👍🏻

src/components/rubric-criteria-groups.js
MitchelPaulin: Yeah, I removed it and just did the check right in the render function

src/components/rubric-criteria-groups.js
imcdonald-d2l: Makes sense to me 👍🏻

src/components/rubric-criteria-groups.js
MitchelPaulin: Having a `_showContent` variable is also an established pattern in other components so I think we should keep it, though I wouldn't be opposed to removing all of them as part of a different story.

imcdonald-d2l: Yeah, we can keep it for now and clean up the other references later (assuming checking `this.entity` works as we hope)

imcdonald-d2l (approved):

src/components/rubric-levels-mobile.js
MitchelPaulin: Np, if you don't have it already https://github.com/streetsidesoftware/vscode-spell-checker is a great extension for vscode

imcdonald-d2l: Good to know, definitely going to add that now!

PR #1467, opened 2021-12-17
De46395 text overlap standards button

jesseag:


src/entities/entity-base.js
```diff
@@ -131,7 +131,7 @@ export class EntityBase extends Entity {
 	}
 
 	getLinkHrefByRel(rel) {
-		return this.getLinkByRel(rel)?.href;
+		return this.getLinkByRel(rel)?.href || '';
```
imcdonald-d2l, line 134: This was added so that it always returns a string by default, that way we don't get issues when trying to fetch hrefs that are set to `undefined`. This could potentially break other areas if they are checking for `undefined` or if there are areas where we don't want to return a string.

src/entities/rubric-criterion-cell-entity.js
```diff
@@ -63,4 +63,7 @@ export class RubricCriterionCellEntity extends EntityBase {
 		return this.properties?.points;
 	}
 
+	getSelfLink() {
+		return this._entity?.href || this._entity?.getLinkByRel('self')?.href || '';
```
imcdonald-d2l, line 67: This was necessary because the entity seems to get into a weird state where the href is added directly on the entity instead of being set to a `rel`.

jesseag:
Thanks to @imcdonald-d2l for his help here with the hypermedia part.

There are 2 things going on in this PR.

1) Bug fix for the defect (DE46395). This is most of the CSS changes along with adding in a new div <div class="input-action"> above the d2l-input-textarea. 
2) Separate bug that was discovered by Ian where an href was getting set to undefined (then back to defined) which caused the span of the text box in <div class=criterion-feedback> to get collapsed/expanded multiple times whenever text was entered into any of the text boxes.

src/entities/entity-base.js
MitchelPaulin: How often do we use this method? Most calls I've seen use `getLinkByRel` directly

imcdonald-d2l: So this function is actually used a fair bit since the above `getHref()` function references it as well. So there are of a total 30ish references to this function. Most of the references seem to be equality checks but there are places where it references the returned value and an empty string would work just fine in those cases.

MitchelPaulin: I'm worried about situations where we would have 

```
if (entity.getLinkHrefByRel('re')) {
    //do something
}
```

but if we have nothing of that kind then this will be fine

src/entities/rubric-criterion-cell-entity.js
MitchelPaulin: We do this in a couple other entity classes too, I wonder what the root cause of this is

src/entities/entity-base.js
imcdonald-d2l: There are no cases in the code where we have a condition like that at the moment and those cases shouldn't be added in the future since we should use `hasLinkByRel()` in those cases.

MitchelPaulin (approved):

PR #1468, opened 2021-12-20
Rubric header width issues - DE45633

imcdonald-d2l:
- Fixes resize observer to catch all re-size events
- Set min-width on `.col-last` so that we don't get re-size jitter when switching from no-score to points rubric (as discussed in stand-up)

Note: new-vdiffs are expected for no-score rubric types as `.col-last` width should be increasing with this change.

https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fdefect%2F611729484089


src/components/editor/rubric-levels-editor.js
```diff
@@ -147,20 +152,11 @@ class RubricLevelsEditor extends composeMixins(
...
-	disconnectedCallback() {
-		window.removeEventListener('resize', (e) => this._checkSize(e));
-		super.disconnectedCallback();
-	}
-
 	async firstUpdated() {
 		// Give the browser a chance to paint
 		await new Promise((r) => setTimeout(r, 0));
 		this._checkSize();
+		new ResizeObserver(() => this._checkSize()).observe(this.shadowRoot.querySelector('#levels-section'));
```
jesseag, line 159: Might be a silly question, but the ResizeObserver is run when the page is initially loaded and then is always active and monitoring this div for changes throughout the lifecycle of the page. Is that correct? 
https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver/observe

src/styles/rubric-editor-cell-styles.js
```diff
@@ -56,7 +56,11 @@ export const rubricEditorCellStyles = css`
 	.col-last {
 		flex: 0 0 auto;
 		min-width: auto;
```
MitchelPaulin, line 58:
```suggestion

```

src/components/editor/rubric-levels-editor.js
imcdonald-d2l: Yes, that is correct as far as I understand it. It will always monitor this div for changes as long as it exists in the DOM.

MitchelPaulin (approved):

jesseag (approved):

jesseag:
@imcdonald-d2l 
Did you want the 3 photo (.png) files checked into git? Git doesn't usually like to have binaries stored in it? Unless they are backed up using Git LFS in a storage (like Artifactory).

MitchelPaulin:
> @imcdonald-d2l Did you want the 3 photo (.png) files checked into git? Git doesn't usually like to have binaries stored in it? Unless they are backed up using Git LFS in a storage (like Artifactory).
> 
> I don't remember those being there as part of the initial PR.

Yeah that's intentional, we checks them right into the repo https://github.com/Brightspace/d2l-rubric/tree/master/test/perceptual/screenshots/ci/golden. Images were merged in from here https://github.com/Brightspace/d2l-rubric/pull/1469

jesseag (approved):

jesseag:
Ahh I see it's from the Vdiff stuff so we have something to reference in case they are changes. Okay 👍

PR #1481, opened 2021-12-23
Fix issue where the criterion name cells text disappears while typing.

jesseag:
Defect: https://rally1.rallydev.com/#/57732444928d/dashboard?detail=%2Fdefect%2F620795438481&fdp=true

Problem: 
- Each keystroke triggers a _handleNameInput event which calls into the async _handleNameSave function. 
- Each call into _handleNameSave gets run after a timeout of 500ms via the DebouncerMixin. As a result, you can have a bunch of save calls waiting to happen.
https://github.com/Brightspace/d2l-rubric/blob/master/src/mixins/debouncer-mixin.js
- If the user types with the correct "timing" and the name is not changing (this.__nameChanging = false) and there are no pending saves (this._nameSaves = 0) (on Line 606) then the this._name gets saved and reflected immediately (Line 805) at the same time the user enters the keystroke. This then gives the impression that as the user enters the keystroke it disappears.

Solution: Re-implemented older/similar solution from polymer and rubric-autosaving-input.js where we always display the user input (event.target.value) if the text box is in focus, is invalid or the number of pending saves is greater than 0. Otherwise, we display the saved value.

Older Polymer code:
https://github.com/Brightspace/d2l-rubric/blob/3.x/src/editor/d2l-rubric-criterion-editor.js

Similar solution in rubric-autosaving-input.js
https://github.com/Brightspace/d2l-rubric/blob/master/src/components/editor/rubric-autosaving-input.js


src/components/editor/rubric-criterion-editor.js
```diff
@@ -537,6 +543,21 @@ class RubricCriterionEditor extends composeMixins(
 		announce(this.localize('criterionDeleted', 'name', this.entity?.name));
 	}
 
+	_handleNameBlur(event) {
+		/*
+		if (this._nameChanging || !this._nameSaves && this._nameError) {
+			this._handleNameSave(event.target.value);
```
jesseag, line 549: this._handleNameSave triggers a second call to 'd2l-siren-entity-save-error' which breaks the unit tests. I will either need to update the unit test to check for 2 events of this type or remove this call to save.

test/d2l-rubric-criterion-editor.test.js
```diff
@@ -90,6 +90,8 @@ describe('d2l-rubric-criterion-editor', () => {
 					flush(() => {
 						expect(subElement.ariaInvalid).to.equal('true');
 						done();
+						// Ignore subsequent invocations due to blur event which triggers another 'd2l-siren-entity-save-error' event
```
devpow112, line 93: Does doing something like this work
```js
const eventHandler = () => {
  flush(() => {
    expect(subElement.ariaInvalid).to.equal('true');
    done();
  })
};

element.addEventListener('d2l-siren-entity-save-error', eventHandler , {once: true});
```
I thought I remember there being a away to have an event handler that only triggers once (I'm probably wrong)

jesseag: It should remove the event listener once it is invoked according to the documentation.
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener


once
A boolean value indicating that the listener should be invoked at most once after being added. If true, the listener would be automatically removed when invoked.

I've made the changes, let's see if it works.

devpow112 (approved):

PR #1483, opened 2022-01-04
DE46618 - Translation Updates

devpow112:
This will need to be hotfixed into `20.22.1` as well.


devpow112:
Looks like there are some manual fixes to be done 😞.

jesseag (approved):
Looks good.

PR #1486, opened 2022-01-05
Inline override adjustments - US134747

imcdonald-d2l:
Adjust inline override styles to align with specs outlined on this [rally ticket](https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F620800416869). 

Additionally, the hover/focus state has been adjusted so that the overridden score value/button isn't visible when hovering/focusing other criterion levels. This was discussed and requested by Fiona. Adding @MarkMoerkerken77 @djwester to this PR just for the sake of getting their awareness on this change. See before/after gif below.

### Before
old-inline-behavior
### After
new-inline-behavior


devpow112:
For the after image we seem to have lost the border on the input box, is that purposeful

Before:
image

After:
image

imcdonald-d2l:
> @imcdonald-d2l for the after image we seem to have lost the border on the input box when it's in focus, is that purposeful?

Good catch, no that wasn't purposeful. I'll look into fixing that 👍🏻

imcdonald-d2l:
Border input fixed: 
fix-input-border

Note that there seems to be an odd transition between borders when clicking on the input field directly. This exists in the polymer version as well. Something we might want to address at a later time.

devpow112 (approved):

PR #1487, opened 2022-01-05
Update html-block API usage

grant-cleary:
See this PR for context: https://github.com/BrightspaceUI/core/pull/1960

Usage should be identical, but sans template. 😄


src/components/rubric-criterion-mobile.js
```diff
@@ -334,8 +334,8 @@ class RubricCriterionMobile extends composeMixins(
 		return html`
 			<div class="criterion-description">
 				<p class="criterion-description-html">
-					<d2l-html-block>
-						<template id="criterion-description-html">${unsafeHTML(this._criterionCells?.[this._currentLevel]?.descriptionHtml)}</template>
+					<d2l-html-block id="criterion-description-html">
+						${unsafeHTML(this._criterionCells?.[this._currentLevel]?.descriptionHtml)}
```
grant-cleary, line 338: The template was formerly referenced here: https://search.d2l.dev/xref/Brightspace/d2l-rubric/src/d2l-rubric-criterion-mobile.js?r=6a2ff5f2#370

I'm not sure it is anymore, or at least I can't find it referenced, but I'm not sure if that's work in progress so I figured I would keep the `id` in place. This should work identically, as the html-block slots its immediate liteDOM children.

devpow112: From what I can see `criterion-description-html` is only used as a CSS class and not as an id. I think leaving it makes sense. We can always clean it up later.

devpow112 (approved):

src/components/rubric-criterion-mobile.js
grant-cleary: Cool, that matches what I could glean from code search. It looks like the ID was used once, but not anymore. Good to know I'm not crazy!

PR #1488, opened 2022-01-05
Remove more `d2l-html-block` nested `template`

devpow112:
Needed for Lit 2.0 upgrade eventually.


imcdonald-d2l (approved):

devpow112:
Having issues with sauce labs 😞. Random failures and disconnects.

PR #1495, opened 2022-01-10
Levels mobile enter issue (+ other cleanup items) - DE46649

imcdonald-d2l:
Adds functionality so that `Enter` key will now select/deselect criterion in mobile view. This functionality didn't seem to previously exist.

This PR also fixes some incorrect logic I added to `watchEntityMap` and some consequences of fixing it. Basically I wasn't updating entities if it existed in the map already. This shouldn't be done as entities should be re-fetched when they are updated. For example, it wasn't updating `_criterionAssessmentMap` as expected and actions for selecting criterion were not getting the proper values messing with a lot of things. 

Defect: https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fdefect%2F621668091693


src/components/rubric-criteria-groups.js
```diff
@@ -19,12 +19,11 @@ class RubricCriteriaGroups extends composeMixins(
 	static get properties() {
 		return {
 			compact: { type: Boolean },
-			assessmentHref: { type: String, attribute: 'assessment-href' },
 			rubricType: { type: String, attribute: 'rubric-type' },
 			readOnly: { type: Boolean, attribute: 'read-only' },
 			assessmentEntity: { type: Object },
-			criterionAssessmentMap: { type: Object },
```
imcdonald-d2l, line 26: `assessmentHref` wasn't used anywhere and `criterionAssessmentMap` should not be a property when this component defines it and emits an event to update the parent. Was causing some fun infinite loop problems.

src/components/rubric-criteria-groups.js
```diff
@@ -44,57 +43,59 @@ class RubricCriteriaGroups extends composeMixins(
...
-		const criterionLinks = assessmentEntity?.criterionAssessmentLinks || [];
-		this._criterionHrefs = [];
-		criterionLinks.forEach(criterionLinkEntity => {
-			const criterionAssessmentLink = criterionLinkEntity.getLinkByRel(Rels.Assessments.assessmentCriterion);
-			if (criterionAssessmentLink) {
-				this._criterionHrefs.push(criterionAssessmentLink.href);
-			}
-		});
-		this.requestUpdate('criterionAssessmentMap', []);
-		this.requestUpdate('_cellAssessmentMap', []);
```
imcdonald-d2l, line 68: `get assessmentEntity` wasn't needed and moved `set` logic to `updated()` instead.

src/mixins/entity-watcher-mixin.js
```diff
@@ -61,10 +61,8 @@ const EntityWatcherMixinInternal = (superclass) => class extends superclass {
 			return new Promise(resolve => resolve(null));
 		}
 		const promiseArray = [];
-		const valueMap = this[entityProp] ? new Map(this[entityProp]) : new Map();
+		const valueMap = new Map();
 		for (const href of hrefs) {
-			// don't fetch href if it already exists in the map
-			if (valueMap.has(href)) continue;
```
imcdonald-d2l, line 67: I shouldn't have added this here in the first place. We need to re-fetch entities in-case there are updates to them.

src/mixins/entity-watcher-mixin.js
```diff
@@ -86,7 +83,10 @@ const EntityWatcherMixinInternal = (superclass) => class extends superclass {
 				}
 			})));
 		}
-		return Promise.all(promiseArray);
+
+		return Promise.all(promiseArray).then(() => {
+			this[entityProp] = valueMap;
```
imcdonald-d2l, line 88: Updating the property when all promises have resolved so we don't update the property too many times and trigger multiple re-renders.

src/components/rubric-levels-mobile.js
```diff
@@ -827,25 +856,29 @@ class RubricLevelsMobile extends composeMixins(
...
 
-	_queueAction(entityHref, updateSelected = false) {
+	_queueAction(entityHref) {
 		if (this.readOnly) return;
-		const entityGetter = () => this.assessmentMap.get(entityHref);
-		this._pendingChanges++;
-		if (this.onSelectOptimistic) {
-			this.onSelectOptimistic(this.assessmentMap.get(entityHref));
+		if (this._keyboardContext.isUpdatingSelectState) {
+			this.debounce('select-criterion-keyboard', () => {
```
imcdonald-d2l, line 862: Adding a debouncer on keyboard actions so that when a user is iterating through the options they don't send a request every single time.

src/components/rubric-levels-mobile.js
```diff
@@ -916,17 +950,17 @@ class RubricLevelsMobile extends composeMixins(
 		}, 1000);
 	}
 
-	_selectAssessedLevel(cells, assessmentEntity) {
+	_selectAssessedLevel() {
 		if (
-			this._pendingChanges !== 0 ||
-			!cells ||
-			!cells.length ||
+			(this.mobileAccordionOpen && !this.overallScoreSelector) ||
```
imcdonald-d2l, line 955: Adding `this.mobileAccordionOpen` is unfortunately necessary since with a debouncer `this.assessmentMap` lags behind with every criterion select action. If you navigate or click rapidly it triggers multiple select requests that are delayed causing the `this.assessmentMap` values to lag behind. When `this.assessmentMap` changes it calls `this_selectAssessedLevel()` in `updated()` and causes visual re-select when it shouldn't. See gif:
select-lag

imcdonald-d2l: @devpow112 @MitchelPaulin do you have any better solutions in mind since this does not seem ideal? Especially when we need to be updating overall score selector on assessment map changes. Should we just not update the selector on assessment map changes if there are requests still in flight?

README.md
```diff
@@ -152,6 +152,8 @@ will be automatically opened for you against your branch.
 
 #### Running Locally
 
+**Note**: if the following commands are run with elevated permissions locally they will not work.
```
devpow112, line 155: Really? Damn 😞. What error does it give?

imcdonald-d2l: It just gives a timeout error every time. We found this a couple weeks back on my machine and I believe you were able to reproduce as well. It is fixed by simply opening a regular terminal without elevated permissions.

test/perceptual/utilities.js
```diff
@@ -54,6 +54,10 @@ export const selectLevelAndFocus = async(page, selector, focus) => {
 			if (selector.includes('d2l-rubric-criterion-mobile')) {
 				element._currentLevel = 2;
 				element._editableScore = 2;
+			} else {
```
devpow112, line 57: Hopefully, if we can ever get assessment data to work/generate, we won't need this.

README.md
devpow112: Ah yes, I remember that. Thanks for noting it in the README

test/perceptual/utilities.js
imcdonald-d2l: Yeah, that would certainly be ideal.

README.md
devpow112: Yup, I remember using that to solve issues. Can be super useful. Good call 👍

README.md
```diff
@@ -163,6 +165,10 @@ npm run test:vdiff
 npm run test:vdiff -- -g some-pattern
 ```
 
+#### Debugging visual diff tests locally
```
MitchelPaulin, line 168:
```suggestion
#### Debugging Visual Diff Tests Locally
```

src/components/rubric-levels-mobile.js
```diff
@@ -452,33 +456,41 @@ class RubricLevelsMobile extends composeMixins(
...
+	_executeSelectAction(entityHref) {
+		const entityGetter = () => this.assessmentMap.get(entityHref);
+		if (this.onSelectOptimistic) {
+			this.onSelectOptimistic(this.assessmentMap.get(entityHref));
+		}
+		this.onCriterionSelect(entityGetter);
+	}
+
+	_getCriterionCellTabIndex(index) {
+		return index === (this._focused === -1 ? 0 : this._focused) ? 0 : -1;
```
MitchelPaulin, line 468:
```suggestion
		return index === Math.max(this._focused, 0);
```

MitchelPaulin, line 468:
```suggestion

		const focused = this._focused === - 1 ? 0 : this._focused;
		return index === focused ? 0 : -1;
```

src/components/rubric-levels-mobile.js
```diff
@@ -308,13 +311,13 @@ class RubricLevelsMobile extends composeMixins(
 	constructor() {
 		super();
 		this._scroll = 0;
-		this._pendingChanges = 0;
 		this._levelEntities = [];
 		this._assessmentEntity = {};
 		this._selected = -1;
 		this._focused = -1;
 		this._hovered = -1;
```
MitchelPaulin, line 318: As a bit of an aside as something I think we should address in a future PR, using `-1` to represent not selected seems to be leading to a lot of nasty 0,-1 flipping and IMO is making things pretty hard to read. Ideally we would use something like `Option<T>` but since were stuck with js `undefined` may be better? It will spit out NaN if you try to do arithmetic with it unlike `null` which will happily convert itself to 0

imcdonald-d2l: Agreed this `-1` stuff can get pretty messy and should be addressed in a separate PR. `undefined` could be a nice alternative and would throw more expected errors when we use it incorrectly.

MitchelPaulin:
@imcdonald-d2l I see you marked the comments as resolved, do you have the changes locally and just haven't pushed them?

MitchelPaulin (approved):

devpow112 (approved):

PR #1509, opened 2022-01-12
Overall score mobile issues - DE46731

imcdonald-d2l:
- Fixes overall levels mobile component so that overall level is updated when assessment entity changes
- Adds the debouncer for click events (not just keyboard interactions). Was getting unexpected behavior when spam clicking within the levels mobile component that debouncing seems to solve. 
- Other small code clean-up items. 

https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fdefect%2F622196855949


src/components/rubric-levels-mobile.js
```diff
@@ -395,7 +396,7 @@ class RubricLevelsMobile extends composeMixins(
 		}
 	}
 
-	deselect() {
+	deselect(callback = noop) {
```
imcdonald-d2l, line 399: Adding callback here so that we can update overall assessment level entity when clearing override.

src/components/rubric-levels-mobile.js
```diff
@@ -626,31 +626,24 @@ class RubricLevelsMobile extends composeMixins(
 
 		// only select/deselect level if drag has moved to a separate criterion
 		// or mouse never dragged in the first place
-		if (!dragBackToSameCriterion || !isDragEvent) {
+		if (!isDeselectingCriterion && (!dragBackToSameCriterion || !isDragEvent)) {
 			this._selectLevelAt(this._focused);
-		}
-
-		if (!this.readOnly) {
-			this.focus();
```
imcdonald-d2l, line 634: Calling `this.focus()` isn't necessary so removing it.

src/components/rubric-levels-mobile.js
```diff
@@ -856,23 +849,18 @@ class RubricLevelsMobile extends composeMixins(
...
 		if (this.readOnly) return;
-		if (this._keyboardContext.isUpdatingSelectState) {
-			this.debounce('select-criterion-keyboard', () => {
-				if (this._keyboardContext) {
-					if (this._keyboardContext?.initialCell === entityHref) {
-						// don't execute action when keyboarding back to initial cell (for left and right keys only)
-						return;
-					}
-					this._keyboardContext.isUpdatingSelectState = false;
+		this.debounce('select-criterion-keyboard', () => {
```
imcdonald-d2l, line 854: Adding debouncer for both keyboard and click actions now. Was noticing issues when selecting/deselecting a given criterion quickly (via click events) that this solves.

src/components/rubric-overall-score-mobile.js
```diff
@@ -76,7 +78,6 @@ class RubricOverallScoreMobile extends composeMixins(
 		this._levels = [];
 		this._assessmentMap = new Map();
 		this._competencies = [];
-		this.watchEntity('assessmentHref', '_assessmentEntity');
```
imcdonald-d2l, line 79: Main issue for overall level not updating was that `_assessmentEntity` (re-named to `_overallLevelAssessmentEntity` for clarity) wasn't updating when it should. `assessmentHref` (re-named to `_overallLevelAssessmentHref` for clarity) wasn't changing on `assessmentEntity` updates in `rubric.js` so it never triggered a re-fetch. I now force-update whenever `assessmentEntity` updates so that we get the latest overall score state.

src/components/rubric.js
```diff
@@ -685,7 +685,7 @@ class Rubric extends composeMixins (
 					?read-only='${this.readOnly}'
 					.token='${this.token}'
 					href='${this.entity.overallLevelsRel?.href || ''}'
-					assessment-href='${this._getOverallLevelAssessmentHref(this.assessmentEntity, this._cachePrimed)}'
```
imcdonald-d2l, line 688: @MitchelPaulin @devpow112 are either you aware of the purpose of `this._cachePrimed` and `this._cachePrimeEntity`? They are defined in `rubric.js` but not used in any meaningful context from what I can tell.

src/components/rubric-overall-score-mobile.js
```diff
@@ -116,17 +117,25 @@ class RubricOverallScoreMobile extends composeMixins(
...
+		if (changedProperties.has('_overallLevelAssessmentEntity') && this._overallLevelAssessmentEntity) {
 			this._onAssessmentEntityChanged();
 		}
+
+		if (changedProperties.has('assessmentEntity') && this.assessmentEntity) {
+			this._overallLevelAssessmentHref = this.assessmentEntity.getLinkByRel(
+				'https://assessments.api.brightspace.com/rels/assessment-overall-level'
+			)?.href || '';
+
+			this._updateOverallLevelAssessment();
```
imcdonald-d2l, line 129: Overall score now updates as expected: 
overall-score-updating

src/components/rubric.js
devpow112: I think this is left over from the polymer version. The polymer version did some sort of cache priming stuff. @MitchelPaulin would probably have more context

MitchelPaulin: It visits the cache primer link if its there, the _cachePrimed variable ensures that the page reloads once the assessment entity does. 

Not sure how important priming the cache is but it was there in the Polymer version so I re-implemented it here.

imcdonald-d2l: > `_cachePrimed` variable ensures that the page reloads once the assessment entity

Do you mind elaborating? The `assessment-href` attribute in this case did not update with assessment entity changes... the string value doesn't change but that is expected. Was the `_cachePrimed` variable supposed to trigger a manual update in this case?

MitchelPaulin: If cahcePrimed is false the function returns `''` once cachePrimed gets set the function then returns the actual href, so this acts as a guard to make sure the cahcePrimer call returned before we pass in the href, which would cause a load in the child component

src/components/rubric-overall-score-mobile.js
devpow112: I wonder if there will be a concern with the delay between changing criteria and the overall score appearing/changing? It looks like it updates after the same is done (which makes sense from a technical standpoint) but we've had complaints about this in the past and had to fake it to have no perceived delay.

src/components/rubric-levels-mobile.js
```diff
@@ -21,6 +21,7 @@ import { styleMap } from 'lit-html/directives/style-map.js';
 
 // radius for determining if drag has occurred
 const DRAG_DISTANCE_PX = 5;
+const noop = () => {};
```
devpow112, line 24: Can this not be just inlined
```js
deselect(callback = () => {}) {
```
Or do you think that's not clear enough so you wanted to name it. The other option is in the `deselect` function to check if callback is undefined and then not do anything with `_queueAction`.

devpow112: Ah nm I see `_queueAction` does a bunch of other stuff.

src/components/rubric-overall-score-mobile.js
devpow112: Maybe as Mike/Fiona. I feel like SNHU will complain right away. If we want to tackle it as another PR as well I'm fine with this going in as is and we can just file another issue for the lag

src/components/rubric-overall-score-mobile.js
```diff
@@ -98,7 +99,7 @@ class RubricOverallScoreMobile extends composeMixins(
 			<d2l-rubric-levels-mobile
 				.href="${this.href}"
 				.token="${this.token}"
-				assessment-href="${this.assessmentHref}"
+				assessment-href="${this._overallLevelAssessmentHref || ''}"
```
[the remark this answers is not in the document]
imcdonald-d2l: Seems to work just fine with `ifDefined` 🎉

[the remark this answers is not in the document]
devpow112: Awesome. We probably have other places we can do this as well but that's not for this PR.

src/components/rubric-levels-mobile.js
mpharoah-d2l: Debouncing shouldn't be necessary with the optimistic updater and could interfere with it. Looks like it already existed for click events though....

src/components/rubric-overall-score-mobile.js
imcdonald-d2l: Yeah, I can follow up with them on this issue. Regardless, it should be in another PR so we don't overcomplicate things in this one.

devpow112 (approved):

src/components/rubric-levels-mobile.js
imcdonald-d2l: This _shouldn't_ interfere with the optimistic updater since in the levels mobile component we don't update the UI when any requests are still in flight (see comment [here](https://github.com/Brightspace/d2l-rubric/pull/1509#discussion_r784126504)). So by the time all requests are complete, the UI _should_ align with what the back-end state is.

PR #1510, opened 2022-01-13
Rubric feedback print styles - US133167

imcdonald-d2l:
- removes feedback arrow in print view
- remove `position: relative;` as it was messing with some print styles. See example:

| Before | After |
| --- | --- |
| position-relative | position-relative-fixed |


jesseag:
> * removes feedback arrow in print view
> * remove `position: relative;` as it was messing with some print styles. See example:
> 
> Before	After
> position-relative	position-relative-fixed

Is it okay that the checkmark is overtop of criterion name?

imcdonald-d2l:
> Is it okay that the checkmark is overtop of criterion name ("amet ipsum") in the after picture?

Yeah, that is not ideal. I'll see if I can add some padding to the text so we can avoid that.

imcdonald-d2l:
@jesseag issue is now fixed!
| Before | After |
| --- | --- |
| check-before | check-after |

MitchelPaulin (approved):

jesseag (approved):
Looks alright to me. It bugs me that we have to use hardcoded values sometimes but in this case I think it makes sense.

devpow112 (approved):

PR #1515, opened 2022-01-17
Don't use root path to import `d2l-rubric`

devpow112:
This may be causing an issue with polymer test in activities repo.


src/components/rubric.js
```diff
@@ -16,7 +10,6 @@ import '@brightspace-ui/core/components/html-block/html-block.js';
 import '@brightspace-ui/core/components/button/button-subtle.js';
 import '@brightspace-ui/core/components/offscreen/offscreen.js';
 import '@brightspace-ui/core/components/icons/icon.js';
-import 'd2l-hypermedia-constants/d2l-hypermedia-constants.js';
```
devpow112, line 19: This shouldn't be needed either. Since we import it explicitly below.

imcdonald-d2l (approved):

PR #1519, opened 2022-01-18
Fix mobile feedback issues DE46693 DE46305

imcdonald-d2l:
Fixes:
- feedback input box wouldn't disappear when feedback was cleared. This was fixed by moving feedback rendering logic to the `rubric-feedback` component so we didn't have to rely on the parent component making unnecessary calls to determine if things should be rendered when the feedback component has all that information already.
- overridden feedback wouldn't stick in mobile view when selecting criterion cells after updating feedback. This is because hypermedia select actions were not getting updated in `levels-mobile` component with overridden feedback. Had to force-update the entities to get the proper action values.

Rally tickets:
- [DE46305](https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fdefect%2F618765425501)
- [DE46693](https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fdefect%2F621932536725)


src/mixins/entity-watcher-mixin.js
```diff
@@ -89,37 +151,6 @@ const EntityWatcherMixinInternal = (superclass) => class extends superclass {
 		});
 	}
 
-	__fetchWatchedEntity(href, entityProp, entityClass) {
```
imcdonald-d2l, line 92: I just moved and renamed this with public notation since it is used outside this mixin.

src/mixins/entity-watcher-mixin.js
```diff
@@ -36,13 +36,75 @@ const EntityWatcherMixinInternal = (superclass) => class extends superclass {
...
+					bubbles: true
+				}));
+				setTimeout(() => { throw error; });
+			}
+
+			resolve(null);
+		}));
+	}
+
+	updateEntityMapEntry(mapPropertyName, hrefToUpdate, entityClass) {
```
imcdonald-d2l, line 70: Added this function so that we can update individual entities in a map without having to update every entity in the map.

src/components/rubric-feedback.js
```diff
@@ -273,139 +268,14 @@ class RubricFeedback extends composeMixins(
 	}
 
 	render() {
-		let newEditor = null;
```
imcdonald-d2l, line 276: Just moved all this logic to a separate render function.

src/components/rubric-criteria-group-mobile.js
```diff
@@ -226,75 +189,22 @@ class RubricCriteriaGroupMobile extends composeMixins(
...
 				?read-only="${this.readOnly}"
+				?has-update-action="${this._lookupMap(criterion)?.hasActionByName('update-critierion-assessment')}"
 			></d2l-rubric-criterion-mobile>
-			${this._renderFeedbackBtn(criterion, criterionNum)}
-			${this._renderFeedback(criterion, criterionNum)}
 			<hr class="line">
 		`;
 	}
-
-	_renderFeedback(criterion, criterionNum) {
```
imcdonald-d2l, line 245: Moved all feedback rendering and related logic to `criterion-mobile` so we don't compute unnecessary feedback render logic in this component and so it is a more logical space.

src/components/rubric-criteria-group.js
```diff
@@ -946,6 +946,7 @@ class RubricCriteriaGroup extends composeMixins(
 						criterion-assessment-href="${this._getCriterionResultHref(criterion, this.criterionResultMap)}"
 						.token="${this.token}"
 						.context="${this.context}"
+						?adding-feedback="${ criterionNum === this._addingFeedback }"
```
devpow112, line 949: nit: extra spaces in `${}`

src/components/rubric-criteria-groups.js
```diff
@@ -52,6 +52,16 @@ class RubricCriteriaGroups extends composeMixins(
 		this.watchEntityMap('_criterionHrefs', '_criterionAssessmentMap', RubricCriterionAssessmentEntity);
 	}
 
+	connectedCallback() {
+		super.connectedCallback();
+		window.addEventListener('rubric-update-criterion-assessment-entity-map', (e) => this._updateCriterionAssessmentMap(e));
```
devpow112, line 57: Does this need to be `window.` or can it just be `this.`?

src/components/rubric-criteria-groups.js
```diff
@@ -52,6 +52,16 @@ class RubricCriteriaGroups extends composeMixins(
...
+	}
+
+	disconnectedCallback() {
+		window.removeEventListener('rubric-update-criterion-assessment-entity-map', (e) => this._updateCriterionAssessmentMap(e));
```
devpow112, line 61: Does this need to be `window.` or can it just be `this.`?

src/components/rubric-criteria-groups.js
```diff
@@ -52,6 +52,16 @@ class RubricCriteriaGroups extends composeMixins(
...
+		super.disconnectedCallback();
```
devpow112, line 62: Should this be the first thing in the function? Usually `super.` stuff is the very first thing.

src/components/rubric-feedback.js
```diff
@@ -582,6 +596,13 @@ class RubricFeedback extends composeMixins(
 				hasPendingSaves: this._pendingFeedbackSaves > 0
 			}
 		}));
+		// we need to update criterion assessment entity map so that levels-mobile component
+		// gets updated select actions with overridden feedback values
+		window.dispatchEvent(new CustomEvent('rubric-update-criterion-assessment-entity-map', {
```
devpow112, line 601: Does this need to be `window.` or can it just be `this.`?

src/components/rubric-criteria-groups.js
imcdonald-d2l: The pattern we have been following in rubrics is to add `super.` at the end but only for `disconnectedCallback` functions. I don't know if it really matters as to the order in which it is called in the disconnected callback. There is no best practice documented in `lit` from what I can tell.

src/components/rubric-feedback.js
imcdonald-d2l: When I was testing locally I was unable to receive the event in `criteria-groups` component when using `this.` (hence why I used `window.` everywhere). I'll double check that's the case though

imcdonald-d2l: Confirmed that using `this.dispatchEvent` doesn't work locally. Something with the DOM structure must make the event not bubble up between these components.

devpow112: Did you try adding `bubbles: true`? I think that `bubbles` is false by default.

devpow112 (approved):

src/components/rubric-feedback.js
imcdonald-d2l: Totally forgot about that... it is working with `this.` now!

PR #1528, opened 2022-01-19
Ignore `@brightspace-ui*` dependencies as well

devpow112:
Updating these with dependabot can cause mangling.


jesseag (approved):
Looks good.

PR #1530, opened 2022-01-19
Kill dependabot

devpow112:
Removing dependabot since it can't help but just mangle stuff. It's wholly incapable of dealing with anything that contains git refs. The auto updater handles everything but major updates which I think will be fine to do manually. Security issue related dependabot fixes are still enabled as they seem to work better.


jesseag (approved):
Approved.

devpow112:
It sounds like there may be an effort to move use towards not using git refs for dependencies and instead use CodeArtifact. If that day comes we can re-introduce this as it will work if we have no git refs.

PR #1537, opened 2022-01-21
Moving to core list

MitchelPaulin:
Switched to d2l-list from dnd-sortable for the re-ordering capabilities. This included some required changes in d2l-list [see here](https://github.com/BrightspaceUI/core/pull/2029).

Note: v-diff will fail until d2l-list is updated


src/components/editor/rubric-criteria-group-editor.js
```diff
@@ -177,56 +186,56 @@ class RubricCriteriaGroupEditor extends composeMixins(
...
+								?is-holistic="${this.isHolistic}"
+								?is-reordered="${this.isReordered}"
+								?updating-levels="${this.updatingLevels}"
+								@d2l-criteria-count-changed="${this._updateCriterionCount}"
+							>
+							</d2l-rubric-criteria-editor>
+						</div>
 					</div>
 				</div>
-			</d2l-scroll-wrapper>
```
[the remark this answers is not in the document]
MitchelPaulin: Update: We do need the functionality that d2l-scroll-wrapper provides. Will look into adding this back as part of another ticket.

src/components/editor/rubric-criteria-editor.js
```diff
@@ -322,6 +268,8 @@ class RubricCriteriaEditor extends composeMixins(
 				composed: true,
 			}));
 		}
+
+		this._criteriaEntities = this._criteriaEntities.map((val, index) => { return { key: index.toString(), content: val }; });
```
jesseag, line 272: Do we know that entity.criterionHrefs will always be defined on Line 238?
Line 238 this._criteriaEntities = entity.criterionHrefs;

If it is not always defined then would we need an optional operator here similar to line 237? If entity.Hrefs is always defined then this comment doesn't really apply.
this._criteriaEntities = this._criteriaEntities?.map((val, index) => { return { key: index.toString(), content: val }; });

MitchelPaulin: A rubric should always have at least one criterion

jesseag: That makes sense.

jesseag (approved):
Looks okay to me. I don't know if you want to wait for a review from dbatiste first.

jesseag (approved):

devpow112:
@MitchelPaulin You'll need tp update the package-lock.json file to get the newer version of `core` that was released with your changes

MitchelPaulin:
The placement of the handle in `d2l-list` caused some elements to shift to the right, updating the goldens

jesseag (approved):

PR #1540, opened 2022-01-21
Fix reverse-levels feedback duplication - DE46822

imcdonald-d2l:
Visually, when reversing levels it appeared that feedback had been duplicated when it shouldn't. ( A refresh fixes it) The issue was that the parent's state was not being kept up-to-date with child changes so it was not triggering a re-render when it should have been. 

| Before | After | 
| --- | --- |
| reverse-before | reverse-after |


src/components/editor/rubric-feedback-editor.js
```diff
@@ -47,12 +47,6 @@ class RubricFeedbackEditor extends composeMixins(
 		`;
 	}
 
-	constructor() {
-		super();
-		this.keyLinkRels = ['self'];
-		this.criterionName = '';
-	}
```
imcdonald-d2l, line 54: Un-related to this defect but we shouldn't be setting public properties in a constructor. I checked and these are set properly in the parent so this won't break anything.

mpharoah-d2l: It looks like these never change. Should they just be constants?

imcdonald-d2l: I'm confused by what you mean. These are public properties set by the parent. They shouldn't be assigned to constants. 
image

mpharoah-d2l: Oh, I was looking at the test thinking it was part of the actual component. I see now

mpharoah-d2l (approved):

PR #1553, opened 2022-01-27
Default demos to lit on

devpow112:
Now that we are integrated into BSI and others default to showing the lit version for demos. Allow switching back to polymer. Vdiff and unit tests should be unaffected since we don't have switches on the page controlling what version of rubrics we use. It is always the lit version.


tools/is-lit-flag-enabled.js
```diff
@@ -1,25 +1,23 @@
...
+			url.searchParams.set('polymer', e.target.on ? 'true' : 'false');
 
 			document.location.href = url.toString();
 		});
 
-		if (litFlagValue === true) {
-			console.log('Lit Enabled');
+		if (polymerFlagValue === true) {
+			return !polymerFlagValue;
 		}
```
imcdonald-d2l, line 20: Do we need this condition? Looks like it won't always return a value with this condition set.

devpow112: I did it this way since it falls through to the console log message and return true below. Saved 1 line of code for the console log. I can change it if thats more readable though.

devpow112:
```js
if (polymerFlagValue !== true) {
  console.log('Lit Enabled');
}

return !polymerFlagValue;
```

MitchelPaulin (approved):

PR #1557, opened 2022-01-28
replace conditional CSS with javascript

zhouk3v:
https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fdefect%2F623145878275&fdp=true


jesseag (dismissed):

src/components/editor/rubric-criteria-groups-editor.js
```diff
@@ -291,8 +284,13 @@ class RubricCriteriaGroupsEditor extends composeMixins(
 			return;
 		}
 
+		const criteriaGroupPadding = styleMap({
+			paddingBottom: criteriaGroupIndex === this._criteriaGroupCount - 1 ? '0px' : '24px'
```
devpow112, line 288: When using `0` for CSS you can just use exactly `0` instead of adding the units. `0` is the same in all units 😄.

mpharoah-d2l:
What was the problem with the CSS? It looks like it's supported by everything that matters
https://developer.mozilla.org/en-US/docs/Web/CSS/:last-of-type

imcdonald-d2l:
> What was the problem with the CSS? It looks like it's supported by everything that matters https://developer.mozilla.org/en-US/docs/Web/CSS/:last-of-type

Kevin and I looked at this together and found that the `:last-of-type` was causing the padding to be set by `0` in all cases.... why Firefox wasn't supporting this properly is another question that I didn't look into.

MitchelPaulin:
> > What was the problem with the CSS? It looks like it's supported by everything that matters https://developer.mozilla.org/en-US/docs/Web/CSS/:last-of-type
> 
> Kevin and I looked at this together and found that the `:last-of-type` was causing the padding to be set by `0` in all cases.... why Firefox wasn't supporting this properly is another question that I didn't look into.

Strange, but if this method works then it looks fine to me

MitchelPaulin (dismissed):

zhouk3v:
> @zhouk3v I'm assuming this defect is related to the space problem in firefox right? I would just include the Rally defect URL as part of your PR. That way everyone knows what the issue this pull request is addressing.

Added, thanks for the reminder.

> > > What was the problem with the CSS? It looks like it's supported by everything that matters https://developer.mozilla.org/en-US/docs/Web/CSS/:last-of-type
> > 
> > 
> > Kevin and I looked at this together and found that the `:last-of-type` was causing the padding to be set by `0` in all cases.... why Firefox wasn't supporting this properly is another question that I didn't look into.
> 
> Strange, but if this method works then it looks fine to me

Some more context: the problem with the :last-of-type property just disappears whenever you try to use inspect element on the criteria. I couldn't find anything as to why Firefox doesn't support this unfortunately.

src/components/editor/rubric-criteria-groups-editor.js
```diff
@@ -291,8 +284,13 @@ class RubricCriteriaGroupsEditor extends composeMixins(
 			return;
 		}
 
+		const criteriaGroupPadding = styleMap({
+			paddingBottom: criteriaGroupIndex === this._criteriaGroupCount - 1 ? 0 : '24px'
```
MitchelPaulin, line 288:
```suggestion
			paddingBottom: criteriaGroupIndex === this._criteriaGroupCount - 1 ? '0' : '24px'
```
Not sure if it matters functionality wise, but this should be a string

MitchelPaulin (approved):

PR #1565, opened 2022-02-01
Fixed EntityWatcherMixin to clean up event listeners once it is done with them

mpharoah-d2l:
I don't think this was actually causing any problems, but it still needed to be fixed to prevent an update to an entity from an old href potentially causing issues with something that is now using a new different href.


devpow112 (approved):
Looks good to me :tada:

PR #1567, opened 2022-02-01
Fix DE45164 Screen reader not reading the correct terms when selectin…

jesseag:
…g/unselecting a cell.

Defect: https://rally1.rallydev.com/#/57732444928d/dashboard?detail=%2Fdefect%2F609063312323&fdp=true

When the user is focused on on a cell and selects an "Enter" key stroke, the screen reader should read out the level name and the status (selected/unselected) of the current level. 

Compromises: Due to the fact that I've had to use aria-live here to trigger screen reader updates and that the announce() function also uses aria-live the user will notice that sometimes the screen reader reads out multiple updates for the current cell. This was confirmed as an issue with the announce function by Dave B. 
https://d2l.slack.com/archives/C0PHG3QB0/p1643314330211600

For example if the user selects a cell then the screen reader will usually read the following: 

Level 4 unselected
Level 4 unselected
Level 4 selected

The last state that is read out will be correct one.


src/components/rubric-criteria-group.js
```diff
@@ -536,9 +537,8 @@ class RubricCriteriaGroup extends composeMixins(
 		return this._getSelfLink(criterionResult);
 	}
 	_getDescriptionCellAriaLabel(levelName, criterionName, criterionCell) {
-		// hack to announce the position of the description cell within the rubric for individuals using screen reader
-		// should be removed with the lit conversion and the new d2l-table component
-		const cellDescription = criterionCell.descriptionText;
+		const cellSelected = (this._getCriteriaAriaChecked(criterionCell, this.cellAssessmentMap) === 'true') ? 'selected' : 'unselected';
+		const cellDescription = criterionCell.descriptionText + cellSelected;
```
imcdonald-d2l, line 541: 1. Unnecessary brackets. Only needs to be `this._getCriteriaAriaChecked(criterionCell, this.cellAssessmentMap) === 'true'`
2. `cellSelected` needs to be localized so that it works in different translations.

src/components/rubric-criteria-group.js
```diff
@@ -670,6 +670,7 @@ class RubricCriteriaGroup extends composeMixins(
 	_handleKey(event) {
 		switch (event.code) {
 			case 'Enter':
+				return this._handleTap(event);
```
imcdonald-d2l, line 673: Is this necessary to add? I would think the previous code should still work:
```
case 'Enter':
case 'Space':
	return this._handleTap(event);
```

src/components/rubric-levels-mobile.js
```diff
@@ -503,6 +504,12 @@ class RubricLevelsMobile extends composeMixins(
 		});
 	}
 
+	_getLevelInfoStatus(index, criterionCellEntity, levelEntity) {
+		const isSelected = (this._selected >= 0 && index === this._selected) ? 'selected' : 'unselected';
```
imcdonald-d2l, line 508: Same comment as above. Unnecessary brackets (this case _might_ need them but I don't think it's necessary) and string needs to be localized.

src/components/rubric-levels-mobile.js
```diff
@@ -503,6 +504,12 @@ class RubricLevelsMobile extends composeMixins(
...
+		const levelText = this._getLevelText(criterionCellEntity, levelEntity);
+		return levelText + " " + isSelected;
```
imcdonald-d2l, line 510: This lint error can be fixed by doing the following:
```
`${levelText} ${isSelected}`
```

src/components/rubric-levels-mobile.js
```diff
@@ -896,7 +907,7 @@ class RubricLevelsMobile extends composeMixins(
 				@mousedown="${this._handleMouseDown}"
 				@focus="${this._handleFocus}"
 				@blur="${this._handleBlur}"
-				@keydown="${this._handleKey}"
+				@keydown="${(event) => this._handleKey(event, index, criterionCellEntity, levelEntity)}"
```
imcdonald-d2l, line 910: This lint error has always been a pain to work around. You can't pass any parameters into `this._handleKey` so you have to get the `index` by doing: `event.target.getAttribute('dataIndex')` which you then use to get the appropriate `criterionCellEntity` and `levelEntity`.

src/components/rubric-criteria-group.js
jesseag: No, it's not necessary, the previous code works. In the code above, case 'Enter' will then fall through to case 'Space. As long as no one changes the ordering of this switch statement (Ex: moves the arrowLeft/arrowRight key statements to the top) in the future then it will be fine to leave it as is.

imcdonald-d2l: I think it's fine to leave as is. Anyone modifying this switch in the future should know to adjust it as needed.

src/components/rubric-criteria-group.js
jesseag: I can put in the localization changes in en.js and update it here but I guess my question is, how can I go about doing this for all languages? Do I have to submit a request to the localization team?

devpow112: At the end of the dev cycle the translation stuff is all handled automatically but what we can serge. Also if you just add it to `en.js` CI has some stuff in place that will open a new PR against this PR "normalizing" all our language files so they have the same number of langterms.

devpow112: https://github.com/Brightspace/d2l-rubric#addingupdating-lang-term

imcdonald-d2l: Suggestion: I think `deselected` is a better word to use compared to `unselected`.

jesseag: This is the requirement I had from Mark which he got from Fiona. I'm indifferent to the change, should I confirm with them prior to making this change? I honestly don't think it makes a big difference. 

> Expected behavior is for the screen reader to read "[level name], unselected" or "[level name], selected" when clicking enter to select or unselect a cell
>

src/components/rubric-levels-mobile.js
MitchelPaulin: Should we disable it? @devpow112 Using an arrow function like this seems like a perfectly reasonable thing to do

devpow112: I don't know if there are any issues that come out of doing this. The main reason it's part of the rules is because lit will try to automatically bind this to anything you put in event handlers. No idea if this will break anything, especially with lit 2.

src/components/rubric-levels-mobile.js
```diff
@@ -708,7 +715,11 @@ class RubricLevelsMobile extends composeMixins(
 			case KEY_CODES.RIGHT_ARROW:
 				return this._handleRight(event);
 			case KEY_CODES.ENTER:
-				return this._handleEnter();
+				this._handleEnter();
+
+				// announce level info and cell status (selected/unselected)
+				announce(this._getLevelInfoStatus(index, criterionCellEntity, levelEntity));
+				return;
```
MitchelPaulin, line 722:
```suggestion
                // announce level info and cell status (selected/unselected)
				announce(this._getLevelInfoStatus(index, criterionCellEntity, levelEntity));
				return this._handleEnter();
```

src/components/rubric-levels-mobile.js
```diff
@@ -708,7 +715,11 @@ class RubricLevelsMobile extends composeMixins(
...
 		}
```
MitchelPaulin, line 723: This `announce` call should be moved under the last case statement, something like 

```
case KEY_CODES.ENTER:
    announce(this._getLevelInfoStatus(index, criterionCellEntity, levelEntity));
    return this._handleEnter();
```

src/components/rubric-levels-mobile.js
MitchelPaulin, line 508: nit: `isSelected` implies this is a boolean when its a string, maybe something like `selectedText` is more appropriate

src/components/rubric-levels-mobile.js
devpow112: Probably to avoid these kind of non-obvious issues https://stackoverflow.com/a/36915968 around when this and other values will be captured.

src/components/rubric-criteria-group.js
```diff
@@ -536,9 +536,8 @@ class RubricCriteriaGroup extends composeMixins(
 		return this._getSelfLink(criterionResult);
 	}
 	_getDescriptionCellAriaLabel(levelName, criterionName, criterionCell) {
-		// hack to announce the position of the description cell within the rubric for individuals using screen reader
-		// should be removed with the lit conversion and the new d2l-table component
-		const cellDescription = criterionCell.descriptionText;
+		const cellSelected = this._getCriteriaAriaChecked(criterionCell, this.cellAssessmentMap) === 'true' ? 'selected' : 'unselected';
+		const cellDescription = criterionCell.descriptionText + cellSelected;
```
MitchelPaulin, line 540: Does there not need to be a space here? 

Something like 

```
const cellDescription = `{criterionCell.descriptionText} {cellSelected}`;
```

src/components/rubric-criteria-group.js
imcdonald-d2l: If the requirement came for Fiona then it's good with me 👍🏻

src/components/rubric-levels-mobile.js
devpow112: Looks like the workaround @imcdonald-d2l mentioned is what the lit docs recommend
https://lit.dev/docs/components/events/#listening-to-events-fired-from-repeated-templates

src/components/rubric-criteria-group.js
jesseag: I changed it back to the original way it was. I'm fine with that.

src/components/rubric-levels-mobile.js
jesseag: Hmm...this works just fine for the index. 

However, I need to add in a second html tag for the level attribute to construct the levelEntity in _handleKey as below.
const levelEntity = new RubricLevelEntity(level);

I've tried:

dataLevel="${level}"
dataLevel=${level}

and all I can see in the debugger that gets passed through to the event handler (_handleKey) is the string "[object Object]". Is there any different syntax that I can use for passing through an object like this?

src/components/rubric-levels-mobile.js
jesseag: The announce call is under the case statement (KEY_CODES.ENTER). Is this just GitHub being weird on formatting in the browser? Or maybe I am not understanding you here? I just did a quick check and other key strokes don't invoke this case statement. 

The only change I see is that you've made it to switch the order of announce/return this_handleEnter(). The reason I have the order set in this way is because this.getLevelInfoStatus relies upon this._selected which can get updated through this.handleEnter(). So I want the state of this.selected updated to the latest state prior to announcing the state to the user. I could potentially refactor this if we don't want to check on the state of this.selected.

src/components/rubric-levels-mobile.js
jesseag: Yeah I can agree on this one for sure as I ran into a similar thing in this code. We need to name functions better. I used a function originally in this PR called isSelected (which checks for whether it's selected as well as does an update of the state under the hood). The naming of the function leads you to think it returns the status. However, it does 2 things upon closer inspection which is confusing.

src/components/rubric-levels-mobile.js
MitchelPaulin: Nvm you're right, I was reading this like Python because of the spacing, its all within the switch braces so this is fine

src/components/rubric-levels-mobile.js
imcdonald-d2l: You can just call `this._levelEntities[index]` and `this.criterionCells[index]` once you have the index in the event handler. These shouldn't be bound to the component as properties

src/components/rubric-criteria-group.js
jesseag: The space has been added.

src/components/rubric-levels-mobile.js
```diff
@@ -503,6 +504,15 @@ class RubricLevelsMobile extends composeMixins(
 		});
 	}
 
+	_getLevelNameStatus(event) {
+		const index = parseInt(event.target.getAttribute('dataIndex'));
+		const levelEntity = this._levelEntities[index];
+
+		const selectedText = this._selected >= 0 && index === this._selected ? this.localize('rubricEditorAccessibility:selected') : this.localize('rubricEditorAccessibility:unselected');
+		const levelName = levelEntity?.properties?.name;
+		return `${levelName} ${selectedText}`;
```
mpharoah-d2l, line 513: We usually try to avoid concatenating lang terms because word order can be different in other languages.

jesseag: Hmm...this is definitely true especially with latin based languages. How can I resolve this? Do I need the localization team to give a suggestion here? I'm returning this value here because of the specifications from Mark (which he got from Fiona) saying that we should return "Level Name Selected status (selected/unselected).

jesseag:
Please check that the french translations look okay as well as they are the important ones. I've updated them to be the following:

Selected = Sélectionné
Unselected = Désélectionné 

This seems more in line with what we have in the rest of our code base (I took a look in search.d2l and then updated them as so).

imcdonald-d2l (approved):

src/components/rubric-levels-mobile.js
imcdonald-d2l: We already do this elsewhere in rubrics so I wouldn't bee too concerned about this case. For example, this lang term:
```
"rubricEditor:scoring": "Scoring: {method}",
```

MitchelPaulin (approved):

PR #1580, opened 2022-02-04
DE46966 -- Fixed the inline rubric assessment view not updating after the popout view is closed

mpharoah-d2l:
https://rally1.rallydev.com/#/57732444928d/dashboard?detail=%2Fdefect%2F623239637233


src/components/rubric-levels-mobile.js
```diff
@@ -461,15 +468,30 @@ class RubricLevelsMobile extends composeMixins(
...
+			if (selectedLevel === this._levelEntities[this._selected]?.href) {
+				selectedLevel = null;
+			}
 		}
-		this.onCriterionSelect(entityGetter, () => {
+
+		this.ouDispatcher.selectLevel(this.criterionHref, selectedLevel.href);
+		this.CriterionCellAssessmentHelper.selectAsync(entityGetter).then(() => {
+			if (--this._pendingSaves === 0) {
+				this.ouDispatcher.rollback(this.criterionHref);
```
mpharoah-d2l, line 489: This syncs the state back up with the entity once all pending requests are completed, so it can get updates from the popout view again

devpow112 (approved):
Looks good to me :tada:

src/components/rubric-levels-mobile.js
```diff
@@ -401,7 +408,7 @@ class RubricLevelsMobile extends composeMixins(
 		}
 	}
 
-	deselect(callback = noop) {
+	deselect() {
```
imcdonald-d2l, line 411: `overall-score-mobile` expects a callback function here to update overall score on deselect. I added this since overall score wasn't updating properly on deselect. If this is no longer needed than it should be removed from overall-score-mobile as well. But should be tested that deselecting overall-score updates everything as expected before removing this callback.

imcdonald-d2l (changes requested):

src/components/rubric-levels-mobile.js
mpharoah-d2l: I think that should happen automatically because of the link hreaders in the siren response, but I'll check it out

imcdonald-d2l: Yeah, if it's working without the callback now that's great. We can just remove the callback in `overall-score-mobile` so it isn't passing a callback unnecessarily if it is working.

imcdonald-d2l (approved):

devpow112 (approved):

PR #1589, opened 2022-02-07
Changed to group-right padding now that we dont have to account for scroll bar in padding

MitchelPaulin:
https://rally1.rallydev.com/#/57732444928d/defects?detail=%2Fdefect%2F623819065683


devpow112 (approved):
Horray!!!!

PR #1593, opened 2022-02-07
Standards Firefox dialog display issue - DE47142

imcdonald-d2l:
Spoke to Dave B. and issue was because of relative z-index issues and firefox not supporting the default dialog as nicely. Easiest solution was move dialog to a parent component so we have one outcomes dialog for all of rubrics (since multiple dialogs is unnecessary). I just chose to put it in `d2l-rubric-criteria-groups-editor` because it seemed like a good spot. This gives us the added benefit by not rendering a dialog for each criterion if the rubric is very large. Dave B send dialogs can be quite expense to render so this will be a nice improvement. 

| Before | After |
| --- | --- |
| dialog-before | dialog-after |


src/components/editor/rubric-criteria-groups-editor.js
```diff
@@ -344,6 +365,28 @@ class RubricCriteriaGroupsEditor extends composeMixins(
 
 		return html`<d2l-rubric-loading></d2l-rubric-loading>`;
 	}
+	_renderOutcomesDialog() {
+		return html`
+			<d2l-dialog
+				id="dialog-outcomes"
+				width="800"
+				title-text="${this.context?.langterms?.['browseOutcomes']}"
+				async
```
imcdonald-d2l, line 374: Since we now only have one dialog it needs to be async. This shows us a loading spinner while we wait for alignment list to load (until we set `asyncState = asyncStates.complete`) and adjusts the dialog height as expected.

devpow112 (approved):

MitchelPaulin:
LGTM, this will cause DOM changes though so make sure you loop in @jzhang52 @Davern https://learningoutcomes.slack.com/archives/CANSREZKJ/p1643817801756369

PR #1597, opened 2022-02-08
[DRAFT] RTL vdiff - US135300

imcdonald-d2l:


src/components/editor/rubric-criteria-group-editor.js
```diff
@@ -18,6 +18,8 @@ import { defineComponent } from '../../helpers/define-component.js';
 import { EntityMixin } from '../../mixins/entity-mixin.js';
 import { EntityWatcherMixin } from '../../mixins/entity-watcher-mixin.js';
 import { LocalizeMixin } from '../../mixins/localize-mixin.js';
+// eslint-disable-next-line no-duplicate-imports
+import { offscreenStyles } from '@brightspace-ui/core/components/offscreen/offscreen.js';
```
devpow112, line 22: You can just remove the above `import '@brightspace-ui/core/components/offscreen/offscreen.js';` import. This import should result in the component being defined as well.

src/components/editor/rubric-criteria-group-editor.js
```diff
@@ -171,11 +166,11 @@ class RubricCriteriaGroupEditor extends composeMixins(
 		return html`
 			${this._renderAlerts()}
 			${this._renderDialog()}
-				<h2 class="screen-reader">${this._getGroupHeadingText(this._groupName, this.showGroupName)}</h2>
+				<h2 class="d2l-offscreen">${this._getGroupHeadingText(this._groupName, this.showGroupName)}</h2>
 				<d2l-offscreen>${this._getRubricStructureLabel(this._levelCount, this._criterionCount)}</d2l-offscreen>
```
devpow112, line 170: You could also change this to be a div with the off-screen class if you want to have 1 less component.

devpow112: Oh sorry I meant changing the `d2l-offscreen` to div as well. The `d2l-offscreen` component is just a simple wrapper and should be avoided in favour of the class style thing if we can.

devpow112: You can if you like. Saves the time having to render the custom component everywhere.

imcdonald-d2l: I'll leave it out of this PR then to minimize the diff then

imcdonald-d2l:
Overall score and holistic border radius issues are issues in the `d2l-input-textarea` component and don't happen in production it seems since they use the `d2l-htmleditor`. Ignoring these issues as they don't seem worth fixing.
image

devpow112:
> Overall score and holistic border radius issues are issues in the `d2l-input-textarea` component and don't happen in production it seems since they use the `d2l-htmleditor`.

Once me move away from the input text area and old html editor this should go away like you said.

PR #1598, opened 2022-02-08
Clean-up vdiff

devpow112:
Simplifies vdiff set-up using learnings from user progress RTL issues.


devpow112:
Assuming this works (worked locally) @imcdonald-d2l this may help you simplify your RTL stuff. If you'd rather get your stuff in first then I can do this I'm fine with that as well.

devpow112:
NM this seems to have issues with total score. Not sure why it's not coming up locally but will abandon this for now.

PR #1601, opened 2022-02-08
DE47071 - fix criterion sliders unselecting

zhouk3v:
[Rally ticket](https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fdefect%2F624076745073&fdp=true)

Main cause: If the user drags out the criterion slider without changing the level,  `dragBackToSameCriterion` in `_handleDragEnd()` will evaluate to false. This will then fire off the `_selectLevelAt()` function with the current level as the argument, which will then deselect the criterion.

Fixes:
 - Added `startLevel` and `endLevel` properties to `_dragContext` to store the original level of the criterion and the current level that the user is on when dragging the slider
 - `dragBackToSameCriterion` now compares `startLevel` and `endLevel` from `dragContext` instead of checking if the cursor is within bounds of the original level's radio button.


src/components/rubric-levels-mobile.js
```diff
@@ -656,10 +657,7 @@ class RubricLevelsMobile extends composeMixins(
 		}
 		const isDragEvent = this._dragContext.isDragEvent;
 		const isDeselectingCriterion = this._focused === this._selected;
-		const dragBackToSameCriterion = e.pageX >= this._dragContext.radioBounds.left &&
-			e.pageX <= this._dragContext.radioBounds.right &&
-			e.pageY >= this._dragContext.radioBounds.top &&
-			e.pageY <= this._dragContext.radioBounds.bottom && isDragEvent;
+		const dragBackToSameCriterion = this._dragContext.startLevel === this._dragContext.endLevel && isDragEvent;
```
MitchelPaulin, line 660: Should this be `e._dragContext.startLevel` ?

MitchelPaulin (approved):

PR #1603, opened 2022-02-08
Disconnect `ResizeObservers`

devpow112:
We should always disconnect them just in case there is lingering stuff that needs to be flushed. Also added a few missing `super` calls.


src/components/editor/rubric-editor.js
```diff
@@ -366,6 +366,8 @@ class RubricEditor extends composeMixins(
 	}
 
 	firstUpdated() {
+		super.firstUpdated();
```
devpow112, line 369: Should always super function we are overloading. This is a lit function so added the `super` call.

src/components/rubric-levels-mobile.js
```diff
@@ -332,13 +332,21 @@ class RubricLevelsMobile extends composeMixins(
...
+
 		window.removeEventListener('mouseup', this._handleDragEnd);
 		super.disconnectedCallback();
 	}
 
 	firstUpdated() {
+		super.firstUpdated();
+
+		this._resizeObserver = new ResizeObserver(() => this._handleResize());
+		this._resizeObserver.observe(this.shadowRoot.host);
```
devpow112, line 348: Not sure why we are watching the `shadowRoot` host instead of just the shadow root. Guessing we want to see when our parent change.

src/components/rubric.js
```diff
@@ -294,6 +299,8 @@ class Rubric extends composeMixins(
 	}
 
 	firstUpdated() {
+		super.firstUpdated();
```
devpow112, line 302: Were missing this again.

src/components/editor/rubric-levels-editor.js
```diff
@@ -274,6 +271,22 @@ class RubricLevelsEditor extends composeMixins(
 		}
 	}
 
+	_handleResize() {
```
devpow112, line 274: Renamed from `_checkSize` to just be more consistent with resize handlers.

src/components/editor/rubric-levels-editor.js
```diff
@@ -274,6 +271,22 @@ class RubricLevelsEditor extends composeMixins(
...
+		afterNextRender(() => {
```
devpow112, line 275: Not sure if the `afterNextRender` is needed here but leaving it as it was.

imcdonald-d2l (approved):

PR #1612, opened 2022-02-09
Silenced localization lint warning

### LINKED ISSUE #1605: Get rid of plural warning
We are getting a split plural warning for
```
"rubricEvaluationAccessibility:competencyIconLabel": "Competency alignment indicator. {name} {number, plural, =1 {competency aligned to} other {competencies aligned to}} {alignedActivity}"
```

MitchelPaulin:
The format we had was correct but the linter didn't like it, this is an equivalent format that the linter can parse more easily . Resolves #1605.


devpow112 (approved):
Thanks for taking care of this. Filed an issue about it yesterday. Updated the description to include it so it will be auto closed when this merges.

PR #1617, opened 2022-02-09
Fixed total score not updating issue

MitchelPaulin:
I guess there is a problem with using getters and setters and lits ability to detect changes and re-render, my guess is it doesn't access the property through the normal channels so never goes through the getter and from lits perspective nothing has changed. 

Will need to keep this in mind going forward.


src/components/editor/rubric-criteria-groups-editor.js
```diff
@@ -338,7 +330,7 @@ class RubricCriteriaGroupsEditor extends composeMixins(
 		}
 
 		let totalScoreText = null;
-		if (this._hasTotalScore) {
+		if (this.totalScore !== 'undefined') {
```
devpow112, line 333: Not sure if you need the quotes.

```suggestion
		if (this.totalScore !== undefined) {
```

MitchelPaulin: You do, since it gets converted to a string when getting passed down from the parent

devpow112: Or does it really need to look for the string of `undefined`?

devpow112 (dismissed):

imcdonald-d2l (commented):
<del>Does this not need to be updated to `this._totalScore` now that get/set functions are removed?</del>

Edit: nevermind, I see that it is a public property

imcdonald-d2l (approved):

src/components/editor/rubric-criteria-groups-editor.js
devpow112: Could you not just change [this](https://github.com/Brightspace/d2l-rubric/blob/3a0bb9455b3990bcab49b89146e78457730cbf4c/src/components/editor/rubric-structure-editor.js#L392) to be `total-score="${ifDefined(this._totalScore)}"`. Would cause it to just not be sent if it's undefined.

MitchelPaulin: That seems to work, I can add it

devpow112 (approved):

PR #1621, opened 2022-02-09
Firefox - points not updating properly - DE47189

imcdonald-d2l:
[Rally ticket](https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fdefect%2F626668256949)
| Before | After |
| --- | --- |
| delete-before | delete-after |

### Issue Explanation
**Note**: fix is pretty hacky so explaining why I did what I did. If you think of a better solution please let me know.

**Basic stack trace:**

1. Dialog opens and you click `Delete`
2. This triggers the `_handleClose()` function in the dialog-mixin which tries to focus the opener element [here](https://github.com/BrightspaceUI/core/blob/main/components/dialog/dialog-mixin.js#L244)
3. This passes it to a core focus helper which checks if the element is focusable [here](https://github.com/BrightspaceUI/core/blob/8edcb36fd32f67e86dc37f35745195f440ddda8c/helpers/focus.js#L186).
   * For chrome, the element passed in is the `<body>` element and focuses the body. This is it's own problem - it won't focus the icon when cancelling delete action.
   * For Firefox, the element passed is the delete button element which we would expect is passed to this helper function. `isFocusable()` ends up returning `false` [here](https://github.com/BrightspaceUI/core/blob/8edcb36fd32f67e86dc37f35745195f440ddda8c/helpers/focus.js#L178) since `node.offsetParent === null`. Not sure why this is the case here 🤔 
4. (Firefox only path) Since the delete button is considered un-focusable it looks for the closest element to focus [here](https://github.com/BrightspaceUI/core/blob/8edcb36fd32f67e86dc37f35745195f440ddda8c/helpers/focus.js#L190). It then grabs the points `d2l-rubric-autosaving-input` element and focuses that instead.
5. This sets `this._isFocused` to true and the actual updated value won't update now since the element thinks it's focused. 

### Fix
Reset focus on points input manually if delete is successful. 

### Discussion
This solution is certainly not ideal. If anyone has any ideas please let me know. Otherwise, should we loop in someone from Gaudi?


jesseag:
It seems like a reasonable solution to me. I don't know that I have a ton of additional feedback for improvements.

jesseag (approved):

jesseag:
The other possible thing you could do would be to grab the elementID and blur it. But this is not really much different than what you are doing right now. You would still be using the same structure/flow. 
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/blur

imcdonald-d2l:
> Not sure if this would work and if its feasible. I would assume we would have to abandon using the core functionality (_handleClose) and the dialog mixin that you refer to above in step 2. But it's a possible idea.

Thanks for the suggestion! Unfortunately I think this will make things more complex and unsure of if it's feasible given the focus happens from a dialog-mixin not from clicking the delete button. I'll merge this as is for now since it seems to work as expected with this change.

PR #1629, opened 2022-02-10
Fix DE47238 Criterion always resets to the original state when toggle…

jesseag:
…d via the enter key.

Defect: https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fdefect%2F626889360247
DE47238

**Old Behaviour:** _executeSelectAction() is run in the _queueAction() function and then runs again via the debouncer after 0.75 seconds has elapsed. This causes the cell to be selected twice which returns it back to the original state.
**New Behaviour:** Adding in an else statement so we don't have 2 calls to _executeSelectAction() in _queueAction().

@imcdonald-d2l 
Please take a quick look at this as I'm wondering if my change here could cause issues with your last changes for the Rapid select/deselect fixes.
https://github.com/Brightspace/d2l-rubric/commit/09a5fa6c69fdb941678acd7da25b322bd9245f07


imcdonald-d2l (approved):
This makes sense, definitely an oversight on my part. Shouldn't interfere with rapid select/deselect.

I wouldn't worry about the small package-lock change. I would just leave it.

MitchelPaulin (approved):

devpow112 (approved):

PR #1639, opened 2022-02-15
BREAKING CHANGE: Lit 2 migration: bumping dependencies that use Lit to next major versions 

annabelsegalld2l:
This change updates the dependencies that use Lit to their next versions. Note that this PR should not merge until the dependencies have been released and we do the Lit 2 switchover. The CI will continue to fail until that time, as the dependencies are not yet available.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "d2l-rubric",
-  "version": "3.84.0",
+  "version": "5.0.0",
```
devpow112, line 3: Will make manual release.

devpow112 (approved):

PR #1640, opened 2022-02-15
DE47036 - Fix showing of `null` button of partially editable alert

devpow112:
Just needed to use `ifDefined` to not add the attribute if the rubric is read-only. Rally [DE47036](https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fdefect%2F623806244905).

**Before Fix (w/o Permission)**
image

**After Fix (w/o Permission)**
image

With Permission
image


src/components/editor/rubric-editor.js
```diff
@@ -766,15 +767,15 @@ class RubricEditor extends composeMixins(
...
 				id="locked-alert"
 				class="rubric-editor-alert"
 				type="default"
-				button-text="${this._isReadOnly ? null : this.localize('rubricEditor:editTexts')}"
-				@d2l-alert-button-press="${this._partialUnlock}"
-			>
-				<d2l-icon icon="d2l-tier1:lock-locked"></d2l-icon>
+				button-text="${ifDefined(this._isReadOnly ? undefined : this.localize('rubricEditor:editTexts'))}"
+				@d2l-alert-button-press="${this._partialUnlock}">
+				<d2l-icon icon="tier1:lock-locked"></d2l-icon>
```
devpow112, line 778: Removed `d2l-` from the icon name as it's not required. Probably will do this everywhere eventually.

MitchelPaulin (approved):

PR #1654, opened 2022-02-17
DE47039 fixes

zhouk3v:
[Rally Ticket](https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fdefect%2F623814566427&fdp=true)

Fixes regarding the total score and overall score components with JAWS and NVDA

Main Changes:
 - Removed the total-grouping-label `d2l-offscreen` element in `rubric-total-score.js` and the associated lang term. This was causing JAWS to not read the total score off properly
 - Moved overallScore aria-label up and added parameters to lang terms for the overall level and description in `rubric-overall-score.js`. The level that NVDA read from the last fix was apparently the level of the heading element.
 
~~Compromises:~~
~~- Overall level description is read twice in NVDA~~


src/components/rubric-overall-score.js
```diff
@@ -184,8 +184,7 @@ class RubricOverallScore extends composeMixins(
 				<span>${this.localize('rubricEvaluation:overallScore')}</span>
 				${this._renderCompetenciesIcon()}
 			</h3>
-			<d2l-scroll-wrapper aria-labelledby="overall-grouping-label">
-				<d2l-offscreen id="overall-grouping-label">${this.localize('rubricEvaluation:overallScore')}</d2l-offscreen>
+			<d2l-scroll-wrapper>
```
zhouk3v, line 187: This `d2l-offscreen` element causes JAWS to read off "Overall Score" twice

src/components/rubric-overall-score.js
```diff
@@ -306,9 +305,9 @@ class RubricOverallScore extends composeMixins(
 				@click="${this._levelClicked}"
 				@keypress="${this._handleKeypress}"
 				tabindex="${this._isAchieved(overallLevel) && this.overallLevelAssessmentHref ? 0 : -1}">
-				<div class="content-container" id="${index}">
+				<div class="content-container" id="${index}" aria-label="${this.localize('rubricEvaluationAccessibility:overallScore', 'overallScore', this._getLevelName(overallLevel), 'overallScoreDescription', overallLevel.descriptionText)}">
 					<div class="content-header" id="${index}">
-						<h4 class="d2l-heading-4" id="${index}" aria-label="${this.localize('rubricEvaluationAccessibility:overallScore')}">${this._getLevelName(overallLevel)} </h4>
+						<h4 class="d2l-heading-4" id="${index}" aria-hidden="true">${this._getLevelName(overallLevel)} </h4>
```
zhouk3v, line 310: The `aria-hidden` property prevents NVDA from reading the overall level twice

src/components/rubric-total-score.js
```diff
@@ -99,10 +99,7 @@ class RubricTotalScore extends composeMixins(
 
 		return html`
 			<div class="out-of-container" tabindex="0">
-				<div class="out-of-text d2l-body-compact" role="group" aria-labelledby="total-grouping-label">
-					<d2l-offscreen id="total-grouping-label">
-						${this.localize('rubricEvaluationAccessibility:totalScoreLabel')}
```
devpow112, line 104: Is this langterm still in use or can it be removed from all the lang files?

devpow112: NM I'm just blond you removed it already

zhouk3v:
> > Compromises: Overall level description is read twice in NVDA
> 
> Where you not able to add an `aria-hidden` attribute somewhere to prevent this?

I could add `aria-hidden` to either the div element or the d2l-html-block element here in `rubric-overall-score.js`

```
<div class="info-container d2l-body-small" id="${index}" aria-hidden="true">
  ${this._localizePoints(overallLevel)}
    <d2l-html-block id="${index}">
	  ${unsafeHTML(overallLevel.descriptionHtml)}
    </d2l-html-block>
</div>
```
Adding it to either section causes JAWS to read out "Overall Level {level} {level description} heading level 3"  I don't know if that is necessarily better than NVDA reading the level description twice.

Note: Originally JAWS would only say "Overall Level" but that was fixed when the `d2l-offscreen` element in `rubric-overall-score.js` was removed.

lang/en.js
```diff
@@ -184,7 +184,7 @@ export default {
 	"rubricEvaluationAccessibility:feedbackOn": "Feedback on {criterionName}",
 	"rubricEvaluationAccessibility:noDescription": "No description provided",
 	"rubricEvaluationAccessibility:noMoreLevels": "No more levels",
-	"rubricEvaluationAccessibility:overallScore": "Overall Score",
+	"rubricEvaluationAccessibility:overallScore": "Overall Score {overallScore} {overallScoreDescription}",
```
zhouk3v, line 187: Should I go through the rest of the lang files and manually add these params? I don't think the automated PR managed to add these in

imcdonald-d2l:
> Adding it to either section causes JAWS to read out "Overall Level {level} {level description} heading level 3" I don't know if that is necessarily better than NVDA reading the level description twice.

What is the `heading level 3` coming from? Level description could be quite long in some circumstances so it would be best to avoid reading it twice if possible.

zhouk3v:
> What is the `heading level 3` coming from? Level description could be quite long in some circumstances so it would be best to avoid reading it twice if possible.

The `heading level 3` comes from this part of the code here:
```
<h3 class="d2l-heading-3">
    <span>${this.localize('rubricEvaluation:overallScore')}</span>
    ${this._renderCompetenciesIcon()}
</h3>
```

Putting `aria-hidden` in either element here makes JAWS read out "Overall Level {level} {level description} seperator". Still trying to find out why it does that.

zhouk3v:
> The `heading level 3` comes from this part of the code here:
> 
> ```
> <h3 class="d2l-heading-3">
>     <span>${this.localize('rubricEvaluation:overallScore')}</span>
>     ${this._renderCompetenciesIcon()}
> </h3>
> ```
> 
> Putting `aria-hidden` in either element here makes JAWS read out "Overall Level {level} {level description} seperator". Still trying to find out why it does that.

Ok managed to find out that JAWS was reading a `<hr>` element in `rubric.js`. I've added `aria-hidden` properties to both the `<hr>` element and the `<span>` element in the previously mentioned `<h3>` element. Both NVDA and JAWS should now read "Overall Level {level} {level description}"

mpharoah-d2l (approved):

PR #1660, opened 2022-02-22
Swap all occurrences of `.then()` `.catch()` `.finally()` with try/catch statements

imcdonald-d2l:
Swapping all occurrences of `.then()` `.catch()` `.finally()` with try catch statements. This allows us to implement `async` functions which we can `await` in unit tests so that we can ensure code has executed as expected.

**Note:** having `async` functions with `try/catch` is the pattern we should be adopting moving forward so that we can can better control things in unit tests. 

**Caveat**: not replacing `.then` references in [entity-watcher-mixin](https://github.com/Brightspace/d2l-rubric/blob/master/src/mixins/entity-watcher-mixin.js#L32) as they are more complex uses that shouldn't be necessary to fix for unit tests 🤞🏻


src/components/editor/rubric-criteria-editor.js
```diff
@@ -103,48 +103,48 @@ class RubricCriteriaEditor extends composeMixins(
...
+			} catch (err) {
 				this.dispatchEvent(new CustomEvent('d2l-rubric-editor-save-error', {
 					detail: {
 						message: err.message,
 					},
 					bubbles: true,
 					composed: true,
 				}));
-			}).then(() => {
+			} finally {
```
devpow112, line 117: Why the change to finally here?

src/components/editor/rubric-criteria-group-editor.js
```diff
@@ -273,27 +275,27 @@ class RubricCriteriaGroupEditor extends composeMixins(
...
+			} catch (err) {
 				this.dispatchEvent(new CustomEvent('d2l-rubric-editor-save-error', {
 					detail: {
 						message: err.message,
 					},
 					bubbles: true,
 					composed: true,
 				}));
-			}).then(() => {
+			} finally {
```
devpow112, line 286: Why the change to finally here?

src/components/rubric-editable-score.js
```diff
@@ -505,75 +505,60 @@ class RubricEditableScore extends composeMixins(
...
 				} else {
 					e.target.value = this.score;
 					return;
 				}
 			}
-
-			this._scoreSaves++;
-			this.score = newScore;
-
-			action.catch(error => {
```
devpow112, line 566: What happened to this code?

src/components/editor/rubric-criteria-editor.js
imcdonald-d2l: Isn't necessary. Good catch.

src/components/editor/rubric-criteria-group-editor.js
imcdonald-d2l: Also not necessary 👍🏻

src/components/editor/rubric-level-editor.js
```diff
@@ -363,24 +366,25 @@ class RubricLevelEditor extends composeMixins(
...
 						bubbles: true,
 						composed: true,
 					}));
 					this._updateName(this.entity);
-				}).catch((err) => {
+				} catch (err) {
 					this._nameInvalid = true;
 					this._nameInvalidError = tryGetSirenErrorMessage(err) || this.localize('rubricEditor:nameSaveFailed');
-				}).then(() => {
+				} finally {
```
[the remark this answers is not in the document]
imcdonald-d2l: Not necessary, will fix 👍🏻

src/components/rubric-editable-score.js
imcdonald-d2l: This code got moved to `_handleScoreFinally` and `_handleScoreSaveError` below. They are executed on the actions in `this._saveScore(newScore)` and `this._clearOverride()`.

devpow112: Ah I see. Ok cool. Thanks for the explanation.

PR #1661, opened 2022-02-22
Inline grading unit tests US136165 - PT 1

imcdonald-d2l:


package.json
```diff
@@ -20,6 +20,7 @@
     "start:vdiff": "wds --node-resolve --watch --open test/perceptual/ --config wds.config.js",
     "test": "npm run lint && npm run test:headless && npm run test:axe",
     "test:axe": "wtr --group aXe --config wtr.config.js",
+    "test:test": "wtr --group test --config wtr.config.js",
```
imcdonald-d2l, line 23: ignore this command. will delete later. just using it locally

test/d2l-rubric-editable-score.test.js
```diff
@@ -0,0 +1,68 @@
...
+
+			const scoreInput = element.shadowRoot.querySelector('#text-area');
+			expect(scoreInput.value).to.equal(pointsRubricAssessmentEntity?.properties?.score.toString());
+
+			const outOfDiv = element.shadowRoot.querySelector('#out-of');
+			expect(outOfDiv.innerText).to.equal(`/ ${pointsRubricEntity.outOf}`);
+		});
+
+		it('verify total score override action is sent', async() => {
+			const stub = sandbox.stub(RubricEditableScore.prototype, '_execAssessmentAction').callsFake(() => Promise.resolve());
```
imcdonald-d2l, line 46: Stubbing out function so that it doesn't execute any actions.

test/d2l-rubric-editable-score.test.js
```diff
@@ -0,0 +1,68 @@
...
+			`);
+
+			const scoreOverride = 123;
+			// fake total-score input
+			element._handleChange();
+			await element._handleBlur({ target: { value: scoreOverride } });
+			expect(stub.callCount).to.equal(1);
+			expect(stub.args[0][0]().name).to.equal('override-total-score');
+			expect(stub.args[0][1].OverallScore).to.equal(scoreOverride);
+			expect(stub.args[0][1].TotalScoreOverridden).to.be.true;
```
imcdonald-d2l, line 64: Verifying arguments passed are as expected.

src/components/rubric-editable-score.js
```diff
@@ -20,7 +20,7 @@ import { RubricCriterionEntity } from '../entities/rubric-criterion-entity.js';
 import { RubricEntity } from '../entities/rubric-entity.js';
 import { styleMap } from 'lit-html/directives/style-map.js';
 
-class RubricEditableScore extends composeMixins(
+export class RubricEditableScore extends composeMixins(
```
imcdonald-d2l, line 23: Necessary to stub out function 😢

package.json
```diff
@@ -80,6 +81,7 @@
     "puppeteer": "^13",
     "query-selector-shadow-dom": "^1",
     "sinon": "^13",
+    "siren-parser": "^8.5.1",
```
devpow112, line 84: Can we change this to just `"siren-parser": "^8",`, will match other uses of it then.

wtr.config.js
```diff
@@ -13,6 +13,10 @@ export default merge(
 			{
 				name: 'aXe',
 				files: getPattern('axe')
+			},
+			{
```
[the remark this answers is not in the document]
imcdonald-d2l: Yeah, this will get removed

test/d2l-rubric-adapter.test.js
```diff
@@ -0,0 +1,83 @@
...
+		it('expands/collapses accordion when clicked', async() => {
+			const element = await fixture(html`
+				<d2l-rubric-adapter
+					compact
+				></d2l-rubric-adapter>
+			`);
+
+			const accordionElement = element.shadowRoot.querySelector('d2l-labs-accordion-collapse');
+			const accordionTrigger = accordionElement.shadowRoot.querySelector('#trigger');
+			expect(accordionElement.opened).to.be.false;
```
devpow112, line 53: Seems like the lit linter doesn't like this line. I can look into disabling it for testing as this seems like a valid use for testing.
```
./test/d2l-rubric-editable-score.test.js
   53:6    error              You are using a property binding without an expression
```

test/d2l-rubric-editable-score.test.js
```diff
@@ -0,0 +1,68 @@
...
+
+		it('verify total score override action is sent', async() => {
+			const stub = sandbox.stub(RubricEditableScore.prototype, '_execAssessmentAction').callsFake(() => Promise.resolve());
+
+			const element = await fixture(html`
+				<d2l-rubric-editable-score
+					.entity="${pointsRubricEntity}"
+					assessment-href="${PointsRubricAssessmentHref}"
+					.totalScore=${pointsRubricAssessmentEntity?.properties?.score}
+					.token="test-token"
```
devpow112, line 53: The lit analyzer doesn't like this line, does it get rid of the issue if you change it to `.token="${'test-token'}"`?

test/d2l-rubric-editable-score.test.js
```diff
@@ -0,0 +1,68 @@
...
+		it('can instantiate total score', async() => {
+			const element = await fixture(html`
+				<d2l-rubric-editable-score
+					.entity="${pointsRubricEntity}"
+					assessment-href="${PointsRubricAssessmentHref}"
+					.totalScore=${pointsRubricAssessmentEntity?.properties?.score}
+				></d2l-rubric-editable-score>
+			`);
+
+			expect(element).to.not.be.undefined;
```
MitchelPaulin, line 35: Is there a way to assert *is defined* rather than not undefined? As is this will pass if element is `null` which is probably not what we want

test/d2l-rubric-editable-score.test.js
imcdonald-d2l: I'll give that a shot

MitchelPaulin: Or just `token="test-token"` should work

test/d2l-rubric-editable-score.test.js
devpow112: expect().to.exist?
https://www.chaijs.com/api/bdd/#method_exist
Looks like they don't recommend it though due to JavaScript's fun type coercion

imcdonald-d2l: Agreed we don't want `null` as well but we do check `expect(element.localName).to.equal('d2l-rubric-editable-score');` right after and that would blow up in both cases. So we technically don't even need the `expect(element).to.not.be.undefined;` check

devpow112: Maybe try instacneOf?
https://www.chaijs.com/api/bdd/#method_instanceof

MitchelPaulin: We could add a helper to utils maybe?

```
export const isDefined(element) {
     expect(element).to.not.be.undefined;
     expect(element).to.not.be.null;
}
```

MitchelPaulin: > Agreed we don't want `null` as well but we do check `expect(element.localName).to.equal('d2l-rubric-editable-score');` right after and that would blow up in both cases. So we technically don't even need the `expect(element).to.not.be.undefined;` check

Agreed in this case its fine, but in general it would be nice to have a general way to assert its not null or undefined.

imcdonald-d2l: So testing locally and this seems to work just fine:
`expect(element).to.be.an.instanceof(RubricEditableScore);` 

So if we export all the component class definitions this is probably the easiest way to make sure it is defined properly.

MitchelPaulin: This works when we have access to the object type, but what about asserting something like fetching a button instance is defined? Do we want to have to import the button from core to assert that?

imcdonald-d2l: That's a good point. we can just use `.exist` then

imcdonald-d2l:
@devpow112 @MitchelPaulin you guys okay if this is merged so people have access to a shared utility and a couple of example tests? Still more tests to right for inline view but figured this would be useful to merge as-is

devpow112 (approved):
Looks good to me :tada:

PR #1663, opened 2022-02-22
Updating Translations 2203 22/02/2022 (`20.22.3`)

devpow112:
Backport of #1662 for `20.22.3`.


lang/pt.js
```diff
@@ -197,7 +197,7 @@ export default {
 	"rubricPrintView:activityNameLabel": "Atividade: {name}",
 	"rubricPrintView:courseNameLabel": "Curso: {name}",
 	"rubricPrintView:studentNameLabel": "Nome: {name}",
-	"rubricShared:dashOutOf": "— / {outOf}",
+	"rubricShared:dashOutOf": "—/{outOf}",
```
mpharoah-d2l, line 200: Did we want the spaces removed here?

devpow112: Some of the other languages do this. Left it cause wasn't sure if it was a conscious choice by the translator or not.

mpharoah-d2l: Interestingly, the Chinese  language updates below do the opposite 🤔
I think we probably want to preserve the whitespace

devpow112: I can revert it. I'll add it to the list of stuff that they'll need to manually fix in the serge database. Otherwise we'll keep getting it each release.

mpharoah-d2l (approved):

PR #1664, opened 2022-02-22
Create Tests - Part 1: Overall Level(s) Editor

MitchelPaulin:
This is part one only targeting the overall level editor. Open to suggestions for anything that could be added.


test/d2l-rubric-overall-level-editor.test.js
```diff
@@ -1,20 +1,60 @@
...
+
+		it('delete renders', () => {
+			const deleteButton = element.shadowRoot.querySelector('#remove');
+			expect(deleteButton).to.exist;
+		});
+
+		it('clicking delete works', async() => {
+			const deleteButton = element.shadowRoot.querySelector('#remove');
+			await deleteButton.click();
+			sinon.assert.calledOnce(deletedStub);
```
imcdonald-d2l, line 57: Not sure this is a valuable test to have. I don't believe verifying a specific function has been called doesn't have much value in itself. We should verify the effect the function has on the component. Open to other opinions though.

test/d2l-rubric-overall-levels-editor.test.js
```diff
@@ -1,20 +1,60 @@
...
+
+	describe('adding levels test', () => {
+		it('buttons presses work', async() => {
+			const appendButton = insideScrollWrapper.querySelector('#append-button');
+			const prependButton = insideScrollWrapper.querySelector('#prepend-button');
+			expect(appendButton).to.exist;
+			expect(prependButton).to.exist;
+
+			await appendButton.click();
+			sinon.assert.calledOnce(appendStub);
```
imcdonald-d2l, line 54: Same thoughts as above. Should we verify the effects of the function not just that the function has been called?

test/d2l-rubric-overall-level-editor.test.js
MitchelPaulin: I was thinking about this as well, could we mimic it having an effect? We wouldn't actually see anything get deleted until the call returns, in which case it would be better to test that as an hm-api-test test?

imcdonald-d2l: Ideally we check that `this.performSirenAction(action)` happed as expected when execution dialog action. Not sure how hard it is to work around the dialog though 🤔

test/d2l-rubric-overall-levels-editor.test.js
```diff
@@ -1,20 +1,66 @@
...
+	});
+
+	describe('adding levels test', () => {
+		it('append button works', async() => {
+			sirenActionStub.resetHistory();
+			const appendButton = insideScrollWrapper.querySelector('#append-button');
+			expect(appendButton).to.exist;
+			setTimeout(() => appendButton.click());
+			const event = await oneEvent(element, 'd2l-rubric-overall-level-added');
+			sinon.assert.calledOnce(sirenActionStub);
```
imcdonald-d2l, line 52: are you able to verify the action passed to `sirenActionStub` using `stub.args` like I did [here](https://github.com/Brightspace/d2l-rubric/blob/master/test/d2l-rubric-editable-score.test.js#L60)?

test/d2l-rubric-overall-level-editor.test.js
MitchelPaulin: Sounds good, I'll look into working around the dialog

MitchelPaulin: Easiest way I could find to work around the dialog is put it in a different function(probably should have been to begin with anyways), this also allows us to test the negative case, where rejecting the dialog doesn't delete anything

imcdonald-d2l: Makes sense to me. Looks good 👌🏻

test/d2l-rubric-overall-level-editor.test.js
```diff
@@ -1,20 +1,74 @@
...
-			<d2l-rubric-overall-level-editor></d2l-rubric-overall-level-editor>
+			<d2l-rubric-overall-level-editor
+				href="/data/instructor/rubrics/organizations/1/2/overalllevels/38.json"
+				token="test"
+			></d2l-rubric-overall-level-editor>
 		`);
 
-		expect(element).to.not.be.undefined;
+		expect(element).to.exist;
+		entity = await getSirenEntity('/data/instructor/rubrics/organizations/1/2/overalllevels/38.json', RubricOverallLevelEntity);
```
imcdonald-d2l, line 23: nit: this only needs to happen once. Can be done in the top level `before` statement see example [here](https://github.com/Brightspace/d2l-rubric/blob/master/test/d2l-rubric-editable-score.test.js#L15).

test/d2l-rubric-overall-levels-editor.test.js
```diff
@@ -1,20 +1,67 @@
...
-			<d2l-rubric-overall-levels-editor></d2l-rubric-overall-levels-editor>
+			<d2l-rubric-overall-levels-editor
+				href="/data/instructor/rubrics/organizations/1/2/overalllevels.json"
+				token="test"
+			></d2l-rubric-overall-levels-editor>
 		`);
 
-		expect(element).to.not.be.undefined;
+		expect(element).to.exist;
+		entity = await getSirenEntity('/data/instructor/rubrics/organizations/1/2/overalllevels.json', RubricOverallLevelsEntity);
```
imcdonald-d2l, line 25: Same as above. Only needs to happen once.

imcdonald-d2l (approved):

PR #1670, opened 2022-02-23
Replaced all uses of not.undefined with to.exist

MitchelPaulin:


test/d2l-rubric-preview-mode.test.js
```diff
@@ -38,17 +38,21 @@ describe('d2l-rubric (preview mode)', () => {
 		let newElement;
 
 		beforeEach(async() => {
-			newElement = await fixture(html`<d2l-rubric></d2l-rubric>`);
+			newElement = await fixture(html`<d2l-rubric
+				href="dummy"
+				token="dummy"
+			></d2l-rubric>`);
```
MitchelPaulin, line 44: Found one test that started failing after doing this, basically if there is no href and token, nothing renders at all, this means the loading icon doesn't render either. 

I'm not sure if this is a bug, do we expect there to be a situation where someone tries to use a rubric without providing an href or token?

test/d2l-rubric-preview-mode.test.js
```diff
@@ -38,17 +38,21 @@ describe('d2l-rubric (preview mode)', () => {
...
 
 		it('do not show content without href', () => {
 			expect(newElement._showContent).to.be.false;
+			expect(!!newElement.hasAlerts).to.be.false;
 		});
 
 		it('check for spinner on page', () => {
 			const subElement = newElement.shadowRoot.querySelector('d2l-rubric-loading');
-			expect(subElement).to.not.be.undefined;
+			expect(subElement).to.exist;
```
MitchelPaulin, line 55: For reference this was the test that started failing

devpow112 (approved):

test/d2l-rubric-preview-mode.test.js
devpow112: I doubt it. I think we can address it if it comes up for some reason.

test/d2l-rubric-adapter.test.js
```diff
@@ -11,15 +11,15 @@ describe('d2l-rubric-adapter', () => {
 			></d2l-rubric-adapter>
 		`);
 
-		expect(element).to.not.be.undefined;
+		expect(element).to.exist;
 		expect(element.localName).to.equal('d2l-rubric-adapter');
 
 		const accordionElement = element.shadowRoot.querySelector('d2l-labs-accordion-collapse');
-		expect(accordionElement).to.not.be.undefined;
+		expect(accordionElement).to.exist;
```
MitchelPaulin, line 18: This PR fixes a lot of issues like this, querySelector returns `null` when it can't find the element, so the test will happily continue over an `undefined` check even if the element could not be found.

PR #1676, opened 2022-02-24
Inline grading tests pt2 - US136165

imcdonald-d2l:


test/d2l-rubric-editable-score.test.js
```diff
@@ -1,65 +1,160 @@
...
+					assessment-href="${CriterionAssessmentHref}"
+					score="${criterionAssessmentEntity.properties.score}"
+					hovered="-1"
+					selected="0"
+					compact
+					token="test"
+				></d2l-rubric-editable-score>
+			`);
+			element.assessmentEntity = criterionAssessmentEntity;
+			element.entity = criterionEntity;
```
imcdonald-d2l, line 134: @MitchelPaulin @devpow112 just an FYI that I have found that any test that relies on a property set with `this.watchEntity(...)` results in a flakey test. It isn't guaranteed that the property will be set when the test executes. Getting around that issue by just setting the property manually.

test/d2l-rubric-editable-score.test.js
```diff
@@ -1,65 +1,160 @@
...
+			await nextFrame();
+			expect(element).to.be.an.instanceof(RubricEditableScore);
+		});
+
+		afterEach(() => {
+			sandbox.restore();
+		});
```
MitchelPaulin, line 141: Would it make more sense to have one `sandbox.restore()` call at the top of the file rather than calling it inside each `describe`?

MitchelPaulin (approved):

test/d2l-rubric-editable-score.test.js
imcdonald-d2l: I thought if it was at the top-level it wouldn't call `restore` in child `describe` tests. I'll verify in my next PR and update if necessary 👍🏻

imcdonald-d2l: Top level `restore` seems to work fine so will be a part of my next PR.

PR #1678, opened 2022-02-24
Feedback and Total Score JAWS fix - DE47039

zhouk3v:
[Rally Ticket](https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fdefect%2F623814566427&fdp=true)

Main Changes
- Added aria-labels to `rubric-feedback.js` and `rubric-total-score.js` so the feedback and total score can be read out by JAWS in firefox

NOTE: The fix for `rubric-feedback.js` is inconsistent, at least from my local testing. We might need to revert that fix if it doesn't end up working during QA as discussed earlier today.


src/components/rubric-feedback.js
```diff
@@ -542,11 +542,14 @@ class RubricFeedback extends composeMixins(
 
 		let nonEditableFeedback = null;
 		if (!this.addingFeedback && !this._canEditFeedback()) {
-			const feedbackHtml = this._pendingFeedbackSaves > 0 ? this._feedback : this.CriterionAssessmentHelper.getFeedbackHtml(this._criterionAssessment);
```
zhouk3v, line 545: I don't think we need to worry about pending saves in the student view where `nonEditableFeedback` is rendered.

devpow112:
For lint issues you should be able to run `npm run format` to fix all the pedantic ones.

src/components/rubric-total-score.js
```diff
@@ -96,17 +96,23 @@ class RubricTotalScore extends composeMixins(
...
-
 		return html`
-			<div class="out-of-container" tabindex="0">
+			<div class="out-of-container" tabindex="0" aria-labelledby="total-grouping-label">
+				<d2l-offscreen id="total-grouping-label" aria-hidden="true">
+					${this.localize('rubricEditorAccessibility:totalScoreOutOf', 
+						'rubricName', this.entity.name, 
+						'score', this.totalScore, 
+						'outOf', this.entity.outOf.toString())}
+				</d2l-offscreen>
```
devpow112, line 106: I think this will take care of the lint issues

```suggestion
				<d2l-offscreen id="total-grouping-label" aria-hidden="true">
					${this.localize(
						'rubricEditorAccessibility:totalScoreOutOf',
						'rubricName', this.entity.name,
						'score', this.totalScore,
						'outOf', this.entity.outOf.toString()
					)}
				</d2l-offscreen>
```

devpow112: NM doesn't take care of it :(

devpow112: I think you'll have to pull the localize call out of the HTML template.

```js
		const totalScoreOutOfText = this.localize(
			'rubricEditorAccessibility:totalScoreOutOf',
			'rubricName', this.entity.name,
			'score', this.totalScore,
			'outOf', this.entity.outOf.toString()
		);
		
		return html`
			<div class="out-of-container" tabindex="0" aria-labelledby="total-grouping-label">
				<d2l-offscreen id="total-grouping-label" aria-hidden="true">${totalScoreOutOfText}</d2l-offscreen>
```

src/components/rubric-total-score.js
```diff
@@ -97,16 +97,22 @@ class RubricTotalScore extends composeMixins(
 			`;
 		}
 
+		const totalScoreOutOf = this.localize('rubricEditorAccessibility:totalScoreOutOf', 'rubricName', this.entity.name, 'score', this.totalScore, 'outOf', this.entity.outOf.toString());
+
 		return html`
-			<div class="out-of-container" tabindex="0">
+			<div class="out-of-container" tabindex="0" aria-labelledby="total-grouping-label">
+				<d2l-offscreen id="total-grouping-label" aria-hidden="true">
```
imcdonald-d2l, line 104: Why are you using `aria-hidden` on a `d2l-offscreen` component? The point of this component is for it to be used in the context of accessibility.

zhouk3v: Leaving out `aria-hidden` causes NVDA on chrome to read out the total score twice. It also makes JAWS on Firefox unable to read out the total score. It might be a better idea to replace the `d2l-offscreen` component with an `aria-label` property to avoid confusion then.

imcdonald-d2l: Yes, makes more sense to have this as a `aria-label` as opposed to the `d2l-offscreen` component.

imcdonald-d2l (approved):

PR #1684, opened 2022-02-25
Switch to offscreen styles over component

devpow112:
Helps speed up rendering as there is less component instantiation.


src/components/rubric-levels-mobile.js
```diff
@@ -386,9 +387,9 @@ class RubricLevelsMobile extends composeMixins(
 					<d2l-icon icon="d2l-tier1:chevron-right"></d2l-icon>
 				</div>
 			</div>
-			<d2l-offscreen class="criterion-status">
+			<div id="criterion-status" class="d2l-offscreen">
```
devpow112, line 390: Switched `criterion-status` from a `class` to `id` as it was only used in one spot.

jesseag (approved):
Looks good to me.

PR #1688, opened 2022-03-01
Add OSLO sanity test

devpow112:
Saw this come-up in a conversation. This just checks that no term names include a `.` which is currently a reserved character when you are using OSLO due to how the transformation works.


imcdonald-d2l (approved):
Good call 👌🏻

PR #1689, opened 2022-03-01
Rubric create test - group(s)

MitchelPaulin:


test/d2l-rubric-criteria-groups-editor.test.js
```diff
@@ -1,76 +1,111 @@
...
+		before(async() => {
 
-			expect(element).to.exist;
-			expect(element.localName).to.equal('d2l-rubric-criteria-groups-editor');
+			multipleGroupElement = await fixture(html`<d2l-rubric-criteria-groups-editor
+				href="../data/instructor/rubrics/organizations/1/8/groups.json"
+				token="test"
+			></d2l-rubric-criteria-groups-editor>`);
+
+			await new Promise((resolve) => { setTimeout(resolve, 1000); });
```
MitchelPaulin, line 51: Open to alternatives to this here, the href I am using has a lot of data and takes a while to load. Unfortunately `fixtureWaitForChangeEvent` doesn't work. I was looking at `fixtureWaitForSpecificChangeEvent` but couldn't think of a helpful event.

test/components/static-data/rubrics/organizations/text-only/199/groups/176.json
```diff
@@ -9,8 +9,18 @@
         {
```
devpow112, line 9: Are you able to switch away from `test/components/static-data` or is there no nice equivalent in `/data`? If it's too much work that's fine. I'd just like to eventually kill all the data we didn't auto generate (anything not under `/data`)

test/d2l-rubric-criterion-editor.test.js
```diff
@@ -3,15 +3,15 @@ import './components/static-data/rubrics/organizations/text-only/199/groups/176/
...
+import sinon from 'sinon';
 
 const eventOptions = { bubbles: true, cancelable: false, composed: true };
 
 describe('d2l-rubric-criterion-editor', () => {
 	let sandbox;
 
 	before(() => {
-		sandbox = createSandbox();
+		sandbox = sinon.createSandbox();
```
devpow112, line 14: Why the change to use `sinon.createSandbox`? Just curious if there was a technical reason.

MitchelPaulin: I think thats left over from me trying to debug some failing tests, I can switch it back

test/d2l-rubric-criteria-group-editor-create.test.js
```diff
@@ -10,12 +10,16 @@ describe('d2l-rubric-criteria-group-editors', () => {
 	const performActionStub = sandbox.stub(RubricCriteriaGroupEditor.prototype, 'performSirenAction').callsFake(() => Promise.resolve());
 
 	before(async() => {
-		element = await fixtureWaitForChangeEvent(
-			html`<d2l-rubric-criteria-group-editor show-group-name></d2l-rubric-criteria-group-editor>`,
-			'static-data/rubrics/organizations/text-only/199/groups/176.json'
+		element = await fixture(
+			html`<d2l-rubric-criteria-group-editor
+				href="../data/instructor/rubrics/organizations/1/2/groups/18.json"
```
devpow112, line 15: I thin base on how the test runner is configured this can just be `/data/instructor/` instead of `../data/instructor/`. I may be wrong though.

imcdonald-d2l: Yeah, should just be `/` not `../`

test/utilities.js
```diff
@@ -77,3 +77,19 @@ export const isVisible = (element) => {
...
+export const isRejected = async(promise) => {
+	try {
+		await promise;
+	} catch (_) {
+		return true;
+	}
+	return false;
+};
+
+export const waitOneSecond = async() => {
```
MitchelPaulin, line 93: Added this as a hack until we can figure out something better

imcdonald-d2l, line 93: Could we make this a utility that takes in a `ms` parameter so we can control the timeout? Would be more useful in the future if we need to tweak the duration.

imcdonald-d2l: When I was playing around with accessibility tests I did something very similar. Don't know of any more graceful solutions to the problem this is trying to solve at the moment.

test/d2l-rubric-criterion-editor.test.js
```diff
@@ -233,6 +233,11 @@ describe('d2l-rubric-criterion-editor', () => {
 				const subElement = element.shadowRoot.querySelector('#remove');
 				expect(subElement).to.be.null;
 			});
+
+			it('dropdown is hidden', () => {
+				const subElement = element.shadowRoot.querySelector('d2l-dropdown-more');
+				expect(subElement).to.be.null;
```
imcdonald-d2l, line 239: Would `to.not.exist` be better than `to.be.null`?

MitchelPaulin: querySelector doesn't return undefined, so checking for null directly here is fine

devpow112: Interesting. Good to know. Wonder if we are over checking anywhere (checking `null` and `undefined`) in our code when we have a return from `querySelector`

devpow112 (approved):
Looks good to me :tada:

PR #1701, opened 2022-03-03
Assessed rubric readonly/preview vdiff tests - US136168

imcdonald-d2l:
Also adds the published warning on editable assessed rubrics. It wasn't there before since it's set with an explicit attribute.


MitchelPaulin:
image

Is this what a list should look like in rtl?

imcdonald-d2l:
> Is this what a list should look like in rtl?

I assume that is incorrect. But that would be an issue with the `d2l-html-block` not an issue on our end.

imcdonald-d2l:
@MitchelPaulin looks like issue will be fixed by Grant. We can merge and ignore this for now. Vdiffs will be updated whenever we get the update from `html-block`. See slack thread [here](https://d2l.slack.com/archives/C0PHG3QB0/p1646332621485569).

test/perceptual/compact-assessed.visual-diff.js
```diff
@@ -29,25 +29,46 @@ describe('compact-assessed', function() {
...
+					const properties = [{ name: 'showPublishedWarning', value: true }];
+					await setProperties(page, selector, properties);
 					await callAssessmentEntityChanged(page, selector);
 
 					const rect = await visualDiff.getRect(page, id);
 
 					await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
 				});
+
+				if (dir === 'ltr') {
```
MitchelPaulin, line 53: Any reason to not have an rtl version of this?

imcdonald-d2l: We discussed in stand-up that we didn't need `rtl` tests for this. I can remove the condition we want. Just going off what was discussed.

MitchelPaulin: I interpreted the conversation this morning as not needing rtl versions of unit tests, not vidff tests, but I could be wrong. In this case though since its visual I think its worth it

imcdonald-d2l: Yeah, I believe there is some value in `rtl` vdiff as well. @devpow112 do you want to add your thoughts before I make any changes?

MitchelPaulin (dismissed):

imcdonald-d2l:
Rebase messed this branch up. Just going to close and re-open a new PR.

imcdonald-d2l:
New PR: https://github.com/Brightspace/d2l-rubric/pull/1712

PR #1706, opened 2022-03-03
Criterion name value missing condition - DE47366

imcdonald-d2l:
On test sites requests are coming back really quickly causing the name to be set by the back-end returned value when the name is still changing. 

Can't reproduce locally since requests are not as quick.

See video attached on defect: https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fdefect%2F627364235353


devpow112 (approved):

src/components/editor/rubric-criterion-editor.js
```diff
@@ -483,8 +483,8 @@ class RubricCriterionEditor extends composeMixins(
 		this._loaMappingHref = entity.loaMappingHref || null;
 	}
 
-	_getDisplayedName(isFocused, isInvalid, pendingSaves, enteredValue, savedValue) {
-		return (isFocused || isInvalid || pendingSaves > 0) ? enteredValue : savedValue;
+	_getDisplayedName() {
+		return (this._nameChanging || this._nameFocused || this._nameError || this._nameSaves > 0) ? this._enteredName : this._name;
```
imcdonald-d2l, line 487: Condition was missing `this._nameChanging` which captures the time in between changing the name and the request being sent by the debouncer. Not 100% sure this will fix the issue since it only happens on tests sites with fast responses - doesn't break existing behavior though.

PR #1715, opened 2022-03-04
Fix `EventBody.EventTypeGuid` telemetry schema issue - DE46661

imcdonald-d2l:
Should fix the following issue:
```
[{"keyword":"enum","dataPath":".EventBody.EventTypeGuid","schemaPath":"#/allOf/1/properties/EventBody/properties/EventTypeGuid/enum","params":{"allowedValues":["D6B021B6-FFA2-4EE8-8206-8710E04396CE","d6b021b6-ffa2-4ee8-8206-8710e04396ce"]},"message":"should be equal to one of the allowed values"}]
```

Still looking into the `EventBody.Context.Id` schema issue.


src/mixins/telemetry-mixin.js
```diff
@@ -217,6 +217,8 @@ const TelemetryMixinInternal = (superclass) => class extends superclass {
 			}
 		}
 
+		this.eventType = isPerformanceLog ? 'PerformanceEvent' : 'TelemetryEvent';
```
mpharoah-d2l, line 220: This is just setting a property on the telemetry service, not the event, so it doesn't do anything

mpharoah-d2l: Oh nevermind, it is used. It would be better to pass it down to `_logEvent` though instead of setting it on `this`

imcdonald-d2l: Yeah, I can change it so it is passed through `_logEvent`

mpharoah-d2l (approved):

PR #1723, opened 2022-03-08
Remove compact-overall-score references - US133770

imcdonald-d2l:
Flag is no longer needed, see slack discussion [here](https://learningoutcomes.slack.com/archives/CANSREZKJ/p1646748859529989). Removing it in the rubrics UI before other references are removed. 

Part of portfolio work, currently the FRA isn't showing overall score since the property is not even being set on `d2l-rubric` [here](https://github.com/Brightspace/folio-app/blob/cf30846d4490b57b9878129969001dd42c330750/src/components/rubrics/rubrics.js#L283). We could toggle it with the flag if we wanted but have decided to remove the flag completely. 

https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F615861431625


devpow112:
Do you feel we need to also remove this from polymer or not worth it?

imcdonald-d2l:
> @imcdonald-d2l do you feel we need to also remove this from polymer or not worth it? I'm leaning towards not worth it. I'm assuming there will also be LMS work we can't do until it's fully removed (i.e. polymer is gone)

I'm fine removing it from polymer - it would be minimal work. That we we can remove it from consistent eval and back-end now and not worry about it later.

devpow112 (dismissed):

devpow112 (approved):

PR #1725, opened 2022-03-08
`3.x` - set compact overall score  to `true` - US133770

imcdonald-d2l:
We are removing feature flag so this property no longer matters. Setting default to `true` so we can remove references in consistent-eval and back-end. 

Related: https://github.com/Brightspace/d2l-rubric/pull/1723

https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F615861431625


devpow112 (dismissed):

imcdonald-d2l:
@devpow112 any concerns about force-merging this? Polymer doesn't like core components referencing `classProperties`

devpow112:
Looks like tests will fail for the same reason 😞. I may just remove both.

devpow112 (approved):

PR #1756, opened 2022-03-17
DE47790 - Add spacebar as keyboard shortcut

zhouk3v:
[Rally Ticket](https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fdefect%2F629703439597)

Changes:

Now checking for the spacebar key code in `_handleKey()`. Renamed `_handleEnter()` to `_toggleSelection()` to reflect that the spacebar can now select/deslect criterion cells.


src/components/rubric-levels-mobile.js
```diff
@@ -755,7 +745,7 @@ export class RubricLevelsMobile extends composeMixins(
 	}
 
 	_handleKey(event) {
-		if (event.keyCode === KEY_CODES.LEFT_ARROW || event.keyCode === KEY_CODES.RIGHT_ARROW || event.keyCode === KEY_CODES.ENTER) {
+		if (event.keyCode === KEY_CODES.LEFT_ARROW || event.keyCode === KEY_CODES.RIGHT_ARROW || event.keyCode === KEY_CODES.ENTER || event.keyCode === KEY_CODES.SPACE) {
```
MitchelPaulin, line 748: This if statement is getting a bit long, maybe switching to a set and using `set.has` would be better

zhouk3v: Should I add the set in the properties of `RubricLevelsMobile` or just declare it in `_handleKey()`?

src/components/rubric-levels-mobile.js
```diff
@@ -764,7 +754,13 @@ export class RubricLevelsMobile extends composeMixins(
...
 				return this._handleRight(event);
 			case KEY_CODES.ENTER:
-				this._handleEnter();
+				this._toggleSelection();
+
+				// announce level name and cell status (selected/unselected)
+				announce(this._getLevelNameStatus(event));
+				return;
+			case KEY_CODES.SPACE:
+				this._toggleSelection();
```
MitchelPaulin, line 763: Looks like the ENTER case and the SPACE case both do the same thing, you could combine them into one statement 

```
case KEY_CODES.ENTER:
case KEY_CODES.SPACE;
    this._toggleSelection();
    announce(this._getLevelNameStatus(event));
```

src/components/rubric-levels-mobile.js
MitchelPaulin: Up to you, declaring it as a property is probably better since the object will only get created once rather than every call to `_handleKey()`

MitchelPaulin (approved):

PR #1761, opened 2022-03-22
DE47831 - Fix height issue

devpow112:
This also get's rid of a weird flicker in the hover styles that was occurring due to a brief delay between the parent rendering and the edit element rendering.

**Before**
before

**After**
after


src/components/rubric-editable-score.js
```diff
@@ -121,8 +123,12 @@ export class RubricEditableScore extends composeMixins(
 					background-color: var(--d2l-color-celestine-plus-2);
 				}
 
-				:host(.assessable:not([editor-styling])) #input-container:focus,
-				:host(.assessable:not([editor-styling])) #input-container:hover {
+				.editing {
+					border-color: transparent !important;
+				}
+
+				:host(.assessable:not([editor-styling])) :not(.editing) #input-container:focus,
```
devpow112, line 130: This is starting to get a bit ridiculous. Might look at optimizing this in the future.

devpow112:
VDiff tests should fail as this was broken in the vdiff tests for assessed rubrics.

src/components/rubric-editable-score.js
```diff
@@ -738,18 +750,35 @@ export class RubricEditableScore extends composeMixins(
...
-				${this._scoreOverridden && !this.readOnly ? html`
-					<div class="d2l-offscreen">${this._localizeStarLabel()}</div>
-				` : null}
+				${this._renderScoreOverriddenOffscreen()}
+			</div>
+		`;
+	}
+
+	_renderScoreOverriddenOffscreen() {
+		if (!this._scoreOverridden || this.readOnly) {
```
devpow112, line 763: I think this is the opposite of `this._scoreOverridden && !this.readOnly` logically. We used to use `this._scoreOverridden && !this.readOnly` to determine if we should render the override offscreen. Now we early exist and don't render.

devpow112:
VDiff look ok to me. There is some resizing occurring because before the height of the edit score component could vary depending on if the text was unaligned. Now it's consistent everywhere.

mpharoah-d2l (approved):

devpow112:
Unit tests are gonna fail. Installing dependencies got stuck for some reason. Maybe a good time to look at doing #1762.

MitchelPaulin (approved):

devpow112:
Safari is being flakier then usual.

PR #1770, opened 2022-03-23
Kzhou/de45418 arrow keys fix

zhouk3v:
[Rally Ticket](https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fdefect%2F610336005919)

Cause: `_queueRadio()` was still being called when the rubric was in read-only mode,  which caused the criterion cell to appear selected

Changes:
- Add guard clause that checks `_isStaticView()` in `_queueRadio()`
- Move `_isStaticView()` check in `__selectRadio()` and `_deselectRadio()` to still allow users to navigate between cells in the same row without triggering styling changes.


MitchelPaulin (approved):

jesseag (approved):

zhouk3v:
Reopening PR to try and retrigger checks

PR #1774, opened 2022-03-24
Switch to `aTimeout`

devpow112:
I really want to get rid of timeouts eventually but for now it's better to use a framework provided function vs our own.


tools/is-lit-flag-enabled.js
```diff
@@ -20,8 +18,6 @@ const isLitFlagEnabledInternal = () => {
 		}
 	}
 
-	console.log('Lit Enabled');
```
devpow112, line 23: Removing as I'm confident this works and it's cluttering test output.

mpharoah-d2l (approved):

PR #1775, opened 2022-03-24
DE45861 - Fix feedback not saved when pop-out window is closed via wi…

jesseag:
…ndow close button.

Rally ticket: https://rally1.rallydev.com/#/57732444928d/dashboard?detail=%2Fdefect%2F613273630567

Bug: Criterion feedback was not being saved to the mobile component when closing out of the assessed pop-out view via the window close button.

Fix:
Added in auto-save="true" as a an attribute from the d2l-htmleditor. This creates a similar flow to selecting the close button when you have unsaved criterion feedback. It triggers a notification asking if you want to "leave the page". To save the feedback, hooked into the unload event ('beforeunload') and called saveFeedback() with the text currently inside the html editor.
https://github.com/BrightspaceUI/htmleditor

Bug:
Feedback_Bug

Fix:
Feedback_Fix


src/components/rubric-feedback.js
```diff
@@ -651,6 +654,12 @@ export class RubricFeedback extends composeMixins(
 		}
 	}
 
+	async _saveFeedbackOnWindowClose(e) {
+		// prevent firing two window close events
+		e.stopPropagation();
+		await this._saveFeedback(this._getEditor().html);
```
devpow112, line 660: What happens in the save takes a while? Does it just wait or is there a limit before it just kills it?

jesseag: I'll do some testing on this to see what I can find.

src/components/rubric-feedback.js
```diff
@@ -651,6 +654,12 @@ export class RubricFeedback extends composeMixins(
 		}
 	}
 
+	async _saveFeedbackOnWindowClose(e) {
+		// prevent firing two window close events
+		e.stopPropagation();
```
jesseag, line 659: Any thoughts on whether e.stopPropagation() is actually needed? I've commented this out and it really doesn't make any difference.

There is one other spot where we use the 'beforeunload' event.
https://github.com/Brightspace/d2l-rubric/blob/master/src/components/rubric.js#L272

MitchelPaulin: I don't think this is necessary, this comment here is also incorrect, stopping the propagation of the event won't prevent it from firing again, it would just prevent any other event listener from seeing the event, right?

src/components/rubric-feedback.js
```diff
@@ -495,6 +497,7 @@ export class RubricFeedback extends composeMixins(
 				class="rubric-cell-editor"
 				type="inline"
 				html=${this._feedback}
+				auto-save="true"
```
MitchelPaulin, line 500: What does this do?

jesseag: https://github.com/BrightspaceUI/htmleditor#properties

image

This setting needs to be set to "true" as this is what will allow the "Leave site" drop-down alert to appear when there is unsaved content in the html-editor".

jesseag: I don't know if it is absolutely necessary to have but it is nice to have it here because now the selecting the "Close" button in the pop-out rubric and selecting the Window.Close button both give you the same prompt if there is unsaved content now. So it is nice to have the flows consistent now for both close actions when there is unsaved content in the editor.

MitchelPaulin: Make sense, you also don't need to set it to "true". Just having it as an attribute works

src/components/rubric-feedback.js
jesseag: Removing this function in my next commit. I will also be await on the async function in the eventListener. And yes, the comment is wrong here for sure.

src/components/rubric-feedback.js
jesseag: So I did a couple of things here to try to test this:

1) Set network speed to be dial-up (56 kbps) in the google chrome network tab. The window still closes even with the slow network speed. 

2) Tried adding in a timeout here to slow down the execution just to see what would happen in this case. This works when the pop-out window remains open and the feedback is saved by clicking elsewhere in the pop-out window. You can monitor this from the network tab and see that the requests to the back-end are delayed. However, when you try to run the same code when you close the window the browser refuses to run it as part of the 'beforeunload' event. I was talking with Mitch about it and he was mentioning that this makes likely sense that the browser won't allow this to happen. If the browser allowed this kind of behaviour then a malicious actor could stop your browser from being closed with an infinite timeout or they could do other malicious types of behaviour. 

Either way, when the Window.close action happens the pop-out window will be closed and the await call here is not going to be a blocking call. So I think that is the main thing we wanted to make sure of here.

`async _saveFeedback(value) {
		this._feedbackModified = false;
		this.feedbackInvalid = false;
		this._pendingFeedbackSaves++;
		this.dispatchEvent(new CustomEvent('save-feedback-start', {
			detail: {
				hasPendingSaves: true,
			}
		}));

		try {
			// NEW TIMEOUT CODE
			const sleep = ms => {return new Promise(resolve => setTimeout(resolve, ms));};
			console.log('function invoked...');
			await sleep(5000);
			console.log('I got here about 5000 milliseconds later');
			//

			await this.CriterionAssessmentHelper.updateFeedbackAsync(
				() => this._criterionAssessment,
				value,
				this.context?.isNewHtmlEditorEnabled
			);
			// update assessment entity to get latest actions in case feedback is overridden
			this._updateCriterionAssessment();`

MitchelPaulin (approved):

PR #1777, opened 2022-03-25
Add mocking of POST requests to vdiff

devpow112:
This helps eliminate some noise in the network requests and component updates. Hopefully be helpful with eliminating timeouts at some point.


test/perceptual/helpers/vdiff.js
```diff
@@ -17,12 +31,12 @@ export const waitForEntityReady = async(page, selector) => {
 							() => resolve()
 						)
 					),
-					2500
+					3500
```
devpow112, line 34: Mocking the POST's actually speeds up the loading of rubrics in vdiff. This causes issues with the save status text still being visible sometimes. Looking at solution for this based on events but so far no luck.

PR #1780, opened 2022-03-25
US136755 - Remove legacy HTML editor (Lit)

devpow112:
Removes the uses of `use-new-brightspace-editor` attribute as well as the legacy HTML editor wrapper.

LMS change: https://github.com/Brightspace/lms/pull/21086
Rubric change (Polymer): https://github.com/Brightspace/d2l-rubric/pull/1782
Consistent Evaluation change: https://github.com/BrightspaceHypermediaComponents/consistent-evaluation/pull/907
Activities change: https://github.com/BrightspaceHypermediaComponents/activities/pull/2498
Feature Flag change: https://github.com/Brightspace/lms-launch-darkly-flags/pull/4107


src/components/rubric.js
```diff
@@ -304,7 +303,7 @@ export class Rubric extends composeMixins(
 		this._context = {
 			...this._context,
 			isFeedbackCopyEnabled: this._enableFeedbackCopy,
-			isNewHtmlEditorEnabled: this.useNewBrightspaceEditor,
+			isRichTextEnabled: true, // not properly supported yet
```
devpow112, line 306: Right now both `d2l-rubric` and `d2l-consistent-evaluation` don't properly  support the disabling of the HTML editor using the `d2l.Preferences.TurnOffWYSIWYG` config. Hard-coding to `true` to allow downstream changes to work as expected.

devpow112:
VDiff changes were just updating from the default input area to use the actual new HTML editor.

src/components/rubric-feedback.js
```diff
@@ -646,7 +638,7 @@ export class RubricFeedback extends composeMixins(
 
 	async _saveFeedbackHandler(event) {
 		if (this._feedbackModified || !this._pendingFeedbackSaves && this.feedbackInvalid) {
-			const value = this.useNewBrightspaceEditor ? this._getEditor().html : event.target.value;
+			const value = event.target.value;
```
mpharoah-d2l, line 641: Is this supposed to be `this._getEditor().html` or was the condition before unnecesary?

devpow112: The `this.useNewBrightspaceEditor` was always `undefined` before. The call should have been `this._context?.useNewBrightspaceEditor`. Removing it entirely seems to work as desired. I'll have to do some more testing after everything is all merged together.

devpow112: Looks like there will be some work around removing the ability to switch to simple text input as well (`rich-text-enabled`/`isRichTextEnabled` is going away as the config it's set by `d2l.Preferences.TurnOffWYSIWYG` is going away). I'll double check this when doing that work.

mpharoah-d2l (approved):

PR #1791, opened 2022-03-28
US137469 - Remove `rich-text-enabled` attribute (Lit)

devpow112:
Remove the `rich-text-attribute` and all uses of it. The config that controlled this no longer exists. We should always show rich text HTML editors. This sadly skips the editor vdiffs for now as there is some rendering issue with the HTML that only happens in vdiff. Will create story to look into later.

Rubrics change (Polymer): https://github.com/Brightspace/d2l-rubric/pull/1821
Activities change: https://github.com/BrightspaceHypermediaComponents/activities/pull/2533
LMS change: https://github.com/Brightspace/lms/pull/21244

https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F630472779497


wds.config.js
```diff
@@ -15,17 +15,49 @@ export const devServerConfig = {
 					}
 				}
 			}
-		})
+		}),
+		{
+			name: 'setup-html-editor-context',
```
devpow112, line 20: Adds needed HTML editor context to all demo and test pages. This does not include vdiff sadly.

demo/rubric-editor.html
```diff
@@ -31,17 +31,6 @@ <h2>Default</h2>
 			</template>
 		</d2l-demo-snippet>
 
-		<h2>Default (Rich Text Enabled)</h2>
```
devpow112, line 34: Not needed since it was just showing the rich text enabled state.

src/components/rubric-feedback.js
```diff
@@ -480,50 +473,20 @@ export class RubricFeedback extends composeMixins(
 		}
 
 		let newEditor = null;
```
MitchelPaulin, line 475:
```suggestion
		let editor = null;
```

devpow112: I'll have to also change the below code
```js
		newEditor = html`<d2l-htmleditor
			id="text-area"
```
I can do that

devpow112: Ended up just moving it down to where that variable was used.

MitchelPaulin (approved):

jesseag (approved):

PR #1798, opened 2022-03-31
Mock vdiff POST and PUT requests

devpow112:
Removes a bunch of errors from the output and speeds up loading a bit.


devpow112:
The vdiff changes are only with the editor. Now that we don't fail on POST/PUT calls the `Saving...` text appears correctly and consistently in diffs.

PR #1809, opened 2022-04-04
Fix vdiff

devpow112:
Add a try/catch around trying to construct assessment URL. Since URL's are relative in vdiff and other testing setups this will fail since the `URL` constructor expects either an absolute URT or a relative URL with a `base` that is absolute.


src/components/rubric.js
```diff
@@ -403,9 +403,18 @@ export class Rubric extends composeMixins(
...
+		}
+
+		try {
+			const paramString = this.assessmentHref.split('?')[1];
+			const params = new URLSearchParams(paramString);
+
+			return !!params.get('workingCopyId');
+		} catch (_) {
+			return false;
+		}
```
devpow112, line 417: Ignore any error and just return `false`. This will also work with relative URL's which are used in testing. Tested manually using demo pages.

src/components/rubric.js
```diff
@@ -403,9 +403,18 @@ export class Rubric extends composeMixins(
...
 	_checkHasWorkingCopy() {
-		if (!this.assessmentHref) return false;
-		const assessmentUrl = new URL(this.assessmentHref);
-		return !!assessmentUrl.searchParams.get('workingCopyId');
+		if (!this.assessmentHref) {
+			return false;
+		}
+
+		try {
+			const paramString = this.assessmentHref.split('?')[1];
```
devpow112, line 411: Any errors from this line will just result in a `return false`.

mpharoah-d2l (approved):

PR #1835, opened 2022-04-08
Fix print issues - DE47996

imcdonald-d2l:
Adjust print logic so rubric components are visible on print. 
https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fdefect%2F630824741381

### Student assignment preview - full page
student-assessed-full
### Student assignment preview - mobile page
student-assessed-mobile
### Rubric pop out - full & mobile
pop-out-print-view


src/components/rubric.js
```diff
@@ -371,7 +373,7 @@ export class Rubric extends composeMixins(
 	}
 
 	isCompact() {
-		return this.forceCompact || this._isMobile;
+		return this._isPrinting ? false : (this.forceCompact || this._isMobile);
```
imcdonald-d2l, line 376: This forces rubrics into full-page mode when printing.

src/components/rubric.js
```diff
@@ -562,19 +564,17 @@ export class Rubric extends composeMixins(
...
 				* {
 					visibility: hidden;
-					height: 0;
 					width: 0;
-					margin: 0;
-					padding: 0;
-					border: none;
-					overflow: hidden !important;
+					height: 0;
+					overflow: visible !important;
```
imcdonald-d2l, line 577: This was necessary so that the whole rubric would print in the pop-out window. The dialog was doing weird things where it would cut off anything past one page without this.

mpharoah-d2l: Might be better to just add this as a new style element so it doesn't overwrite whatever was in the first style tag

imcdonald-d2l: We revert these styles on [line 594](https://github.com/Brightspace/d2l-rubric/pull/1835/files#diff-4e97158339536b32e2ea6e322008d8870bff80d08b27fbff472afe6adc2a6f0eR594) so I don't think that should be necessary

mpharoah-d2l: It could break print styling if something important was in there though

imcdonald-d2l: I'm a bit confused, what do you mean when you say "in there"?

MitchelPaulin: Its `+=` in the case so I don't think anything will get overridden

src/components/rubric.js
MitchelPaulin: Minor thing but you could just write this as 

```
!this.isPrinting && (this.forceCompact || this._isMobile)
```

src/components/rubric.js
MitchelPaulin: But to Matt's point I think I agree that it would be better to add a new style element rather than appending to an existing one.
```js
// add print
document.getElementsByTagName('style').unshift(print_style);
// remove print style
document.getElementsByTagName('style').shift();
```

Something like above might be better, and just getting rid of keeping the temp style around. I think thats what @mpharoah-d2l meant

imcdonald-d2l: I'll try `push()` `pop()` first to see if that works 👍🏻

mpharoah-d2l: Oh, I misread it as `=` instead of `+=`. That works

mpharoah-d2l (approved):

src/components/rubric.js
imcdonald-d2l: I leave it as is then - playing with generating a style element and appending it is producing issues. Didn't try all the approaches but we can adjust later if necessary.

PR #1843, opened 2022-04-11
DE45861 Feedback fix for when window closes. Changing to display wind…

jesseag:
…ow to warn user that feedback may be lost via Window.close.

Rally ticket: https://rally1.rallydev.com/#/57732444928d/dashboard?detail=%2Fdefect%2F613273630567

Updates to ticket:
- Changed ticket after stand up discussion to warn user when they are trying to close the window and there is unsaved feedback.

Notes:
- Created `_feedbackEditorInFocus` because `_feedbackInFocus` tracks the state of whether the editing toolbar in the feedback window is in focus. In addition, this variable gets set to true at times when it should not be set to true.
- `@focusin` event does not get fired every time you focus on a feedback box. It only focuses the first time you focus on a feedback window. As a result, I added in the feedback event `@d2l-htmleditor-focus` to fire every time the user focuses in on an feedback window.
- `this._getEditor().blur();` on Line 376 does not work on Firefox due to the fact that Firefox fires events in different orders however this line works on Chrome. Not sure, if I should remove this line or not so we have consistent behaviour between browsers. The idea here was that if the user selected Window.Close and then selects cancel on the drop down alert the editor immediately blurs which starts the Save process automatically. The only real negative of this on Firefox is that if the user decides to edit text again after selecting the cancel button and hits Window.Close we will immediately close the window and will not display the Unsaved changes alert again. The other small concern would be that the state of `_feedbackEditorInFocus` will be incorrect temporarily. However, as soon as you select/blur away from the textbox then the state will be restored back to the correct state. 
- The if statement(s) in the `_confirmUnsavedChangesBeforeUnload(e)` function are there to ensure that we only display the Unsaved Changes alert when the user selects the Window.Close button. If we do not have the if statements there then when the user selects the Save & Close button then 2 drop down alerts prompting the user to save their Unsaved changes will be shown in succession which would be undesirable behaviour. 
- A separate PR will be created in the consistent-eval repo to update the "Close" button to the new term ("Save & Close") but will confirm with Rebecca L first. Once this is confirmed, I'll link the PR here.


jesseag:
Chrome

Feedback_Fix_2_Chrome

Firefox
Feedback_Fix_2_Firefox

src/components/rubric-feedback.js
```diff
@@ -363,14 +368,36 @@ export class RubricFeedback extends composeMixins(
 		return this.localize('rubricEvaluationAccessibility:feedbackOn', 'criterionName', name);
 	}
 
+	_confirmUnsavedChangesBeforeUnload(e) {
+		// Editor is not read-only and feedback is in focus
+		if (this._getEditor() !== null && this._feedbackEditorInFocus) {
+			// Feedback has been modified so display the messsage
+			if (this._feedback !== this._getEditor().html) {
+				this._getEditor().blur();
```
jesseag, line 376: This is the line that has different behaviours between Chrome/Firefox. Refer to PR's initial comment for a more detailed description.

src/components/rubric-feedback.js
```diff
@@ -363,14 +368,36 @@ export class RubricFeedback extends composeMixins(
 		return this.localize('rubricEvaluationAccessibility:feedbackOn', 'criterionName', name);
 	}
 
+	_confirmUnsavedChangesBeforeUnload(e) {
+		// Editor is not read-only and feedback is in focus
+		if (this._getEditor() !== null && this._feedbackEditorInFocus) {
```
devpow112, line 373: Can you just do 1 call to `this._getEditor` and store the element it returns for user in each of these calls (`const editor = this._getEditor()` and then `editor !== null` and `this._feedback !== editor.html`, etc.). Calling it multiple times just means it has to do a query multiple times instead if once.

src/components/rubric-feedback.js
```diff
@@ -363,14 +368,36 @@ export class RubricFeedback extends composeMixins(
...
+
 	_feedbackChanged(event) {
 		this._feedback = event.detail.value;
 	}
 
+	_feedbackEditorHandler() {
+		this._feedbackEditorInFocus = true;
+	}
+
 	_focus() {
```
devpow112, line 396: Not directly related but why do we have an empty `_focus()` function? If no one calls it we should remove it.

src/components/rubric-feedback.js
```diff
@@ -363,14 +368,36 @@ export class RubricFeedback extends composeMixins(
...
+		// Editor is not read-only and feedback is in focus
+		if (this._getEditor() !== null && this._feedbackEditorInFocus) {
+			// Feedback has been modified so display the messsage
+			if (this._feedback !== this._getEditor().html) {
+				this._getEditor().blur();
+				this._feedbackEditorInFocus = false;
+
+				// Chrome requires returnValue to be set.
+				e.preventDefault();
+				e.returnValue = 'Rubrics';
```
devpow112, line 381: Can this just be an empty string? Just curious

jesseag: Yes, it can be. The default value is an empty string. I had it set to 'Rubrics' when I was doing some testing and seeing if I could differentiate between the CE alert and our alert based on the return value. I'll remove it.

src/components/rubric-feedback.js
jesseag: I'll remove it.

src/components/rubric-feedback.js
jesseag: Yes, I can make this change. It definitely is better to do it this way as Grant mentioned multiple calls to the editor can cause performance issues.

imcdonald-d2l (approved):

devpow112 (approved):

jesseag:
Are we fine to merge this without the consistent-eval changes? I am still waiting to hear back from Rebecca. The consistent-eval changes would just be renaming a function/variables and updating the lang term to whatever we decide Ex: "Save & Close". There would be no logic change in the consistent eval repo. It sounds like at this point that Rebecca would prefer not to change the name of the button.

PR #1853, opened 2022-04-14
Don't show total score when null - DE48027

imcdonald-d2l:
https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fdefect%2F631011112327


devpow112 (approved):

zhouk3v (approved):

src/components/editor/rubric-criteria-groups-editor.js
```diff
@@ -333,7 +333,7 @@ export class RubricCriteriaGroupsEditor extends composeMixins(
 		}
 
 		let totalScoreText = null;
-		if (this.totalScore !== undefined) {
+		if (this.totalScore !== null && this.totalScore !== undefined) {
```
imcdonald-d2l, line 336: This adjusts total score so it doesn't show in `Analytic No Score` AND `Holistic` rubrics. @MitchelPaulin I know you did some `null` changes, but I don't think this is one of those concerns when we want to show things when `totalScore = null`

PR #1860, opened 2022-04-19
Suppress some TinyMCE related logs in unit tests

devpow112:


imcdonald-d2l (approved):

devpow112:
Abandoning I think this is causing HTML editor stuff to explode since it's doing a full initialization. Not sure. Will look into later.

PR #1864, opened 2022-04-20
Add exists helper

devpow112:
We do a lot of `thing === null || thing === undefined` type of statements since our linter rules don't allow `==`. Add helper to just check if exists. Might push into `core` if we use it elsewhere but for now we have one.


MitchelPaulin:
Everywhere we had `something !== null || something !== undefined` would return true no matter the value of `something`, whoops

devpow112:
> Everywhere we had `something !== null || something !== undefined` would return true no matter the value of `something`, whoops.
> 
> Just wanted to check that everything still works as expected with `something !== null && something !== undefined`

This seems fine. I think 3 of the 4 locations were around localization type of stuff that would have just thrown an error if it was incorrect before. Yeah I noticed that as well that's why I decided to do this.

PR #1866, opened 2022-04-20
Clean-up some existance checks

devpow112:
Simplify existence checks. Now doing `value == null` is allowed by the linter (https://github.com/Brightspace/eslint-config-brightspace/pull/49).


src/components/rubric-editable-score.js
```diff
@@ -627,7 +627,7 @@ export class RubricEditableScore extends composeMixins(
 			return null;
 		}
 
-		if (score !== null || score !== undefined) {
```
devpow112, line 630: This check was also just wrong. It would always be `true`. We only want this to be `true` and allow the use of `score` when it exists. Corrected this check.

src/components/rubric-levels-mobile.js
```diff
@@ -843,7 +843,7 @@ export class RubricLevelsMobile extends composeMixins(
 			const hoveredLevel = this.assessmentMap.get(this._getSelfLink(this.criterionCells[index]));
 			const score = hoveredLevel?.properties?.score;
 
-			if (score !== null || score !== undefined) {
```
devpow112, line 846: This check was also just wrong. It would always be `true`. We only want this to be `true` and allow the use of `score` when it exists. Corrected this check.

devpow112:
There are other `=== null` or `=== undefined` we can probably expand/clean-up but I left all of those as is since it would be much more testing to make sure we didn't actually want to explicitly check for `null` or `undefined`.

devpow112:
I'll update our other repos with the new lint config as well so this is allowed in all our stuff.

src/components/rubric-adapter.js
```diff
@@ -222,7 +222,7 @@ class RubricAdapter extends composeMixins(
 		// A holistic rubric only has one criteria
 		for (const cell of assessmentCriterionCells[0]) {
 			if (cell.hasClass('selected')) {
-				if (cell.properties.score !== null && cell.properties.score !== undefined) {
+				if (cell.properties.score != null) {
```
MitchelPaulin, line 225: This might cause problems if the score is 0

devpow112:
I double checked and at least using jsfiddle `0` and `false` are not coerced for `== null`. According to jsfiddle running `console.log(0 == null)` is `false` and `console.log(0 != null)` is `true`. This is also true for `false` (`console.log(false == null)` is `false` and `console.log(false != null)` is `true`). The only allowed check is `undefined == null` which is `true`.

MitchelPaulin (approved):

mpharoah-d2l (approved):

PR #1868, opened 2022-04-21
Added new vdiff data with assessor info

MitchelPaulin:


MitchelPaulin:
Unfortunately a lot of our tests rely on the hard coded file path ids so the drag and drop approach doesn't work, going to take a more surgical approach and just update the relevant bits

PR #1892, opened 2022-04-27
Hid the rubric archived status in readonly(student) view

MitchelPaulin:


src/components/rubric.js
```diff
@@ -691,7 +691,7 @@ export class Rubric extends composeMixins(
 		}
 
 		let archivedStatus = null;
-		if (this._isArchived) {
+		if (this._isArchived && !this.readOnly) {
```
imcdonald-d2l, line 694: Do we need a condition like this in mobile view as well? If the user re-sizes their window they could get into a read-only mobile view.

PR #1900, opened 2022-05-02
Hid the archived tag from students

MitchelPaulin:
Only shows the archived tag for anyone who has permissions to either assess or edit the current rubric, fix for both the compact and full screen view.

In a previous PR to hm-api-tests I added back the status types to the titles of the rubrics, the vdiff changes are mostly adding the title back. Another change is there is no longer a `not graded yet` on some of the rubrics. I noticed the students would see this `not graded yet` on a rubric preview. Hid the graded info if no assessment href was present. 

- back end change https://github.com/Brightspace/lms/pull/22681
- hm-api-test change https://github.com/Brightspace/hm-api-tests/pull/302

https://rally1.rallydev.com/#/57732444928ud/custom/373260458992?detail=%2Fuserstory%2F633538539905


src/components/rubric.js
```diff
@@ -691,21 +693,21 @@ export class Rubric extends composeMixins(
...
+		let assessorText = null;
 		const assessorName = this._assessorUserEntity?.userName;
 		if (assessorName) {
 			if (assessorName === 'unknown') {
 				assessorText = this.localize('rubricEvaluation:gradedByUnknown');
 			} else {
 				assessorText = this.localize('rubricEvaluation:gradedBy', 'name', assessorName);
 			}
-		} else {
+		} else if (this.assessmentHref) {
```
jesseag, line 710: When the component is initialized, I assume `this.assessmentHref = ''`

I would think we should display the text "not Graded yet" in that instance. In this instance, we won't display the text if the string is blank.

jesseag, line 710: When the component is initialized, I assume `this.assessmentHref = ''`

I would think we should display the text "not Graded yet" in that instance. In this instance, we won't display the text if the string is blank.

MitchelPaulin: I think even if the rubric is not graded yet, an assessment href is still present, a string may be empty on first render but its listed as a property so a render update would happen immediately after

jesseag: So the render update would guarantee the string wouldn't be empty even if no assessment has occurred?

jesseag (approved):
Looks good to me.

imcdonald-d2l (approved):

devpow112 (approved):

PR #1909, opened 2022-05-05
Remove duplicate langterms

devpow112:
There were a couple of terms that had the same value, removed them and adjusted names. Added unit tests to catch this sort of thing in the future.


mpharoah-d2l:
I assume it was a dependency update that caused the visual diff change?

devpow112:
> I assume it was a dependency update that caused the visual diff change?

@mpharoah-d2l I'm assuming so. It was a slight height change. Rerunning them to see if they fix themselves or not.

mpharoah-d2l (approved):

PR #1914, opened 2022-05-09
Hid rounded borders if there is feedback following the cell

MitchelPaulin:
Fixes the first issue in the ticket https://rally1.rallydev.com/#/57732444928ud/custom/373260458992?detail=%2Fuserstory%2F632211388427

This magic CSS incantation does not show the rounded borders if there is feedback directly following the cell


devpow112:
> This magic CSS incantation does not show the rounded borders if there is feedback directly following the cell

😆

devpow112 (approved):

PR #1916, opened 2022-05-09
Right align criterion name in rtl

MitchelPaulin:
Right align criterion name in RTL, by using a flex box we also handle the case of aligning the outcomes icon appropriately 

temp6


src/components/rubric-criteria-group.js
```diff
@@ -178,6 +178,12 @@ export class RubricCriteriaGroup extends composeMixins(
 					border: 1px solid var(--d2l-color-celestine);
 				}
 
+				:host([dir="rtl"]) .criterion-box {
```
devpow112, line 181: Should we have a default `.criterion-box` that has `display` and `justify-content` in it?
```css
.criterion-box {
  display: flex;
  flex-direction: row;
  justify-content: space-between;
}

:host([dir="rtl"]) .criterion-box {
  flex-direction: row-reverse;
}
```

src/components/rubric-criteria-group.js
```diff
@@ -178,6 +178,12 @@ export class RubricCriteriaGroup extends composeMixins(
 					border: 1px solid var(--d2l-color-celestine);
 				}
 
+				.criterion-box {
+  					display: flex;
```
devpow112, line 182: Hates the indenting here

MitchelPaulin:
@devpow112 I read there was some accessibility issues with `row-reverse` and screen readers not reading out the dom in the order it appears, found an alternative solution which just wraps everything in a flex box and lets the rtl vs ltr be handled automatically.

devpow112 (approved):

mpharoah-d2l (dismissed):

devpow112:
Looks like some padding or text size change in core? All kind of of small text changes.

mpharoah-d2l (approved):

PR #1921, opened 2022-05-10
Fixed feedback hover

MitchelPaulin:
Fixed both the jittering and highlighting issue with the feedback

temp6


src/components/rubric-feedback.js
```diff
@@ -91,13 +91,11 @@ export class RubricFeedback extends composeMixins(
...
 
 				.feedback-arrow {
-					margin-top: calc(-0.5rem - 42px);
+					margin-top: calc(-0.5rem - 43px);
 					width: 0;
 					height: 0;
 					border: 12px solid transparent;
 					position: absolute;
-					transition: border-color 0.5s;
-					transition-timing-function: ease;
```
MitchelPaulin, line 100: Not sure about this one, removed the transition so its consistent with the rest of the outline, I can add the transition to both if we want it

src/components/rubric-feedback.js
```diff
@@ -573,8 +556,6 @@ export class RubricFeedback extends composeMixins(
 				<div
 					class="feedback-wrapper"
 					?data-desktop="${!this.compact}"
-					@mouseover="${this._addFocusStylingToFeedbackWrapper}"
-					@mouseout="${this._removeFocusStylingFromFeedbackWrapper}"
```
MitchelPaulin, line 577: No longer needs to be handled by the feedback

src/components/rubric-feedback.js
imcdonald-d2l: makes sense to keep in consistent in my opinion 👍🏻

imcdonald-d2l (approved):

MitchelPaulin:
Vdiff changes were just from the 1px buffer I added to prevent the movement during hovering, each image was just `1px * numberOfFeedbackCells` larger

PR #1926, opened 2022-05-11
Made checkmark respect rtl

MitchelPaulin:


src/components/rubric-criterion-cell.js
```diff
@@ -59,9 +61,16 @@ class RubricCriterionCell extends composeMixins(
 				.check-icon.corner {
 					position: absolute;
 					bottom: 12px;
+				}
+
+				:host(:not([dir="rtl"])) .check-icon.corner {
```
devpow112, line 66: Why do we specifically need this instead of just including it in the class above?

devpow112 (approved):

src/components/rubric-criterion-cell.js
MitchelPaulin: If both right and left are set the checkmark sits in the center. And yeah I thought it was a bit akaward to try and unset the right property. Thought it would be better to just add a case for it

PR #1928, opened 2022-05-11
Pop out accessibility - DE48275

imcdonald-d2l:
Adding rubric name, accessibility status and graded status to aria label on rubric table. This will be read out when tabbing into the rubric table. Examples:

**Ungraded rubric**
`Assessing {rubricName} rubric. Rubric has not been graded yet. This table lists criteria...`
**Ungraded archived rubric**
`Assessing {rubricName} rubric. Rubric has been archived. Rubric has not been graded yet. This table lists criteria...`
**Graded rubric**
`Assessing {rubricName} rubric. Graded by D2L Support. This table lists criteria...`
**Graded archived rubric**
`Assessing {rubricName} rubric. Rubric has been archived. Graded by D2L Support. This table lists criteria...`


src/components/rubric-criteria-group.js
```diff
@@ -385,10 +385,21 @@ export class RubricCriteriaGroup extends composeMixins(
...
+		let gradedByText =  '';
+		if (this.context?.langterms?.assessorText) {
+			const assessorText = this.context.langterms.assessorText;
+			if (assessorText === this.localize('rubricEvaluation:notGradedSummary')) {
+				gradedByText = this.localize('rubricEvaluationAccessibility:unassessedRubric').concat(' ');
+			} else {
+				gradedByText = assessorText.concat('. ');
+			}
+		}
+		const tableAriaLabel = assessingRubricText + rubricArchivedText + gradedByText + this.localize('rubricEvaluationAccessibility:rubricSummaryA11y');
```
devpow112, line 399: We do this is a bunch of spots already but concatenating strings that will be translated can be problematic. The English structure of a sentence doesn't always match another languages structure. Is there any way we can just make all the combinations of the string for translation or use localization templating stuff instead `{}`. to fill in an overall sentence.

imcdonald-d2l: I had thought of this but didn't know if templating would help, for example the templated string would look like this:
```
"rubricEvaluationAccessibility:rubricSummaryA11y": "{assessingRubricText} {archivedRubricText} {gradedByText} This table lists criteria and...",
```

So when translating, they aren't really able to change the sentence structure since they don't have any context. Also, if one of the three text variables doesn't exist we have would have an extra space in the string. 

I believe `assessingRubricText` and `gradedByText` will always exist since a rubric will always have a name and some sort of evaluation status. We could have two separate lang terms for when archived text does/does not exist. For example:

```
"rubricEvaluationAccessibility:rubricSummaryA11y": "Assessing {rubricName} rubric. {gradedByText} This table lists criteria and...",
"rubricEvaluationAccessibility:rubricSummaryA11yArchived": "Assessing {rubricName} rubric. Rubric has been archived.{gradedByText} This table lists criteria and..."
```

So not a whole lot better.. not sure if it is worth it. Thoughts @devpow112?

devpow112: Honestly as a company we suck at thing kinds of stuff so just leave it. There's no good solution other then just writing out all the options in full which is dumb.

imcdonald-d2l:
Update: Mike/Andrew have approved the wording so good to review whenever.

src/components/rubric-criteria-group.js
```diff
@@ -385,10 +385,21 @@ export class RubricCriteriaGroup extends composeMixins(
...
+		let gradedByText =  '';
+		if (this.context?.langterms?.assessorText) {
+			const assessorText = this.context.langterms.assessorText;
+			if (assessorText === this.localize('rubricEvaluation:notGradedSummary')) {
+				gradedByText = this.localize('rubricEvaluationAccessibility:unassessedRubric').concat(' ');
+			} else {
+				gradedByText = assessorText.concat(this.localize('rubricShared:sentenceStop')).concat(' ');
+			}
+		}
+		const tableAriaLabel = assessingRubricText + rubricArchivedText + gradedByText + this.localize('rubricEvaluationAccessibility:rubricSummaryA11y');
```
MitchelPaulin, line 399: Rather than adding spaces to the end of all the strings it might be better to use the join method. Something like

```
[assessingRubricText, rubricArchivedText, gradedByText, this.localize('rubricEvaluationAccessibility:rubricSummaryA11y')].join(' ');
```
Would concatenate all the terms separated by a space

MitchelPaulin (approved):

devpow112 (approved):

src/components/rubric-criteria-group.js
imcdonald-d2l: Good to know, yeah that would be a bit cleaner. I'll do that 👍🏻

imcdonald-d2l: Actually, when testing locally it would add extra spaces when one of the lang term doesn't exist so will just leave as is for now. `rubricArchivedText` could equal `''` so it there would be a double space between the two other variables.

PR #1935, opened 2022-05-16
Reverting DE45861 Feedback fix.

jesseag:
This PR reverts:
https://github.com/Brightspace/d2l-rubric/pull/1843

Will be reverting this in all main branches as well as 22.6, 22.5. This is due to the fact of this fix being hacky as well as causing other bugs/issues in Safari that we cannot fix at this moment.


imcdonald-d2l (approved):
You can just comment `/backport` on the PR and it will create a backport PR automatically

PR #1936, opened 2022-05-16
[Back-port 20.22.5] Reverting DE45861 Feedback fix.

jesseag:
Backporting revert
Original PR: https://github.com/Brightspace/d2l-rubric/pull/1935


imcdonald-d2l:
Vdiff PR looks safe to merge - looks like the font color might have changed slightly 😕

jesseag:
> Vdiff PR looks safe to merge - looks like the font color might have changed slightly

Do I need to run it by the team? I don't see the change in font color. The only thing I can see that has changed is some of the Vdiffs are taller by 2px. I'm not even really sure why Vdiffs got changed. My best guess is because I removed `@d2l-htmleditor-focus="${this._feedbackEditorHandler}"` from the `rubric-cell-editor` and Vdiffs thought there was a potential visual update since I changed html.

imcdonald-d2l:
> Do I need to run it by the team?
No, I would just merge it. Should be safe.

MitchelPaulin (approved):
Most of the vidff stuff looks fine, some minor text shifting probably from changes in core, The removed code doesn't look like it would effect it

PR #1954, opened 2022-05-25
VDiff speed improvements

devpow112:
No longer use strict timeouts and instead wait on network traffic idle a bit better. The default network wait times don't work cause of hypermedia slowness. Also better async management around waiting for component rendering/updating. This greatly reduces vdiff run time and generates more accurate/less flaky results. Down from 30 min to just 8ish. This also removes the need for the demo help to add tokens since they are just added directly in the HTML. Still more changes to make that will drive down that time but want to get this in since the last bit is taking longer to sort out.


devpow112:
Figure out why compact assessed failed

devpow112:
VDiff changes are due to a change in the width of one of the tests as well as some focus changes due to asking components to update and waiting till components actually indicate they are ready.

devpow112:
Couple of minor flaky issues but will fix those separately as they are not frequent.

MitchelPaulin (approved):

devpow112:
Ok apparently something broke from yesterday to today. Will try and figure out 😞.

test/perceptual/d2l-rubric-editor.visual-diff.js
```diff
@@ -28,7 +28,7 @@ describe('d2l-rubric-editor', function() {
 					`${visualDiff.getBaseUrl()}/test/perceptual/${this.title}.visual-diff.html?dir=${dir}`,
 					{ waitUntil: ['load'] }
 				);
-				await page.waitForNetworkIdle({ idleTime: 2500 });
+				await page.waitForNetworkIdle({ idleTime: 2000, timeout: 400000 });
```
devpow112, line 31: This defaults to 30 seconds which seemed right up against the page load time for the rubrics editor page.

devpow112:
Hopefully that will take care of it. Will result in a new set of vdiffs being merged with minor changes

mpharoah-d2l (approved):

devpow112:
Finally passed!!!! Will work on compact flakiness after this merges. Apologies for the re-work.

MitchelPaulin (approved):

PR #1961, opened 2022-05-26
Improve flakiness of vdiff compact assessed tests

devpow112:
Hopefully fixes issues with compact assessed tests being flaky in terms of assessment info being loaded in time for the screenshot. There's still room for improvement (want to try and combine html pages into 1 page), but will do that separately. Makes test a bit slower (11min vs 8min) but still much better. The combining of the compact pages should reduce the time again in the future.


devpow112:
Still some flakiness around rubric editor but I think I have a fix. Will apply that later today

imcdonald-d2l (approved):

PR #1965, opened 2022-05-26
Add partial edit read-only vdiff

devpow112:
This also toys with some newer functionality in puppeteer that allows you to interact with elements directly on the page within the test framework as opposed to doing `page.evaluate`. Test that was adjusted is still skipped due to issue with HTML editor but it does result in the same (bad) vdiff.


test/perceptual/d2l-rubric-editor.visual-diff.js
```diff
@@ -42,19 +42,31 @@ describe('d2l-rubric-editor', function() {
...
-					{ name: '_textEditable', value: true },
-					{ name: '_showReadOnlyRubric', value: false },
-					{ name: '_showLockedAlert', value: false }
-				]);
+				const rect = await visualDiff.getRect(page, id);
+
+				await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { clip: rect });
+			});
+
+			it.skip('partial edit', async function() {
```
devpow112, line 55: If we can figure out why the HTML editor doesn't resize properly then we can unskip this test.

test/perceptual/helpers/vdiff.js
```diff
@@ -131,3 +131,9 @@ export const sanitizeName = (name) => {
 		.join('-')
 		.toLowerCase();
 };
+
+export const getElementHandle = async(page, selector) => {
```
devpow112, line 135: Hoping we can switch some of our more heavy handed helpers to just use this and poke at the element handle it gives back.

devpow112:
Renamed `waitUntilComponentReady` to `waitForComponentReady` to be more consistent with puppeteer stuff.

MitchelPaulin (approved):

imcdonald-d2l (approved):

PR #1971, opened 2022-05-27
Add better support for deep query selector

### LINKED ISSUE #1744: Look into ways to speed up vdiff
Maybe some sort of rubric loading is done events system.

devpow112:
This registers a native custom CSS selection handler to puppeteer. There is a built in `pierce/` selector but it doesn't work they way we need it. Luckily the `query-selector-shadow-dom` package provides a ready built plugin for puppeteer. Register that plugin as `shadow/` and switch to using it. Now any CSS selector given that starts with `shadow/` will pierce shadow roots using `query-selector-shadow-dom`. This is a good candidate to push into the vdiff framework. Resolves #1744.


devpow112:
Currently the `openCompactDemoPageElements` doesn't work in the demo pages due to timing. Basically it runs before elements are ready. This is broken without y changes so I'll have to look into fixing it.

mpharoah-d2l (approved):

imcdonald-d2l (approved):

PR #1977, opened 2022-05-30
Render read only rubric when rubric is shared - DE46459

imcdonald-d2l:
When viewing a shared rubric we didn't allow a read-only view to be displayed. This PR fixes that issue as well as a small DOM cleanup item.

| Before | After |
| --- | --- |
| shared-rubric-before | shared-rubric-after |


src/components/rubric.js
```diff
@@ -730,6 +730,10 @@ export class Rubric extends composeMixins(
 			}
 		};
 
+		if (!archivedStatus && !assessorText) {
+			return;
+		}
```
imcdonald-d2l, line 735: This is unrelated to the shared rubric issue but cleans up a styling issue. It would render the div container even if nothing was rendered within it. See below:
| Before | After |
| --- | --- |
| div-container-before | div-container-after |

devpow112 (dismissed):

src/components/rubric.js
imcdonald-d2l: Just an FYI that vdiff changes are all related to this change.

devpow112:
> Visual diff tests failed - pull request #1979 has been opened with the updated goldens.

Looks like this resulted in a bunch of whitespace under the title being added, I'm guessing cause something is now "showing" that wasn't before. We may want to look into eliminating that whitespace if it's not needed (can be a future thing).

MitchelPaulin (approved):

devpow112 (approved):

PR #1980, opened 2022-05-30
US139689 - Add back ability to disable HTML editor for evaluation (Polymer)

devpow112:
This is needed as a stop gap for figuring out the issues with HTML editor and portfolio. For now we will be disabling rich text in portfolio for rubric evaluation. Partial revert of https://github.com/Brightspace/d2l-rubric/pull/1821. Decided to go with explicit disabling of rich text instead of explicit enabling of it.

Lit: https://github.com/Brightspace/d2l-rubric/pull/1981.

https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F638847736809


devpow112:
Due to my laziness removing this, re-enabling it was easy 😄.

d2l-rubric.js
```diff
@@ -288,7 +288,7 @@ const template = `<dom-module id="d2l-rubric">
 						enable-feedback-copy="[[enableFeedbackCopy]]"
 						read-only="[[readOnly]]"
 						compact="[[compact]]"
-						rich-text-enabled="[[richTextEnabled]]"
+						rich-text-enabled="[[!richTextDisabled]]"
```
jesseag, line 291: If we are disabling this, would we want `rich-text-enabled="[[richTextDisabled]]"` right?

Or we could change line 498 to `value: true` and leave Line 291 unchanged.

imcdonald-d2l (approved):

PR #1981, opened 2022-05-30
US139689 - Add back ability to disable HTML editor for evaluation (Lit)

devpow112:
This is needed as a stop gap for figuring out the issues with HTML editor and portfolio. For now we will be disabling rich text in portfolio for rubric evaluation. Partial revert of https://github.com/Brightspace/d2l-rubric/pull/1791. Decided to go with explicit disabling of rich text instead of explicit enabling of it.

Polymer: https://github.com/Brightspace/d2l-rubric/pull/1980.

https://rally1.rallydev.com/#/57732444928d/custom/373260458992?detail=%2Fuserstory%2F638847736809


devpow112:
Will force a dependency update after the polymer version is merged so we have the dependency chain correct in `master`.

src/components/rubric-feedback.js
```diff
@@ -458,6 +470,52 @@ export class RubricFeedback extends composeMixins(
...
+			></d2l-htmleditor>`;
+		} else {
+			const isIOS = /iPad|iPhone/.test(navigator.userAgent) && !window.MSStream;
+
+			newEditor = html`<div class="feedback-heading d2l-label-text">
+				${this.localize('rubricEvaluation:criterionFeedback')}
+				<div class="feedback-heading-extra d2l-label-text">
+					${this._feedbackHeadingExtra || ''}
+				</div>
+			</div>`;
```
imcdonald-d2l, line 498: `nit` variable name `newEditor` isn't entirely reflective of how it's being used. Here we assign it to a header value. Should this be included `legacyEditor` variable?

devpow112: This is just a revert of what was previously removed. I was being lazy and didn't do any clean-up. I can rename these to something like `htmlEditor` and `plainTextInput` if that's more clear.

devpow112: This really should just be part of the `legacyEditor` (`plainTextInput`) block anyways.

imcdonald-d2l: Some small clean up wouldn't hurt if it's going to be kicking around for a bit. If it will be removed soon (🤞🏻) we can ignore for now

devpow112: No idea on re-removal timeline so I'll do some clean-up.

imcdonald-d2l (approved):

PR #1993, opened 2022-06-08
Fix automation issues

devpow112:
For some reason in automation an inline comment is just being displayed. Looks like something is striping the starting comment tag. Pull out comment into JavaScript comment instead of HTML. Recommend reviewing with [white-space diff off](https://github.com/Brightspace/d2l-rubric/pull/1993/files?w=1).

image


devpow112:
I was able to repo this once locally. Seems like lit is removing the comment tag for some reason.

mpharoah-d2l (approved):

